PR #4, opened 2019-04-23
UnxUtils: Add version 2007.03.01

niheaven:
ref and close: https://github.com/lukesampson/scoop/pull/3357
close: https://github.com/lukesampson/scoop/issues/3288


niheaven:
Yes, it's not updated since 2003, so why someone else still use it? :smirk:

Ash258:
Add closing directive for core PRs (for all your ported PRs). No need to manually close it then.

PR #6, opened 2019-04-23
Python: Use install script instead of original installer

niheaven:
ref and close: https://github.com/lukesampson/scoop/pull/3197
close https://github.com/lukesampson/scoop/issues/2612


bucket/python.json
```diff
@@ -1,35 +1,32 @@
 {
     "homepage": "https://www.python.org/",
-    "license": "https://docs.python.org/3/license.html",
+    "license": "Python-2.0",
     "version": "3.7.3",
+    "depends": "dark",
```
Ash258: Add description

bucket/python.json
```diff
@@ -1,35 +1,32 @@
...
-            "InstallAllUsers=$(@{$true=1;$false=0}[$global])",
-            "TargetDir=\"$dir\"",
-            "AssociateFiles=0",
-            "Shortcuts=0",
-            "InstallLauncherAllUsers=$(@{$true=1;$false=0}[$global])"
-        ],
-        "keep": true
+        "script": [
+            "&(file_path dark dark.exe) -nologo -x \"$dir\\_tmp\" \"$dir\\$fname\" | Out-Null",
+            "Remove-Item \"$dir\\_tmp\\AttachedContainer\\launcher.msi\" | Out-Null",
```
Ash258: I would rather go with. ALso remove-Item do not show any data.

```suggestion
            "@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {",
            "    Remove-Item \"$dir\\_tmp\\AttachedContainer\\$_\"",
            "}",
```

bucket/python.json
```diff
@@ -1,35 +1,32 @@
...
+            "Remove-Item \"$dir\\_tmp\\AttachedContainer\\path.msi\" | Out-Null",
+            "Remove-Item \"$dir\\_tmp\\AttachedContainer\\pip.msi\" | Out-Null",
+            "Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\" | ForEach-Object { extract_msi $_ $dir }",
+            "Remove-Item \"$dir\\_tmp\", \"$dir\\*.msi\", \"$dir\\$fname\" -Force -Recurse",
+            "&(\"$dir\\python.exe\") -E -s -m ensurepip -U --default-pip | Out-Null",
+            "if ($global) { $pathext = (env 'PATHEXT' $true) -replace ';.PYW?', ''; env 'PATHEXT' $true \"$pathext;.PY;.PYW\" }"
```
Ash258: Make this if multile please.

bucket/python.json
```diff
@@ -1,8 +1,9 @@
 {
     "homepage": "https://www.python.org/",
+    "description": "A programming language that lets you work quickly and integrate systems more effectively.",
     "license": "Python-2.0",
     "version": "3.7.3",
-    "depends": "dark",
+    "depends": "main/dark",
```
Ash258: No need to prefix it with main. Everything will behave completely same as before

niheaven: Okay, I thought every 'main' manifest should be explicit writen.

bucket/python.json
```diff
@@ -1,34 +1,40 @@
...
-            "Shortcuts=0",
-            "InstallLauncherAllUsers=$(@{$true=1;$false=0}[$global])"
-        ],
-        "keep": true
+        "script": [
+            "&(file_path dark dark.exe) -nologo -x \"$dir\\_tmp\" \"$dir\\$fname\" | Out-Null",
+            "@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {",
+            "    Remove-Item \"$dir\\_tmp\\AttachedContainer\\$_\"",
+            "}",
+            "Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\" | ForEach-Object { extract_msi $_ $dir }",
```
Ash258: I am not sure if this will work, gci returns only file name since `toString()` will be called (most probably) when passing to function. 

So this will try to exract $_ inside current directory (where user is)

Ash258: i would go with:

```suggestion
            "(Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\").FullName | ForEach-Object { extract_msi \"$_\" \"$dir\" }",
```

bucket/python.json
```diff
@@ -44,23 +50,23 @@
...
             },
             "32bit": {
-                "url": "https://www.python.org/ftp/python/$version/python-$version.exe#/py3.exe"
+                "url": "https://www.python.org/ftp/python/$version/python-$version.exe"
             }
         },
         "hash": {
             "url": "https://www.python.org/downloads/release/python-$cleanVersion/",
-            "find": "<td><a href=\".*$basename\".*</a></td>\\s+<td>Windows</td>\\s+<td>.*</td>\\s+<td>([A-Fa-f\\d]{32})"
+            "find": "$basename[\\S\\s]+?$md5"
```
Ash258: Will this get correct hash?
If i remember right I made it this strict to match only the correct one hash.

bucket/python.json
Ash258: Or this should work too:

```suggestion
            "Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\" | ForEach-Object { extract_msi \"$($_.FullName)\" \"$dir\" }",
```

Ash258: Yeah to string will be called 
https://github.com/lukesampson/scoop/blob/131cf26b74c7aa74e2d8bbf39bd2f1205718a80a/lib/decompress.ps1#L21

So whole object will be converted into string. And it will look like this:
image

Write-Output expect object and format it using table.
But write-host convert everything to string, which return only name

bucket/python.json
niheaven: Yes, the regex find the next md5 hash after basename.

checkver

bucket/python.json
niheaven: I inserted `debug "$path"` in `extract_msi`, and found that `System.IO.FileSystemInfo` outputted by `Get-ChildItem` will transform to full path in double quote. So this is enough for extraction...

python

Ash258: What powershell version?

niheaven: And double quote is added inside `extract_` function, so it doesn't need to add them while calling function, `extract_msi $path $to` is okay here, white spaces will pass through.

niheaven: > 
> 
> What powershell version?

5.1 in Windows 10. I can't test in v3 now, but if https://github.com/lukesampson/scoop/pull/3330 is merged, there's no problem.

Ash258: That is not true. While powershell is executing `Invoke-Expression` (https://github.com/lukesampson/scoop/blob/master/lib/install.ps1#L1058), there are all variables "replaced" replaced with actual values.

So this:
`extract_7zip $dir`, which is valid in normal powershell
become this:
`extract_7zip C:\Users\UserName WIth Spaces\directory` which lead to error.

Test

Ash258: That is the reason why every variable which contains path, needs to be quoted in all manifests

Ash258: PWSH 6.2.0 vs Powershell 5.1

A

niheaven: The function is `Invoke-Expression "lessmsi x "$path" "$to\""` (remove some \`, https://github.com/lukesampson/scoop/blob/master/lib/decompress.ps1#L52), and I check every extract function and every one quoted `$path` in double quote, like this:

quote

niheaven: Yes, the full name of `$path` that passed through in function is strange, and it works in PS5 and PS6. I'm not sure why, may investigate tomorrow.

niheaven: Oh, I got your idea... I'll fix the quote😅

Ash258: You showed wrong line. 

Invoke expression will look like this:

```powershell
&(file_path dark dark.exe) -nologo -x \"C:\ User Space\\_tmp\" \"C:\ User Space\\$fname\" | Out-Null
@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {
    Remove-Item \"C:\User Sapce\\_tmp\\AttachedContainer\\$_\"
}
Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\" | ForEach-Object {
    extract_msi C:\User Space C:\User Space
}
```

Ash258: At least you conviced me to create new HyperV VM with username with spaces for testing :D

niheaven:
@Ash258 I try to add space in dir name, and it pass through `$path` correctly. Do we misunderstand something?

If pass through a string to `Invoke-Expression`, it will not expand variables in the string, so white spaces will pass through, and adding double quotes besides these variables is in fact unnecessary。

white-space

niheaven:
And for FullName, since `Get-ChildItem` return `System.IO.FileSystemInfo`, and this type is pass to inner function, if needed, fullname is used.

fullname

Ash258:
Final word on the fullnames. Here is default powershell (windows 10, right) and latest pwsh.
There you can see different behaviour. Same folder, same system, different output.

image

https://github.com/PowerShell/PowerShell/issues/7084

niheaven:
Yes, for safety, FullName is better.

niheaven:
What about double quotes? `Invoke-Expression` with param as `[String]` needn't double quotes, I'm sure.

Ash258:
I will test lately today with clean W10. Now I will be busy with work stuff for few next hours.

bucket/python.json
```diff
@@ -1,34 +1,40 @@
...
-            "Shortcuts=0",
-            "InstallLauncherAllUsers=$(@{$true=1;$false=0}[$global])"
-        ],
-        "keep": true
+        "script": [
+            "&(file_path dark dark.exe) -nologo -x \"$dir\\_tmp\" \"$dir\\$fname\" | Out-Null",
+            "@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {",
+            "    Remove-Item \"$dir\\_tmp\\AttachedContainer\\$_\"",
+            "}",
+            "(Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\").FullName | ForEach-Object { extract_msi $_ $dir }",
```
Ash258: I will not have time in near future to test when quotes are needed. I would suggest to follow best practises and quote upper context variables, which contains whitespaces.
```suggestion
            "(Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\").FullName | ForEach-Object { extract_msi $_ \"$dir\" }",
```

niheaven: Ok, add quotes now, and wait for further test.

bucket/python.json
```diff
@@ -1,34 +1,40 @@
 {
     "homepage": "https://www.python.org/",
-    "license": "https://docs.python.org/3/license.html",
+    "description": "A programming language that lets you work quickly and integrate systems more effectively.",
+    "license": "Python-2.0",
     "version": "3.7.3",
+    "depends": "dark",
```
Ash258: We could open discussion if prefixing main bucket requirements is suitable. When i am thinking about this, i do not know how current implementation will handle multiple manifests with same name.

For Examle `main/dark`, 'Ash258/dark' If dark is in both buckets, mine will be used (alphabetic order, as far my testing with simple `scoop install Wavebox`, `Ash258` bucket is preferred over `extras`, which contains Wavebox too)

PR #15, opened 2019-04-28
Add ffmpeg-shared version 4.1.3

StelFux:
Hello,
I need the ddl libraries with ffmpeg, which are not provided in the stastic version, for Audacity.

### DIFF
--- a/bucket/ffmpeg-shared.json
+++ b/bucket/ffmpeg-shared.json
@@ -1,0 +1,35 @@
+{
+    "version": "4.1.3",
+    "homepage": "https://ffmpeg.zeranoe.com/builds/",
+    "license": "GPL-3.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://ffmpeg.zeranoe.com/builds/win64/shared/ffmpeg-4.1.3-win64-shared.zip",
+            "hash": "0b974578e07d974c4bafb36c7ed0b46e46b001d38b149455089c13b57ddefe5d",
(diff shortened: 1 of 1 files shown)


bucket/ffmpeg-shared.json
```diff
@@ -0,0 +1,35 @@
+{
+    "version": "4.1.3",
+    "homepage": "https://ffmpeg.zeranoe.com/builds/",
+    "license": "GPL-3.0",
```
Ash258: Wrong identifier

bucket/ffmpeg-shared.json
```diff
@@ -0,0 +1,35 @@
+{
+    "version": "4.1.3",
+    "homepage": "https://ffmpeg.zeranoe.com/builds/",
```
Ash258: No description

r15ch13:
https://github.com/lukesampson/scoop/pull/3404

bucket/ffmpeg-shared.json
StelFux:
```suggestion
    "license": {
        "identifier": "GPL-3.0-only"
    },
```
Is it better this way?

r15ch13: Already fixed with https://github.com/scoopinstaller/scoop-main/commit/ab92aa791c959d26be824986844d571c2fc45286 :smile:

Ash258: No. Do not use object unless you put URL.

Just use `"license": "GPL-3.0-only"`

PR #34, opened 2019-05-05
avro-tools: Add version 1.8.2

blachniet:
This is my first contribution to Scoop, so please let me know if there is anything additional I need to do.

### DIFF
--- a/bucket/avro-tools.json
+++ b/bucket/avro-tools.json
@@ -1,0 +1,25 @@
+{
+    "version": "1.8.2",
+    "description": "A collection of Avro tools. Data serialization system.",
+    "homepage": "http://avro.apache.org/",
+    "license": "Apache-2.0",
+    "url": "http://apache.org/dist/avro/avro-1.8.2/java/avro-tools-1.8.2.jar#/avro-tools.jar",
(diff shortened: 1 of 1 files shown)


bucket/avro-tools.json
```diff
@@ -0,0 +1,25 @@
...
+            "java/oraclejdk",
+            "java/openjdk"
+        ]
+    },
+    "checkver": {
+        "url": "http://avro.apache.org/releases.html",
+        "regex": "Avro ([\\d.]+) Released"
+    },
+    "autoupdate": {
+        "url": "http://apache.org/dist/avro/avro-$version/java/avro-tools-$version.jar#/avro-tools.jar",
```
h404bi:
```suggestion
        "url": "https://www.apache.org/dist/avro/avro-$version/java/avro-tools-$version.jar#/avro-tools.jar",
```

h404bi:
Oh, too late.

bucket/avro-tools.json
Ash258: I used `http` because main page is not https so i did not change url. But if the file is available then yes, `https` would be better

PR #41, opened 2019-05-08
ghostscript: Fix checkver and update to 9.27

ssfjhh:
Fix typos.

### DIFF
--- a/bucket/ghostscript.json
+++ b/bucket/ghostscript.json
@@ -1,19 +1,19 @@
 {
     "homepage": "https://www.ghostscript.com",
-    "version": "9.26",
+    "version": "9.27",
     "license": "AGPL-3.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs926/gs926w64.exe#/dl.7z",
-            "hash": "sha512:4c961b852bd207e9a1cae00355c5e582ad047dec54912ae5e7d326c4eacc0c8300bf49164d7ba5817f9ab5a7b9cb5d60f235b162dd30c0d6e8706566f7e0a605",
+            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs927/gs927w64.exe#/dl.7z",
+            "hash": "sha512:49fc973bfae12f366a837bcc39d93e3d8e3c88359f7d45279df899f0b2941c264efc47815966867d332b32e87ca8d88697edf7792b162645ed7a549cecc373b2",
             "bin": [
                 "bin\\gswin64.exe",
                 "bin\\gswin64c.exe"
             ]
         },
         "32bit": {
-            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs926/gs926w32.exe#/dl.7z",
-            "hash": "sha512:5e3f9e8303e1b65c9c6b67dbd3c5a999c7b7eeafe499638d3bbe718ccb1a1b26928dd441d65431b423f385b1a70d698e5027f7ee701c963eb410d3f79fde5a40",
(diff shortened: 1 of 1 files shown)


bucket/ghostscript.json
```diff
@@ -1,19 +1,19 @@
 {
     "homepage": "https://www.ghostscript.com",
-    "version": "9.26",
+    "version": "9.27",
     "license": "AGPL-3.0",
```
Ash258: Missing description

bucket/ghostscript.json
```diff
@@ -1,19 +1,19 @@
...
                 "bin\\gswin64.exe",
                 "bin\\gswin64c.exe"
             ]
         },
         "32bit": {
-            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs926/gs926aw32.exe#/dl.7z",
-            "hash": "sha512:5e3f9e8303e1b65c9c6b67dbd3c5a999c7b7eeafe499638d3bbe718ccb1a1b26928dd441d65431b423f385b1a70d698e5027f7ee701c963eb410d3f79fde5a40",
+            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs927/gs927w32.exe#/dl.7z",
+            "hash": "sha512:ef18548c62a5107f951d839d7c83ea5e03b35b2b85fd2c820b95687fb28e9ebfc799c2efe87d6db16879a8d9d995fe56f0b6a1248df0d6ee41d5851edfce9f83",
             "bin": [
```
Ash258: Why to create separate bin, when bin folder is added to path??

bucket/ghostscript.json
```diff
@@ -1,19 +1,19 @@
...
+    "version": "9.27",
     "license": "AGPL-3.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs926/gs926aw64.exe#/dl.7z",
-            "hash": "sha512:4c961b852bd207e9a1cae00355c5e582ad047dec54912ae5e7d326c4eacc0c8300bf49164d7ba5817f9ab5a7b9cb5d60f235b162dd30c0d6e8706566f7e0a605",
+            "url": "https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs927/gs927w64.exe#/dl.7z",
+            "hash": "sha512:49fc973bfae12f366a837bcc39d93e3d8e3c88359f7d45279df899f0b2941c264efc47815966867d332b32e87ca8d88697edf7792b162645ed7a549cecc373b2",
             "bin": [
                 "bin\\gswin64.exe",
```
Ash258: If you want to keep binaries, i would really suggest to add new one, renamed without architecture.

`gswin64c -> gswin` I assume, that `c` means console

bucket/ghostscript.json
niheaven: No `bin` dir should ok.

bucket/ghostscript.json
niheaven: For binary names, some apps that use ghostscript are looking for such file names, so renaming is not prefered.

Ash258: > to add new one

I did not said delete current ones. I said add new one with renames, so user can freely type `gscript` | `gs` instead of `gswin64c`
When you add renames, it will not break anything

niheaven: Got it.

Ash258:
Delete bin from path will not break anything?

What does bin folder contains? Did bin folder grow over time (some packages, ...)?

niheaven:
> 
> 
> Delete bin from path will not break anything?
> 
> What does bin folder contains? Did bin folder grow over time (some packages, ...)?

There are four files in `bin`, two binaries, one dll and one lib. The other scripts is in `lib` dir. No package will affect this dir.

GS

PR #48, opened 2019-05-10
Add gettext version 0.20-v1.16

edgardmessias:


bucket/gettext.json
```diff
@@ -0,0 +1,48 @@
+{
+    "homepage": "http://mlocati.github.io/articles/gettext-iconv-windows.html",
```
Ash258: Use https
```suggestion
    "homepage": "https://mlocati.github.io/articles/gettext-iconv-windows.html",
```

image

edgardmessias:
> Why is shared preferred over static?

Small size, and both work well in my environment.

PR #64, opened 2019-05-15
aws-iam-authenticator: Fix checkver

brandon93s:
Extra quote is causing `checkver` to fail.


brandon93s:
image

This check seems unnecessary. If there isn't conflicts, it should be OK to squash and merge.

@r15ch13 @h404bi

PR #65, opened 2019-05-15
Add Apache Kafka, a distributed streaming platform

nightscape:


bucket/kafka.json
```diff
@@ -0,0 +1,49 @@
...
+            "java/openjdk"
+        ]
+    },
+    "checkver": {
+        "url": "https://kafka.apache.org/downloads",
+        "regex": "current stable version is ([\\d.]+)\\."
+    },
+    "autoupdate": {
+        "url": "https://www.apache.org/dist/kafka/$version/kafka_2.12-$version.tgz",
+        "extract_dir": "kafka_2.12-$version"
```
Ash258: Add hash extraction

```suggestion
        "hash": {
            "url": "$url.sha512"
        },,
        "extract_dir": "kafka_2.12-$version"
```

A

bucket/kafka.json
```diff
@@ -0,0 +1,49 @@
+{
+    "version": "2.2.0",
+    "description": "Apache Kafka. A distributed streaming platform.",
+    "homepage": "https://kafka.apache.org/",
+    "license": "Apache-2.0",
+    "url": "https://www.apache.org/dist/kafka/2.2.0/kafka_2.12-2.2.0.tgz",
```
Ash258: What is that `2.12`? Where does it come from?

nightscape: Kafka is written in Scala. `2.12` is the Scala version with which this binary of Kafka was compiled.

bucket/kafka.json
nightscape: I tried with the `.sha1` file but that didn't work.
My guess is that the format used in that file is not supported in scoop:
```
kafka_2.12-2.2.0.tgz: 360B 39BD 5B7C D259 A47D  D277 966C 3314 0EFF FB8C
```
I'll try with the `.sha512` version as well, but that one looks similar:
```
kafka_2.12-2.2.0.tgz: 63FDCD05 B857A096 D2E72804 253590BD 086E4F30 46FBFA54
                      6F186602 E48F5471 2ED6DE2A 346A5B05 D5335870 BF5268FC
                      B2BFD3EF 9C4F4BBE A379091A 79604556
```

nightscape: Nope, same problem:
```
.\bin\checkver.ps1 kafka -u
kafka: 2.2.0 (scoop version is 2.1.0) autoupdate available
Autoupdating kafka
Searching hash for kafka_2.12-2.2.0.tgz in https://www.apache.org/dist/kafka/2.2.0/kafka_2.12-2.2.0.tgz.sha512
Could not find hash in https://www.apache.org/dist/kafka/2.2.0/kafka_2.12-2.2.0.tgz.sha512
Downloading kafka_2.12-2.2.0.tgz to compute hashes!
Loading kafka_2.12-2.2.0.tgz from cache
Computed hash: a009624fae678fa35968f945e18e45fbea9a30fa8080d5dcce7fdea726120027
```

The constructed URL is correct, but scoop doesn't understand the (rather strange) notation in the file.

r15ch13: Will work on a fix for removing the linebreaks from the hash.

The `.sha1` file keeps everything in a single line and it can be matched with this: 

```json
        "hash": {
            "url": "$url.sha1",
            "regex": "$basename:[\\x20\\t]+([a-fA-F0-9\\s]+)\\n+"
        }
```

nightscape: @r15ch13 that worked, thanks for the tip!
Amended to the original commit.
@Ash258 good to merge?

PR #72, opened 2019-05-17
Add koffee version 0.4.1

mattstermiller:
Adding Koffee, which is my gui-based, keyboard-driven file explorer.


mattstermiller:
Thanks. Created PR on extras here: https://github.com/lukesampson/scoop-extras/pull/2165

PR #75, opened 2019-05-18
srvman: Fix download url

brandon93s:


brandon93s:
> Add Description
> Add License

Trying to restore download functionality, no sense in blocking a fix to add those to an existing manifest.

> Why not autoupdate?
> That number after getfile is random?
Yes. It has also been at 1.0 for quite a while --- not actively changing.

Ash258:
Then why not add some regex like `getfile/(?<id>[\\d]+)/....7z` and autoupdate with `$matchId/...7z`

brandon93s:
It's been 1.0 since the manifest was added in 2016: 

https://github.com/lukesampson/scoop/pull/1155/files

May be better off just removing the checkver. Thoughts?

r15ch13:
https://github.com/ScoopInstaller/Main/pull/77

PR #79, opened 2019-05-20
julia: Use installer.script instead of original installer

niheaven:
`julia-installer.exe` is a NSIS installer.

### DIFF
--- a/bucket/julia.json
+++ b/bucket/julia.json
@@ -13,21 +13,10 @@
         }
     },
     "installer": {
-        "file": "julia-installer.exe",
-        "args": [
-            "/S",
-            "/D=$dir"
+        "script": [
+            "Expand-7zipArchive \"$dir\\julia-installer.exe\" \"$dir\" -Removal",
+            "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall.exe\" -Recurse -Force"
         ]
-    },
-    "uninstaller": {
-        "file": "uninstall.exe",
-        "args": "/S"
-    },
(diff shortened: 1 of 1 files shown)


bucket/julia.json
```diff
@@ -13,22 +13,11 @@
...
-        "file": "uninstall.exe",
-        "args": "/S"
-    },
-    "env_set": {
-        "JULIA_HOME": "$dir\\bin",
-        "JULIA_EXE": "julia.exe",
-        "JULIA": "$dir\\bin\\julia.exe",
-        "JL_PRIVATE_LIB_DIR": "bin"
-    },
     "bin": [
```
Ash258: Irrelevant rename
```suggestion
    "bin": "bin\\julia.exe",
```

bucket/julia.json
```diff
@@ -13,22 +13,11 @@
...
+        "script": [
+            "Expand-7zipArchive \"$dir\\julia-installer.exe\" \"$dir\" -Removal",
+            "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall.exe\" -Recurse -Force"
         ]
     },
-    "uninstaller": {
-        "file": "uninstall.exe",
-        "args": "/S"
-    },
-    "env_set": {
```
Ash258: Why to remove these sets?

niheaven: I've installed julia using original installer, and it didn't set these environment variables. These may be set by previous version, and didn't be used by current one.

niheaven:
Done.

PR #91, opened 2019-05-27
msys2: Add shortcuts and persist, fix bins and path

niheaven:
Close https://github.com/lukesampson/scoop/issues/2884

For `MSYS2_ROOT`, I haven't found ref for this env path, and for the newest `msys2_shell.cmd`, any other envs is not needed.

For original `env_add_path`, all binaries in root path are ported (`autorebase.bat` is only needed by 32bit msys2), and remove adding root to env.

### DIFF
--- a/bucket/msys2.json
+++ b/bucket/msys2.json
@@ -14,13 +14,48 @@
             "hash": "sha1:ff86c3e4ef8777074fd394510b95943d0c943956"
         }
     },
+    "pre_install": [
+        "if ($architecture -eq '32bit') {",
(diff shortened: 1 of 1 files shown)


r15ch13:
ping @indrakaw

indrakaw:
This is good.

indrakaw:
I'm sorry, I already have latest MSYS2 installed. What should I do to get the persistent directory updated?

niheaven:
`scoop update msys2 -f`, run `msys2` in any console to finish setup, and then copy all content in your old msys2's `home` dir to `persist\msys2\home`.

PR #93, opened 2019-05-27
bzip2: Update to version 1.0.6.1

brandon93s:


bucket/bzip2.json
```diff
@@ -21,5 +22,15 @@
     "pre_install": "cp \"$dir\\bzip2.exe\" \"$dir\\bunzip2.exe\"; cp \"$dir\\bzip2.exe\" \"$dir\\bzcat.exe\"",
```
Ash258: Add binaries for this:

bin: [bzip2.exe, [bzip2.exe, bzcat], [bzip2.exe, bunzip2]]

brandon93s: Current bin:

```json
    "bin": [
        "bunzip2.exe",
        "bzcat.exe",
        "bzip2.exe",
        "bzip2recover.exe"
    ]
```

What is your proposal?

Ash258: You do not see problem when you copy same file twice with different name. 

That is nonsense. 

Just use that as i wrote

```yml
bin:
  - bzip2.exe
  - - bzip2.exe
    - bzcat
  - - bzip2.exe
    - bunzip2
```

_Too lazy to quote json_

rasa:
According to https://github.com/philr/bzip2-windows/releases , this release depends on vcredist2015. I wonder is there a way to find that out programmatically, say with cygpath?

PR #100, opened 2019-05-30
winpython: Add 'extract_dir'

niheaven:
Example of `innosetup: true` w/ `extract_dir`.

Please MERGE AFTER https://github.com/lukesampson/scoop/commit/e6b2e3d31e246d2af70bd32807159b761fb65e0d merged into `master`

PS: I don't like this manifest that use WinPython as pure Python distribution is a waste.


Ash258:
I do not think that lukesampson/scoop#3466 is in master.

niheaven:
Just works fine for 'master', using old treatment for `extract_dir` and `extract_to`.

PR #101, opened 2019-06-02
add subinacl

vmvmpvm:
subinacl is a tool from Microsoft Windows Resource Kits


vmvmpvm:
Just in-case you think this is a duplicate of rktools2k3, this downloads directly from Microsoft website and is just one of the tools.

vmvmpvm:
As far as I see it is not waiting on me, please let me know if I am missing something if it is waiting on me.

PR #110, opened 2019-06-05
Prometheus: Add description and persist data and config

nightscape:


bucket/prometheus.json
```diff
@@ -15,9 +16,17 @@
         }
     },
     "bin": [
-        "prometheus.exe",
+        [
+            "prometheus.exe",
+            "prometheus",
+            "--config.file $persist_dir/prometheus.yml"
```
Ash258: You do not create this file. Does it exists inside downloaded archive?

nightscape: Yes, that file is in the archive and contains some sensible default values.

PR #112, opened 2019-06-06
helm: Add tiller.exe binary

whut:
Along with helm executable there is also tiller executable, which should also be installed.

## Rationale

Originally Helm was designed to run tiller component inside the Kubernetes clusters, but for security reasons (as then it runs with e.g. cluster-admin permissions, which means that each user running helm client in the end have cluster-admin permissions) it is better to run tiller locally. This is supported and documented at https://helm.sh/docs/using_helm/#running-tiller-locally-1. To be able to use this approach, tiller component should be installed with helm.

### DIFF
--- a/bucket/helm.json
+++ b/bucket/helm.json
@@ -9,7 +9,10 @@
         }
     },
     "extract_dir": "windows-amd64",
-    "bin": "helm.exe",
+    "bin": [
+        "helm.exe",
+        "tiller.exe"
+    ],
     "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/helm.json
```diff
@@ -9,7 +9,10 @@
         }
     },
     "extract_dir": "windows-amd64",
-    "bin": "helm.exe",
+    "bin": [
+        "helm.exe",
+        "tiller"
```
Ash258: A
```suggestion
        "tiller.exe"
```

PR #117, opened 2019-06-07
Add sox (army knife for audio processing)

boltomli:
It is a popular tool for audio processing just like ffmpeg for video (and audio, of course).


boltomli:
> Wrong URL
> Hash from sourceforge
> Autoupdate, checkver
> 
> 
> Fixed manifest

Updated per review, thanks!

PR #123, opened 2019-06-11
latex: Fix manifest and update to 2.9.7086

### LINKED ISSUE #102: latex@2.9.6942: download via aria2 failed
as the title goes。。。

maybe the reasons is：

```
The requested URL /ctan/systems/win32/miktex/setup/windows-x86/miktex-portable-2.9.6942.exe was not found on this server.
```

ps: texlive may be wanted

niheaven:
Close #102, close https://github.com/lukesampson/scoop/issues/3497

MiKTeX has 64bit installer, but with different `env_add_path`, so after a small hotfix of scoop-core, 64bit will add to the manifest. Now 32bit works fine.

### DIFF
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -5,7 +5,7 @@
         "identifier": "Freeware",
         "url": "https://miktex.org/copying"
     },
-    "url": "https://miktex.org/download/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-2.9.7086.exe",
+    "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-2.9.7086.exe",
     "hash": "1d4b3d343b86e2c0b573518969f9e3e73253ca1b08e4d65f854b09be68730c71",
     "installer": {
         "args": "--portable=\"$dir\" --unattended --private"
@@ -24,7 +24,7 @@
         "regex": "basic-miktex-([\\d.]+).exe"
     },
     "autoupdate": {
-        "url": "https://miktex.org/download/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-$version.exe",
+        "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-$version.exe",
         "hash": {
             "url": "https://miktex.org/download",
             "regex": "$basename</td>[\\S\\s]*?$sha256"
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -13,7 +13,7 @@
     "bin": [
         [
             "texmfs\\install\\miktex\\bin\\miktex-console.exe",
-            "miktex-portable",
+            "miktex",
             "--hide --mkmaps"
         ]
     ],
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -1,22 +1,33 @@
 {
     "homepage": "https://miktex.org",
-    "version": "2.9.6942",
+    "version": "2.9.7086",
     "license": {
         "identifier": "Freeware",
         "url": "https://miktex.org/copying"
     },
-    "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/miktex-portable-2.9.6942.exe#/dl.7z",
-    "hash": "20e42d52e7601f7b0abaf042408a18b39ffd84b3bb5c2a51f9b70a6555ea5897",
+    "url": "https://miktex.org/download/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-2.9.7086.exe",
+    "hash": "1d4b3d343b86e2c0b573518969f9e3e73253ca1b08e4d65f854b09be68730c71",
+    "installer": {
+        "args": "--portable=\"$dir\" --unattended --private"
+    },
+    "bin": [
+        [
+            "texmfs\\install\\miktex\\bin\\miktex-console.exe",
+            "miktex-portable",
+            "--hide --mkmaps"
+        ]
+    ],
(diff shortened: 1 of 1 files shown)


r15ch13:
Please change URL back to ftp.fau.de. miktex.org/download will redirect to HTTP mirrors and the download will fail.

niheaven:
Using a repository in Germany is somewhat unfriendly for users from other continents... Is there any other solution?

r15ch13:
https://miktex.org/pkg/repositories

niheaven:
Yes, it's based on visitor's location 😃 Okay, let it go, user could download from nearest mirror and move installer to cache.

a59:
I am sorry if I am being ignorant here; I don't know much about scoop manifests.

What I know about the MiTeX installer is that the CTAN mirrors are just copies of the files found on the main site [https://miktex.org/download]https://miktex.org/downloadl).

The Basic Installer (either 64 or 32 bit) can be used either as a general installer or as a portable installer. There is an option within the installer to choose "installer for all users" or "install for this user". Choosing "install for this user" allows the user to select a location.

If you want to install a portable version, the MiTeK website suggest you download the basic installer and rename it "miktex-portable.exe". All this seems to do is bypass the "install for all users" selection and change the default install location from "C:\Users\myname\AppData\Local\Programs\MiKTeX 2.9" to "C:/miktex-portable". 

The download basic*.exe file contains an application "miktex-setup-wizard.exe" and a lot of .tar files. I couldn't work out how to extract them using 7z which complained that it "Can not open the file as archive".

You can also download a separate command line utility called "miktexsetup-2.9.6942-x64.zip" which contains one file "miktexsetup.exe" which can be used to install everything from the command line downloading what it needs from either a remote or local repository of those .tar files.

The command "miktexsetup --verbose --local-package-repository=C:\Users\myname\miktex-repo --package-set=complete download" will download a compete set to a local directory which then can be used to install the package.

niheaven:
"https://miktex.org/download/ctan" will redirect to CTAN download mirrors "http://mirror.ctan.org" and may change to `http` protocol that leads a download failure.

Installer arguments are from installer [reference](https://docs.miktex.org/2.9/manual/setupwiz.html) and MiKTeX is installed as portable in this PR.

MattKang:
> Please change URL back to ftp.fau.de. miktex.org/download will redirect to HTTP mirrors and the download will fail.

This is not ideal. Really should be using the mirror redirects for the best server per user. The main miktex.org download link redirects fine for me

niheaven:
Without `--private`:

without-private

`64bit`: [Hotfix](https://github.com/lukesampson/scoop/pull/3503)

> MiKTeX has 64bit installer, but with different `env_add_path`, so after a small hotfix of scoop-core, 64bit will add to the manifest. Now 32bit works fine.

bucket/latex.json
```diff
@@ -1,22 +1,33 @@
...
     "license": {
         "identifier": "Freeware",
         "url": "https://miktex.org/copying"
     },
-    "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/miktex-portable-2.9.6942.exe#/dl.7z",
-    "hash": "20e42d52e7601f7b0abaf042408a18b39ffd84b3bb5c2a51f9b70a6555ea5897",
+    "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-2.9.7100.exe",
+    "hash": "893505ea259a77a8a3744b64a1975a20abc57e4d270edc2d03affad36b3427dd",
+    "installer": {
+        "args": "--portable=\"$dir\" --unattended --private"
```
Ash258: Use array

bucket/latex.json
```diff
@@ -1,22 +1,33 @@
...
+    },
+    "bin": [
+        [
+            "texmfs\\install\\miktex\\bin\\miktex-console.exe",
+            "miktex",
```
Ash258: It is miktex-console, not miktex.

Add miktex-console without renaming and then new with renaming

bucket/latex.json
```diff
@@ -1,22 +1,33 @@
...
+        "regex": "basic-miktex-([\\d.]+).exe"
     },
     "autoupdate": {
-        "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/miktex-portable-$version.exe#/dl.7z",
+        "url": "https://ftp.fau.de/ctan/systems/win32/miktex/setup/windows-x86/basic-miktex-$version.exe",
         "hash": {
-            "url": "https://miktex.org/download#portable",
-            "find": "SHA-256:</td>\\s*<td>\\s*([a-fA-F0-9]{64})"
+            "url": "https://miktex.org/download",
+            "regex": "$basename</td>[\\S\\s]*?$sha256"
```
Ash258: Use `.*?` everywhere or `[\\S\\s]*?` everywhere

bucket/latex.json
```diff
@@ -1,22 +1,33 @@
 {
     "homepage": "https://miktex.org",
-    "version": "2.9.6942",
+    "version": "2.9.7100",
```
Ash258: No description

bucket/latex.json
niheaven: It is created by MiKTeX with name 'miktex-portable', so I rename it to just 'miktex'. 'miktex-console' itself can be accessed since it's in `PATH`.

Ash258: Yeah but without parameters

niheaven: So for this one with parameters, 'miktex' is preferred, and for general usage, 'miktex-console' can be used.

Ash258: I do not agree, but whatever.

PR #135, opened 2019-06-17
autoit: Add bins, shortcuts and suggestion

niheaven:
autoit-script-editor: https://github.com/lukesampson/scoop-extras/pull/2289

Just likes DjVuLibre, it is used mainly as CLI tool, so put it here instead of `extras`.


niheaven:
Its full software name is `AutoIt v3` indeed...

> AutoIt v3 is a freeware BASIC-like scripting language designed for automating the Windows GUI and general scripting. It uses a combination of simulated keystrokes, mouse movement and window/control manipulation in order to automate tasks in a way not possible or reliable with other languages.

niheaven:
And `AutoIt v2`'s latest update is released on 18th Dec, 2002, e.g., it's been `v3` from 2002 and no one would know when to be `v4`.

PR #143, opened 2019-06-17
curl: Fix autoupdate URLs

rasa:


rasa:
Misread the log. A .tar.xz file is valid.

Issue #139, opened 2019-06-17
latex: Use updated installer


rasa:
Per https://miktex.org/download#portable, it seems we'll have to do something like:
```
miktexsetup --package-set=basic download
miktexsetup install
```
See also https://miktex.org/faq/setup-cli and https://miktex.org/howto/deploy-miktex

niheaven:
It's the same that use installer and be installed as portable, and this prevent downloading additional files that in install scripts.

rasa:
Already being done in #123

PR #146, opened 2019-06-18
Update latex.json to v2.9.7100

MattKang:
Portable MiKTeX installation has changed


niheaven:
See https://github.com/ScoopInstaller/Main/pull/123

rasa:
@MattKang Thank you for the submission, but I think it’s best to let #123 take its course first. See also #139.

PR #148, opened 2019-06-18
jetbrains-toolbox: Move to extras

Ash258:
- Add description
- Use jsonpath for checkver
- Do not install. Extraction is enough
- Create default configuration file if there is none.
    - Old implementation was completely nonsense
    - It was basically adding install_location all the time even when that file exists.
         - So it override user setting
			- This is unacceptable since it break if user change this location in GUI

Why is this manifest inside Main bucket when it is clearly GUI application with only small command line support

image


niheaven:
So why not move it to `extras`?

PR #149, opened 2019-06-18
ack: Fix autoupdate URL

rasa:
Sorry, I thought I had tested this.


niheaven:
LGTM, but why not update to 3.0.0 and rename file to just `ack` instead of `ack-single-file`?

PR #160, opened 2019-06-20
msival2: Add version 5.0.18362.1

niheaven:
One of MSI Development Tools


niheaven:
Of course we can. In fact the version number is the Windows SDK version that ships these tools (MsiVal2 and Orca), but it is not related with version number of the tool itself, just for convenience.

These tools seem not to be updated for a long time, Orca's version itself is 5.0.10011.0, and I cannot find MsiVal2's version.

niheaven:
These download URLs are cached from Windows SDK installer, and maybe next time URLs from visualstudio.microsoft.com will be used 😃

PR #168, opened 2019-06-21
use SourceForge download instead of MSYS2 repo

atomic-mattie:
`repo.msys2.org` has been down every time I've tried to use it for some
time. Use SourceForge's MSYS2 mirror instead.


atomic-mattie:
In case anyone comes by later and sees this issue closed and doesn't understand why, it appears to me that the MSYS2 repo is working now, making the issue moot.

PR #181, opened 2019-06-27
Add act 0.1.3

Ash258:


niheaven:
Why not add 32bit?

Ash258:
For some reason I assumed second binary is for arm 🤷‍♀️

PR #182, opened 2019-06-27
Add nyagos 4.4.4_2

ironsand:
I create a bucket for `nyagos`.

This is short description for NYAGOS from github page.
> NYAGOS is the commandline-shell written with the Programming Language GO and Lua.


ironsand:
Thanks for the review! I changed the bucket file.

Issue #150, opened 2019-06-18
Outstanding Excavator issues


rasa:
Here's a list of the outstanding Excavator issues found in log https://scoop.r15.ch/main/mud-20200203-010001.log at [https://scoop.r15.ch/main/](https://scoop.r15.ch/main/?sort=time&order=desc):

## Needing triage (newest to oldest):

None.

## Fixable:

None.

## Being fixed:

1. apimtemplate: 0.3-gamma (scoop version is 0.2-beta) autoupdate available. Fixed in #646. Thanks @rasa!

## Version parsing issues:

None.

## Missing binaries:

1. [pngquant](../blob/master/bucket/pngquant.json): couldn't match ` 'Binary for Windows.*?([\d.]+)'` in https://pngquant.org/ . See #299.
1. azure-functions-core-tools: 3.0.2106 (scoop version is 2.7.2100) autoupdate available. ERROR Could not update azure-functions-core-tools 32bit. Missing binaries at https://github.com/Azure/azure-functions-core-tools/releases
1. azure-ps: couldn't match `'download/v?(?<release>[\w-.]+)/Az-Cmdlets-([\d.]+)-x64\.msi'` in https://github.com/Azure/azure-powershell/releases/latest
1. ccl: 1.11.8 (scoop version is 1.11.5) autoupdate available. Missing binaries at https://github.com/Clozure/ccl/releases
1. jdupes: 1.14.0 ~~1.13.3~~ (scoop version is 1.13.2) autoupdate available. ERROR Could not update jdupes 32bit. Missing binary at https://github.com/jbruchon/jdupes/releases
1. rethinkdb: 2.4.0 (scoop version is 2.3.6) autoupdate available. https://rethinkdb.com/docs/install/windows/ lists https://download.rethinkdb.com/windows/rethinkdb-2.4.0.zip which is 404.
1. dig: 9.14.9 (scoop version is 9.14.8) autoupdate available. Missing 32bit build. See https://ftp.isc.org/isc/bind9/9.14.9/
1. pdf2djvu: 0.9.14 (scoop version is 0.9.13) autoupdate available. See https://github.com/jwilk/pdf2djvu/releases
1. [artifact](../blob/master/bucket/artifact.json): Missing windows builds, See https://github.com/vitiral/artifact/releases. Fix coming soon in vitiral/artifact#246, vitiral/artifact#271.

## Network issues:

None.

## Upstream issues:

[...]

PR #186, opened 2019-06-28
dosbox: update to 0.74-3 & add autoupdate

hoilc:
#150


hoilc:
> Is installation needed?
> 
> image

I think it's unnecessary.

> https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/branches/0_74_3/scripts/dosbox-installer.nsi

It seems that the installer does nothing special.

> https://www.vogons.org/viewtopic.php?f=31&t=57052#wrap

And the minimal set which dosbox needs is said to be
```
dosbox.exe
SDL.dll
SDL_net.dll
dosbox.conf
```

It could be modified in another PR. 

Also it‘s ideal that someone who is familiar with dosbox could change and test it.

PR #189, opened 2019-06-28
edgedriver: Update to version 82.0.449.0 (Chromium)

niheaven:
Legacy version has been moved to `versions`.


Ash258:
Edge with chromimum is still in development and available for insiders (https://www.microsoftedgeinsider.com/en-us/) I do not think this is good move to replace current version with future version.

niheaven:
So I put legacy version in `version`, and add description here to figure out it is Chromium version.

I'll add a note in this manifest.

Ash258:
Yeah, but this should be merged as soon as chromium edge is realease globally as stable for all users and normal edge is retired.

niheaven:
@Ash258 Edge is now stable: https://www.microsoft.com/en-us/edge

PR #191, opened 2019-06-28
ruby: Add au.hash and fix checkver

niheaven:


h404bi:
I think `rubyinstaller.org/downloads/archives/` is the same as `github.com/oneclick/rubyinstaller2`. Its data is hosted on GitHub pages. Better to change `checkver` to monitor `https://github.com/oneclick/rubyinstaller.org-website/blob/master/_data/downloads.yaml`

PR #195, opened 2019-06-30
radare2: Fix 'extract_dir'

vii1:
Directory struct changed in 3.6.0

### DIFF
--- a/bucket/radare2.json
+++ b/bucket/radare2.json
@@ -12,19 +12,19 @@
     },
     "bin": [
         [
-            "bin/radare2.exe",
+            "bin\\radare2.exe",
             "r2"
         ],
-        "bin/r2agent.exe",
-        "bin/rabin2.exe",
-        "bin/radare2.exe",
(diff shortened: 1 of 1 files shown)


bucket/radare2.json
```diff
@@ -7,20 +7,20 @@
...
-        "rabin2.exe",
-        "radare2.exe",
-        "radiff2.exe",
-        "rafind2.exe",
-        "ragg2.exe",
-        "rahash2.exe",
-        "rarun2.exe",
-        "rasm2.exe",
-        "rax2.exe"
+        "bin/r2agent.exe",
```
niheaven:
```suggestion
        [
            "bin/radare2.exe",
            "r2"
        ],
        "bin/r2agent.exe",
```

niheaven: Just add `r2` as alias.

Ash258: Use backslashes

Issue #205, opened 2019-07-03
Could you please add ccl (Clozure Common Lisp)


priyadarshan:
[Clozure Common Lisp](https://ccl.clozure.com/) is an advanced and free Common Lisp implementation.

Windows XP and later (x86, x86-64) executable is here:
https://github.com/Clozure/ccl/releases.

Filename says x86, but zip file includes 64-bit.

That is a portable install.

Thank you!

[Directed here from lukesampson/scoop-extras/issues/2401]

priyadarshan:
I can confirm it works nicely. Thank you so much.

PR #206, opened 2019-07-03
Add ccl 1.11.5

linsui:
Close https://github.com/ScoopInstaller/Main/issues/205

@priyadarshan Should I rename those bin or add more  path to env?

### DIFF
--- a/bucket/ccl.json
+++ b/bucket/ccl.json
@@ -1,0 +1,20 @@
+{
+    "version": "1.11.5",
+    "homepage": "https://ccl.clozure.com/",
+    "description": "Clozure Common Lisp",
+    "license": "Apache-2.0",
+    "url": "https://github.com/Clozure/ccl/releases/download/v1.11.5/ccl-1.11.5-windowsx86.zip",
+    "hash": "da5d8830e38742a5e76dc09a5fdebf8cb8a09a6fee87d7919ffd67ae2584cdd4",
+    "extract_dir": "ccl",
+    "env_add_path": "scripts",
+    "bin": [
+        "wx86cl.exe",
+        "wx86cl64.exe"
+    ],
+    "checkver": {
+        "github": "https://github.com/Clozure/ccl"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Clozure/ccl/releases/download/v$version/ccl-$version-windowsx86.zip"
+    }
+}
--- a/bucket/ccl.json
+++ b/bucket/ccl.json
@@ -9,6 +9,8 @@
     "architecture": {
         "64bit": {
             "bin": [
+                "wx86cl.exe",
+                "wx86cl64.exe",
                 [
                     "wx86cl.exe",
                     "ccl32"
--- a/bucket/ccl.json
+++ b/bucket/ccl.json
@@ -6,11 +6,32 @@
     "url": "https://github.com/Clozure/ccl/releases/download/v1.11.5/ccl-1.11.5-windowsx86.zip",
     "hash": "da5d8830e38742a5e76dc09a5fdebf8cb8a09a6fee87d7919ffd67ae2584cdd4",
     "extract_dir": "ccl",
-    "env_add_path": "scripts",
-    "bin": [
-        "wx86cl.exe",
-        "wx86cl64.exe"
-    ],
+    "architecture": {
+        "64bit": {
+            "bin": [
+                [
+                    "wx86cl.exe",
+                    "ccl32"
+                ],
+                [
+                    "wx86cl64.exe",
+                    "ccl"
+                ]
+            ]
+        },
+        "32bit": {
+            "bin": [
+                [
+                    "wx86cl.exe",
+                    "ccl"
+                ],
+                [
+                    "wx86cl64.exe",
+                    "ccl64"
+                ]
+            ]
+        }
+    },
     "checkver": {
         "github": "https://github.com/Clozure/ccl"
     },
--- a/bucket/ccl.json
+++ b/bucket/ccl.json
@@ -23,6 +23,8 @@
         },
         "32bit": {
             "bin": [
+                "wx86cl.exe",
+                "wx86cl64.exe",
                 [
                     "wx86cl.exe",
                     "ccl"


priyadarshan:
It seems scoop prefers "latest release" (2017), vs "latest pre-release" right?

I am not fluent with scoop dev-side, but it seems there is no need to rename those bin or add more path to env.

I can definitely test once merged.

Thank you so much for considering this, and so quickly.

bucket/ccl.json
```diff
@@ -0,0 +1,20 @@
...
+    "homepage": "https://ccl.clozure.com/",
+    "description": "Clozure Common Lisp",
+    "license": "Apache-2.0",
+    "url": "https://github.com/Clozure/ccl/releases/download/v1.11.5/ccl-1.11.5-windowsx86.zip",
+    "hash": "da5d8830e38742a5e76dc09a5fdebf8cb8a09a6fee87d7919ffd67ae2584cdd4",
+    "extract_dir": "ccl",
+    "env_add_path": "scripts",
+    "bin": [
+        "wx86cl.exe",
+        "wx86cl64.exe"
```
Ash258: What is different in these files??
Is it used to produce different compiled binary when talking about architectures? Like 64.exe produce only 64bit output?

bucket/ccl.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "1.11.5",
+    "homepage": "https://ccl.clozure.com/",
+    "description": "Clozure Common Lisp",
+    "license": "Apache-2.0",
+    "url": "https://github.com/Clozure/ccl/releases/download/v1.11.5/ccl-1.11.5-windowsx86.zip",
+    "hash": "da5d8830e38742a5e76dc09a5fdebf8cb8a09a6fee87d7919ffd67ae2584cdd4",
+    "extract_dir": "ccl",
+    "env_add_path": "scripts",
```
Ash258: >  In the scripts/ directory of the ccl directory, there are two files named ccl and ccl64. Copy these files into /usr/local/bin or some other directory that is on your path, and then edit them so that the value of CCL_DEFAULT_DIRECTORY is your ccl directory. You can then start up the lisp by typing ccl or ccl64.

If there are only 2 files, why not add them as bins?

Ash258: Or will this folder grow in time?

bucket/ccl.json
priyadarshan: > Like 64.exe produce only 64bit output?

Yes, exactly.

They are exactly the same, beside one is 64-bit, the other 32-bit. The difference is that the Lisp developer may develop an application with 64-bit, and deploy (ie create an exe file) both 64-bit and 32-bit. But to create one or the other one needs the proper corresponding binary.

bucket/ccl.json
priyadarshan: As far as I know, more files may be added in the future, like for example, right now ther are also some shell scripts.

linsui: Ok, I'll add them as bin

bucket/ccl.json
Ash258: @priyadarshan Thanks.

@linsui Keep it as it is

bucket/ccl.json
priyadarshan: Thank you so much.

linsui:
@Ash258 Is this all right?

Ash258:
> Ok, I'll add them as bin

https://github.com/ScoopInstaller/Main/pull/206#discussion_r299993159

linsui:
> As far as I know, more files may be added in the future, like for example, right now ther are also some shell scripts.

Should I keep `env_add_path` or just add them into `bin` ?

Ash258:
I cannot tell as I am not and never will be user of this application.

Scripts, except `ccl` looks like helpers for development.

linsui:
> It can also be handy to run Clozure CL straight from a terminal prompt. In the scripts/ directory of the ccl directory, there are two files named ccl and ccl64. Copy these files into /usr/local/bin or some other directory that is on your path, and then edit them so that the value of CCL_DEFAULT_DIRECTORY is your ccl directory. You can then start up the lisp by typing ccl or ccl64.

As ccl or ccl64 is a bash script, 

> # Change the definition of CCL_DEFAULT_DIRECTORY below to refer to
> # your Clozure CL installation directory.  The lisp will use this
> # environment variable to set up translations for the CCL: logical
> # host.

> # Any definition of CCL_DEFAULT_DIRECTORY already present in the
> # environment takes precedence over definition made below.

I guess it makes no sense to add them into path and we should just use wx86cl.exe and wx86cl64.exe

bucket/ccl.json
```diff
@@ -6,11 +6,32 @@
...
     "hash": "da5d8830e38742a5e76dc09a5fdebf8cb8a09a6fee87d7919ffd67ae2584cdd4",
     "extract_dir": "ccl",
-    "env_add_path": "scripts",
-    "bin": [
-        "wx86cl.exe",
-        "wx86cl64.exe"
-    ],
+    "architecture": {
+        "64bit": {
+            "bin": [
```
Ash258:
```suggestion
            "bin": [
                "wx86cl.exe",
                "wx86cl64.exe",
```

bucket/ccl.json
```diff
@@ -6,11 +6,32 @@
...
+                    "ccl32"
+                ],
+                [
+                    "wx86cl64.exe",
+                    "ccl"
+                ]
+            ]
+        },
+        "32bit": {
+            "bin": [
```
Ash258:
```suggestion
            "bin": [
                "wx86cl.exe",
                "wx86cl64.exe",
```

bucket/ccl.json
linsui: `ccl` is handy and I guess on linux users can input ccl to start ccl. Just like this

批注 2019-07-07 200318

`script/ccl` is just a  script call wx86cl.exe on different system.
```
`exec ${CCL_DEFAULT_DIRECTORY}/${OPENMCL_KERNEL} "$@"
```

Ash258: Did i said to remove the renames??

I just added files without renames.

Ash258: User who requested this package told you to keep it without renaming. And you still renamed it. 

So keep at least original names and add whatever you want in additional.

linsui: I get it.

PR #210, opened 2019-07-04
monero-cli: Add version 0.14.1.0

niheaven:
For 32bit

### DIFF
--- a/bucket/monero-cli.json
+++ b/bucket/monero-cli.json
@@ -49,7 +49,7 @@
         },
         "hash": {
             "url": "https://getmonero.org/downloads/hashes.txt",
-            "find": "$basename.*([a-fA-F0-9]{64})"
+            "regex": "$basename,\\s+$sha256"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


bucket/monero-cli.json
```diff
@@ -0,0 +1,55 @@
...
+            "64bit": {
+                "url": "https://dlsrc.getmonero.org/cli/monero-win-x64-v$version.zip"
+            },
+            "32bit": {
+                "url": "https://dlsrc.getmonero.org/cli/monero-win-x86-v$version.zip"
+            }
+        },
+        "hash": {
+            "url": "https://getmonero.org/downloads/hashes.txt",
+            "find": "$basename.*([a-fA-F0-9]{64})"
```
Ash258:
```suggestion
            "regex": "$basename,\\s+$sha256"
```

niheaven: Why `,\\s+`?

Ash258: `monero-win-x64-v0.14.1.0.zip, ca4367c565026ea23bf90180cdd750a1d6adfab36ee19435e91ed740a369a6fc`

niheaven: `.*` works, too. But okay, let's use exactly match

Ash258: That is nonsense to use greedy `.*` for two characters

PR #212, opened 2019-07-04
jx: updated executable name in shim definition

kapoun:
Starting from version 2.0.340, the executable is named just `jx.exe`, not `jx-windows-amd64.exe`. Due to this change, a shim for the executable is not correctly created, which results in an incomplete installation. This pull request should fix this behaviour.


bucket/jx.json
```diff
@@ -11,7 +11,7 @@
     },
     "bin": [
```
Ash258:
```suggestion
    "bin": "jx.exe",
```

PR #224, opened 2019-07-06
m4: Add version 1.4.14-1

issaclin32:
Homepage for **m4**: http://gnuwin32.sourceforge.net/packages/m4.htm
~Homepage for **regex**: http://gnuwin32.sourceforge.net/packages/regex.htm~

~Both packages are from `GNUWin32`, and `m4` is dependent on "regex2.dll" from the `regex` package.~

Modified according to suggestion from niheaven:
Keep `m4` only, and download its dependency files (m4-1.4.14-1-dep.zip) along.

Notes:
~* The developer provides MD5 checksum, but Scoop's `autoupdate` can automatically fetch SHA-1 HASH from Sourceforge repos, so there is no need to add "hash" descriptor to the autoupdate section.
(another example: https://github.com/ScoopInstaller/Main/blob/master/bucket/which.json)~

* Remove autoupdate section since `GNUWin32` is frozen since 2010.

~Possible Issues:~
~* Is it allowed for packages to contain **only DLL files**? 
`regex2.dll` itself can also be updated by GNUWin32's developers. (Therefore, its version/url may change in the future.) 
I think the best approach is to treat `regex2.dll` as an individual package.~

~* `regex` is a pretty generic package name. Should I change it to a more specific name such as `gnuwin32-regex`?~


niheaven:
- Only `m4` could be added, and use two download urls that one for `m4` while another for `m4-dep` (`regex`)
- Delete `checkver` and `autoupdate` since GnuWin32 project is frozen from 2010
- Change download urls to general domain

<details>
<summary>Fixed Manifest</summary>

```json
{
    "homepage": "http://gnuwin32.sourceforge.net/packages/m4.htm",
    "version": "1.4.14-1",
    "license": "GPL-3.0-or-later",
    "description": "UNIX macro processor (from GNUWin32)",
    "url": [
        "https://downloads.sourceforge.net/project/gnuwin32/m4/1.4.14-1/m4-1.4.14-1-bin.zip",
        "https://downloads.sourceforge.net/project/gnuwin32/m4/1.4.14-1/m4-1.4.14-1-dep.zip"
    ],
    "hash": [
        "4dd8f36ae83dae682cf442a3544608f53cd4ea021f12563d701714f8ced524a6",
        "ee507393612a46209f5d28ec50e2d2aa1b5a88f5958dca9ce3c9f04174e54407"
    ],
    "bin": "bin\\m4.exe"
}
```

</details>

PR #229, opened 2019-07-09
Add bison, byacc, btyacc (from GNUWin32) and winflexbison

issaclin32:
Notes:
* `GNUWin32` has been frozen since 2010. Therefore **checkver** and **autoupdate** is not used in this package.

### DIFF
--- a/bucket/bison.json
+++ b/bucket/bison.json
@@ -1,0 +1,15 @@
+{
+    "homepage": "http://gnuwin32.sourceforge.net/packages/bison.htm",
+    "version": "2.4.1",
+    "license": "GPL-3.0-or-later",
+    "description": "Yacc-compatible parser generator (from GNUWin32)",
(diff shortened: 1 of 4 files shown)


r15ch13:
Would https://github.com/lexxmark/winflexbison be a good addition to that? Stumbled over it a couple of days ago. (don't know what it does, though :grin:)

issaclin32:
> Would https://github.com/lexxmark/winflexbison be a good addition to that? Stumbled over it a couple of days ago. (don't know what it does, though 😁)

Thanks for the suggestion. 
I named the shim-exes as `bison` and `flex` (rather than `win_bison` and `win_flex`) according to [this usage example](https://github.com/meyerd/flex-bison-example). 

This will cause duplicate names when the user installs both `bison` and `winflexbison` packages. I assume this is acceptable.

PR #230, opened 2019-07-10
BIND: Add version 9.14.3

linsui:
`dig` has been in scoop and it's a part of `bind` . Should I remove `dig.json` ? There is a `vcredist` in dir, could use it or must use `suggest` ?

### DIFF
--- a/bucket/dig.json
+++ b/bucket/dig.json
@@ -1,34 +1,12 @@
 {
-    "homepage": "https://www.isc.org/",
+    "version": "nightly",
+    "homepage": "https://www.isc.org/bind/",
     "description": "dig (domain information groper) is a flexible tool for interrogating DNS name servers",
     "license": "MPL-2.0",
-    "version": "9.14.3",
-    "architecture": {
-        "64bit": {
-            "url": "https://ftp.isc.org/isc/bind9/9.14.3/BIND9.14.3.x64.zip",
-            "hash": "9182c6593d11581909d12706e663252c357696def102b260da02c98ec1ac7b86"
-        },
-        "32bit": {
-            "url": "https://ftp.isc.org/isc/bind9/9.14.3/BIND9.14.3.x86.zip",
-            "hash": "1ded768b090e633b1930f99e457c1f73b3955614aff621ec13ea12c186884472"
-        }
-    },
-    "bin": "dig.exe",
-    "checkver": {
-        "url": "https://www.isc.org/downloads/",
-        "re": "Current-Stable[\\S\\s]*?BIND(.+)\\.x64\\.zip\""
-    },
-    "autoupdate": {
-        "architecture": {
-            "64bit": {
-                "url": "https://ftp.isc.org/isc/bind$majorVersion/$version/BIND$version.x64.zip"
-            },
-            "32bit": {
-                "url": "https://ftp.isc.org/isc/bind$majorVersion/$version/BIND$version.x86.zip"
-            }
-        }
-    },
-    "suggest": {
-        "vcredist": "extras/vcredist2012"
-    }
+    "depends": "bind",
+    "url":"https://www.isc.org/bind/",
+    "notes": [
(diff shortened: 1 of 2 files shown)


linsui:
@rasa Should I remove `dig.json` as it's just a part of bind?

rasa:
> @rasa Should I remove `dig.json` as it's just a part of bind?

Removing it disrupts users. Perhaps we rewrite it to simply depend on bind, and add a note that it’s deprecated, and will be removed in the future?

Perhaps something like (untested)?:

```json
{
    "homepage": "https://www.isc.org/",
    "description": "dig (domain information groper) is a flexible tool for interrogating DNS name servers",
    "license": "MPL-2.0",
    "version": "9.14.3",
    "checkver": {
        "url": "https://www.isc.org/downloads/",
        "re": "Current-Stable[\\S\\s]*?BIND(.+)\\.x64\\.zip\""
    },
    "autoupdate": {
        "url": "https://example.com/"
    },
    "depends": "bind",
    "url": "https://example.com/",
    "hash": "",
    "notes": [
        "The dig package has been deprecated, and will be removed in the future.",
        "Please use the bind package instead."
    ]
}

**Comment by linsui** [comment]
Good idea. 👍 I'll test it.

File: bucket/dig.json [hunk 1]
```diff
@@ -1,34 +1,12 @@
...
-            "32bit": {
-                "url": "https://ftp.isc.org/isc/bind$majorVersion/$version/BIND$version.x86.zip"
-            }
-        }
-    },
-    "suggest": {
-        "vcredist": "extras/vcredist2012"
-    }
+    "depends": "bind",
+    "url":"https://www.isc.org/bind/",
```
Ash258: I do not understand.

Instead of keeping funtional manifest and adding notes you will break whole manifest.

rasa: Nothing broke. I tested it locally, and it worked fine. What do you think is broke? Is there a better way to do this? Should we just leave the dig package alone, and let there be two possible digs, if they install both bind and dig?

linsui: It works fine for me. If you have suggestions, please tell me.

Ash258: This manifest is completely useless now.
There is blank URL with just webpage. That does not make sense. Remove whole manifest or keep it working as berfore and add notes about bind

linsui: I think it's acceptable as a temporary solution. But I find another problem. Scoop won't check `depend` when update so `bind` won't be installed automatically when update dig. Is it a bug?

linsui:
We really need some way to migrate to a new package painless or just show a note when manifest is removed.

bucket/dig.json
rasa: @Ash258 The URL is required (at least according to the [wiki](https://github.com/lukesampson/scoop/wiki/App-Manifests#required-properties), but not the [schema](https://github.com/lukesampson/scoop/blob/b9b48182b4d921aa92e9b3e53ab2d2c376d467d7/schema.json#L526)) but is unneeded in this context, as we just are installing the `depends app. Perhaps it should be clear that that’s the case:
```
"url": "https://example.com/#url-parameter-required-but-is-not-needed-for-this-metapackage”
```
Or is the URL param no longer needed?

rasa: @Ash258 Nevermind. @linsui has implemented your suggestion, which is fine.

Ash258: It is needed.

But what is wrong with fully working `dig.json` with just notes about adopting to BIND??
Keeping manifest which does not install anything is completely nonsense in official bucket.

rasa: @Ash258 Your proposal is a fine solution, but my proposed dig manifest installed dig by depending on bind, which has dig. In fact, both dig and bind now install the exact same download. The only difference is dig exposes only dig.exe via a shim, and bind exposes all ~15 .exes in the download via an `env_add_path`. Sorry for the confusion.

PR #231, opened 2019-07-10
Syncthing: portable

linsui:


rasa:
@linsui Why not add the logic from syncthing-portable to syncthing, as they share the same binary?

linsui:
Just lik this?

```
    "bin": [
        "syncthing.exe",
        [
            "syncthing.exe",
            "syncthing-portable",
            "-home \"$dir\\config\""
        ]
    ],
```

Or replace the bin?

rasa:
Replace the bin, and add the persist stuff too.

Ash258:
Then you have to handle transition of data for current users.

If you change it to portable then all users data stored in appdata (or whenever) will not be used, so if you want to replace it then add installer script with moving this file.

linsui:
Yeah, I'll do it later...

bucket/syncthing.json
```diff
@@ -14,7 +15,20 @@
...
+            "syncthing.exe",
+            "syncthing",
+            "-home \"$dir\\config\""
+        ]
+    ],
+    "persist": "config",
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\config\") -and (Test-Path \"$Env:LocalAppdata\\Syncthing\")) {",
+        "    New-Item \"$dir\\config\" -ItemType 'Directory' -Force | Out-Null",
+        "    Copy-Item \"$Env:LocalAppdata\\Syncthing\\*\" \"$dir\\config\" -Exclude LOCK -Recurse -Force",
```
rasa: Should this be Move-Item or Cooy-Item?

linsui: I prefer to copy them because I don't want to break other thing. If users install syncthing and uninstall -p , they won't lose anything.

PR #232, opened 2019-07-10
Yarn: Re-Add Prefix Global Directory Argument

JustinGrote:
"Prefix" was changed to "Global-Folder" but "Prefix" is still required for global tools to set up their commands into the correct location rather than localappdata. Example is azure-functions-core-tools.

Both options can be specified concurrently and safely.


rasa:
I'm not a yarn user, but these changes make sense.

PR #249, opened 2019-07-17
Tor: Move from extra, add 64bit and portable

linsui:
https://github.com/lukesampson/scoop-extras/pull/2471

### DIFF
--- a/bucket/tor.json
+++ b/bucket/tor.json
@@ -26,10 +26,11 @@
         "torrc"
     ],
     "pre_install": [
+        "$current = current_dir $dir",
         "$content = @(",
-        "    \"DataDirectory `\"$dir\\Data\\Data`\"\"",
-        "    \"GeoIPFile `\"$dir\\Data\\Tor\\geoip`\"\"",
-        "    \"GeoIPv6File `\"$dir\\Data\\Tor\\geoip6`\"\"",
+        "    \"DataDirectory `\"$current\\Data\\Data`\"\"",
+        "    \"GeoIPFile `\"$current\\Data\\Tor\\geoip`\"\"",
+        "    \"GeoIPv6File `\"$current\\Data\\Tor\\geoip6`\"\"",
         ")",
         "if (!(Test-Path \"$persist_dir\\torrc\")) {",
         "    New-Item -Force -Path \"$dir\" -Name torrc | Out-Null",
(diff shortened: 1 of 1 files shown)


linsui:
Why `$dir` in `pre_install` is `tor/0.4.0.5` ? I expect it is `C:\Users\username\scoop\apps\$app\current` as doc says.

bucket/tor.json
```diff
@@ -0,0 +1,60 @@
...
+            "hash": "76761d916772f6c5c3c43dbe4d9ca846fe6526fcbe8974b88b11a00a196e609d"
+        },
+        "32bit": {
+            "url": "https://dist.torproject.org/torbrowser/8.5.4/tor-win32-0.4.0.5.zip",
+            "hash": "8d411a394597bab5ec6fe446a2cda21279c830c5fdd19c3de82aa1028ffd7e96"
+        }
+    },
+    "bin": [
+        [
+            "Tor/tor.exe",
```
Ash258: Use backslashes

Ash258:
I do not know what variable you were trying to use, but $dir is working as expected.

image

bucket/tor.json
```diff
@@ -0,0 +1,60 @@
...
+        "torrc"
+    ],
+    "pre_install": [
+        "$content = @(",
+        "    \"DataDirectory `\"$dir\\Data\\Data`\"\"",
+        "    \"GeoIPFile `\"$dir\\Data\\Tor\\geoip`\"\"",
+        "    \"GeoIPv6File `\"$dir\\Data\\Tor\\geoip6`\"\"",
+        ")",
+        "if (!(Test-Path \"$persist_dir\\torrc\")) {",
+        "    New-Item -Force -Path \"$dir\" -Name torrc | Out-Null",
```
Ash258: You do not need to create file. Just add `-Force` to Add-Content

linsui:
I expect it is the `current` dir but I get `0.4.0.5` in `torrc` .

Ash258:
That is correct.

linsui:
Then how can I get the `current` path? If the `0.4.0.5` is write into `torrc` and be persisted, it can't work after update.

Ash258:
In `pre_install` there is no current folder. Current folder is created after installer

Ash258:
`$current = current_dir $dir`

linsui:
I got. Thanks!

PR #252, opened 2019-07-18
Refactor perl manifest to use bin

DaanDeMeyer:
Currently, every executable under the moon is available via the PATH when installing the perl manifest. This creates issues when compiling C/C++ code since build tools find the compilers/linkers included in perl manifest and decide to use those over other installed compilers.

I changed the manifest to only add perl.exe and wperl.exe to the PATH. It is extremely likely that there are other useful executables buried in the perl bin directory but it seems like a reasonable approach to add these on a as-needed basis when issues are reported by users who expect some binary to be available from the perl manifest (compared to polluting the PATH of everyone who installs the perl manifest only to have perl.exe available).

This is a breaking change but I feel the current situation is rather excessive and should be fixed.

### DIFF
--- a/bucket/perl.json
+++ b/bucket/perl.json
@@ -19,10 +19,9 @@
         "## NOTE: conversion to byte[] avoids adding an extra trailing newline to the output file",
         "[System.IO.File]::WriteAllBytes(\"$dir\\portable.perl\", ([byte[]][char[]]((Get-Content -Raw \"$dir\\portable.perl\") -replace \"(?ms)^HomeDir:.*?^(?=\\S)\",\"\")))"
     ],
-    "env_add_path": [
-        "perl\\site\\bin",
-        "perl\\bin",
-        "c\\bin"
+    "bin": [
+        "perl\\bin\\perl.exe",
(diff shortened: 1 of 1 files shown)


niheaven:
Even `cpan` is excluded since it's in `bin`... For those who don't want these bins they could manually move down Perl's path in `$PATH`, totally remove them is a worse choice.

DaanDeMeyer:
I just started with only perl.exe and wperl.exe because I'm not at all familiar with Perl. Of course, cpan and other tools can be added as well if its a widely used tool within perl. The idea is just to just not pollute the PATH with tools that are not used by anyone.

I'll update the pr to add cpan and related stuff to "bin" as well.

DaanDeMeyer:
I added cpan to the list of binaries and added perl/site/bin back to the PATH so that perl binaries installed explicitly by the user via cpan are available from the PATH (which a user would reasonably expect).

DaanDeMeyer:
Ping, if needed "perl\\bin" can be added back to the PATH as well, but I really think "c\\bin" should not be added to the PATH.

DaanDeMeyer:
Ping

PR #253, opened 2019-07-20
dotnet-sdk: include env_add_path and required dependencies

myty:
Two changes in this PR:
1) I was getting an error in vs code omnisharp and it turns out the sdk root folder needs to be on the path.
2) According to the .NET Core manual installation instructions, https://docs.microsoft.com/en-us/dotnet/core/windows-prerequisites?tabs=netcore2x#net-core-dependencies, [extras/vcredist2015](https://github.com/lukesampson/scoop-extras/blob/master/bucket/vcredist2015.json) should be included as a dependency.  I wasn't sure if this should be a "depends" or a "suggest" in the manifest.

### DIFF
--- a/bucket/vim-nightly.json
+++ b/bucket/vim-nightly.json
@@ -2,28 +2,28 @@
     "homepage": "https://www.vim.org",
     "description": "A highly configurable text editor for efficiently creating and changing any kind of text.",
     "license": "Vim",
-    "version": "8.2.0096",
+    "version": "8.2.0100",
     "architecture": {
         "64bit": {
             "url": [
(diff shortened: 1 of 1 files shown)


bucket/dotnet-sdk.json
```diff
@@ -14,6 +14,8 @@
         }
     },
     "bin": "dotnet.exe",
+    "depends": "extras/vcredist2015",
```
Ash258: Vcredists needs to be only suggested

```suggestion
    "suggest": {
        "Visual C++ Redistributable 2015": "extras/vcredist2015"
    },
```

myty: Cool, was thinking that might be the recommended way to do it. 👍

PR #257, opened 2019-07-22
zstd: Add extract_dir, remove shortcut

### LINKED ISSUE #256: zstd@1.4.1: Can't shim 'zstd.exe': File doesn't exist.
It seems like version 1.4.1 starts to put everything into a new subfolder.
The archive for the 64 Bit build now has everything inside a `zstd-v1.4.1-win64` folder.

rasa:
Fixes #256 
It's a console app, so a shortcut is meaningless.


rasa:
Installed fine locally.

PR #260, opened 2019-07-23
nmap: Extract and install npcap

linsui:
Now npcap will be installed in nmap directory. Due to unknown reasons, uninstall will fail and need to uninstall manaully.

### DIFF
--- a/bucket/nmap.json
+++ b/bucket/nmap.json
@@ -1,18 +1,26 @@
 {
-    "homepage": "https://nmap.org",
     "version": "7.70",
-    "license": "https://svn.nmap.org/nmap/COPYING",
-    "url": "https://nmap.org/dist/nmap-7.70-setup.exe",
+    "description": "A free and open source utility for network discovery and security auditing.",
+    "homepage": "https://nmap.org",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://svn.nmap.org/nmap/COPYING"
+    },
+    "url": "https://nmap.org/dist/nmap-7.70-setup.exe#/dl.7z",
     "hash": "7828197e66f5f6efb0c89d21cdf3c34769da525437147eea72cf713a6d4e7a11",
-    "extract_dir": "nmap",
+    "pre_install": [
+        "Move-Item \"$dir\\npcap*.exe\" \"$dir\\npcap.exe\"",
+        "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall.exe\", \"$dir\\vcredist*\" -Recurse -Force"
+    ],
     "installer": {
+        "file": "npcap.exe",
         "args": [
             "/S",
-            "/D=$dir"
+            "/D=$dir\\npcap"
         ]
     },
     "uninstaller": {
-        "file": "Uninstall.exe",
+        "file": "npcap\\Uninstall.exe",
         "args": "/S"
     },
     "env_add_path": "bin",
@@ -31,9 +39,12 @@
     ],
     "checkver": {
         "url": "https://nmap.org/download.html",
-        "re": "nmap-([\\d.]+)-setup.exe"
+        "regex": "nmap-([\\d.]+)-setup"
     },
     "autoupdate": {
-        "url": "https://nmap.org/dist/nmap-$version-setup.exe"
+        "url": "https://nmap.org/dist/nmap-$version-setup.exe#/dl.7z"
+    },
+    "suggest": {
+        "vcredist": "extras/vcredist2013"
     }
 }


bucket/nmap.json
```diff
@@ -1,18 +1,26 @@
 {
-    "homepage": "https://nmap.org",
     "version": "7.70",
-    "license": "https://svn.nmap.org/nmap/COPYING",
-    "url": "https://nmap.org/dist/nmap-7.70-setup.exe",
+    "description": "A free and open source utility for network discovery and security auditing.",
+    "homepage": "https://nmap.org",
+    "license": {
+        "identifier": "Freeware",
```
rasa: GPL-2.0-Only per https://svn.nmap.org/nmap/COPYING

linsui: It's not only a GPL 2, should I change license to GPL-2.0-only or only chsnge the identifier and keep the url?

linsui:
The uninstall need privileges. nmap uninstaller won't uninstall npcap so there is no problem before. By the way, could anyone have a look at https://github.com/lukesampson/scoop/issues/3567 ? With sudo it works. Will it be fixed or we need to use sudo instead?

bucket/nmap.json
rasa: You can leave the URL:
```
"license": {
"identifier": "GPL-2.0-Only",
"url": ""
}

**Comment by rasa** [comment]
In my opinion, if an app needs elevation to install/uninstall, it should depend on sudo, and then use it to install the app. See https://github.com/lukesampson/scoop-extras/blob/e73c63bfbe3f369a126c4c74fd9baf8e7ff08281/bucket/pritunl-client.json#L16 for an example/

**Comment by niheaven** [comment]
See https://github.com/lukesampson/scoop-extras/pull/2464, you could use `-RunAs`.

**Comment by linsui** [comment]
@rasa @niheaven Thanks. What's the suggested way?

**Comment by niheaven** [comment]
I suggest `-RunAs` that needn't `depends: sudo`, and please check if works.

But this param was fixed in https://github.com/lukesampson/scoop/pull/3547 and not be merged into `scoop/master` yet, so you might halt your PR and wait for merging (has been merged into `develop`)

**Comment by linsui** [comment]
@niheaven It works well, thanks. If it can be write as 

```
    {
        "file": "Uninstall.exe",
        "args": "/S",
        "admin": true
    }
```

it would be better 😂

**Comment by niheaven** [comment]
Wow, right, I'll do this works in https://github.com/lukesampson/scoop/pull/3502

**Comment by linsui** [comment]
Please review.

File: bucket/nmap.json [hunk 2]
```diff
@@ -4,19 +4,19 @@
...
+    "pre_install": [
+        "Move-Item \"$dir\\npcap*.exe\" \"$dir\\npcap.exe\"",
+        "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\Uninstall.exe\", \"$dir\\vcredist*\" -Recurse -Force"
+    ],
     "installer": {
-        "args": [
-            "/S",
-            "/D=$dir"
-        ]
+        "script": "Invoke-ExternalCommand -FilePath \"$dir\\npcap.exe\" -ArgumentList \"/S /D=$dir\\npcap\" -RunAs | Out-Null"
```
Ash258: Unescaped dir after `/D`

```suggestion
        "script": "Invoke-ExternalCommand -FilePath \"$dir\\npcap.exe\" -ArgumentList \"/S /D=\"\"$dir\\npcap\"\"\" -RunAs | Out-Null"
```

Try installation with single quotes

bucket/nmap.json
```diff
@@ -4,19 +4,19 @@
...
     },
     "uninstaller": {
-        "file": "Uninstall.exe",
-        "args": "/S"
+        "script": "Invoke-ExternalCommand -FilePath \"$dir\\npcap\\Uninstall.exe\" -ArgumentList \"/S\" -RunAs | Out-Null"
```
Ash258:
```suggestion
        "script": "Invoke-ExternalCommand -FilePath \"$dir\\npcap\\Uninstall.exe\" -ArgumentList '/S' -RunAs | Out-Null"
```

bucket/nmap.json
linsui: It doesn't work with quotes.

rasa: nsis requires `/D$dir`. It breaks with quotes. See https://github.com/lukesampson/scoop/pull/2003

linsui:
@Ash258 RunAs shows a UAC pop. 
Npcap need privelege to install so if maintainers want to keep main bucket non-privilege we can move it to extras or non-portable. It seems that the only criteria which is executed strictly now is non-GUI...

PR #270, opened 2019-07-27
nim: Persist config directory

YenForYang:


Ash258:
I cannot find any mention about config folder inside application directory.

https://github.com/nim-lang/nimble#configuration

There is only `C:\Users\<YourUser>\AppData\Roaming\nimble\nimble.ini` inside documentation.

@YenForYang Provide url to documentation where this folder is mentioned

YenForYang:
Yikes forgot I made this issue, sorry the delayed response.

From [Nim Compiler User Guide - Configuration files](https://nim-lang.org/docs/nimc.html#compiler-usage-configuration-files), has the following line

>The nim executable processes configuration files in the following directories (in this order; later files overwrite previous settings):
> - `$nim/config/nim.cfg`, `/etc/nim/nim.cfg` (UNIX) or `<Nim's installation director>\config\nim.cfg`(Windows). This file can be skipped with the `--skipCfg` command line option.
> ...

YenForYang:
Ah I just thought it would be reasonable to assume that the `config` folder has other `config` related files (probably details are buried in the documentation somewhere).

YenForYang:
There are two other files at the moment in the folder, `nimdoc.cfg` and `nimdoc.tex.cfg`. I guess you could separately include those files with `nim.cfg` but I think it would be less future-proof.

PR #272, opened 2019-07-27
 Fix ~156 license tags

linsui:


bucket/7zip.json
```diff
@@ -2,7 +2,7 @@
     "homepage": "https://www.7-zip.org/",
     "description": "A multi-format file archiver with high compression ratios",
     "license": {
-        "identifier": "LGPL-2.0-only,BSD-3-Clause",
+        "identifier": "LGPL-2.0-only|BSD-3-Clause",
```
Ash258: `,` is valid.

https://github.com/lukesampson/scoop/blob/b9b48182b4d921aa92e9b3e53ab2d2c376d467d7/libexec/scoop-info.ps1#L72

linsui: > If multiple licenses, seperate licenses with a comma (,). If dual licensed, seperate licenses with a pipe symbol (|).

bucket/cloc.json
```diff
@@ -2,7 +2,7 @@
     "homepage": "https://github.com/AlDanial/cloc",
     "description": "Counts blank lines, comment lines, and physical lines of source code in many programming languages",
     "version": "1.82",
-    "license": "GPL-2.0",
+    "license": "GPL-2.0-only|Artistic-2.0",
```
Ash258: Decide which format you want to prefer and then use it everywhere.

You are mixing `,` and `|` without any pattern or anything.

linsui: > If multiple licenses, seperate licenses with a comma (,). If dual licensed, seperate licenses with a pipe symbol (|).

I just follow the doc.

bucket/7zip.json
rasa: [Dual license](https://en.m.wikipedia.org/wiki/Multi-licensing) means the end-user can choose which license to use. Simultaneous licenses are when different files are licensed under different licenses.

linsui: Then maybe we should use `,` in most manifest? And I have another question, do we need to list all license? For some app it's really diffcult. I prefer to list one or two license and add a url. If we have a clear and doable rule in doc it will be much easier. I'll check again later.

rasa: In my opinion, if there are more than 2 or 3 licenses, we just add `,...` at the end. With one caveat: if they are not all open-source licenses (e.g. cockroachdb) we list all non-open source licenses first.

linsui: Thanks for your suggestion. I'll apply it later. Could you add it into doc?

rasa: I added it to the docs at https://github.com/lukesampson/scoop/wiki/App-Manifests#license :

If it is difficult to determine all of the different licenses, it is acceptable to add `,...` at the end of the SPDX list. If there are both open-source and not open-source licenses, please list all non-open source licenses first.

linsui: Thanks! It's much more clear. I just wonder do we need `Non-licensed` ? Generally if an app has no license it means all right reserved and I prefer to just  tag it as `Freeware` .

rasa: > I just wonder do we need `Non-licensed` ?

We certainly don’t need it. Perhaps we just use `Unknown` for when software doesn’t appear to have a license. In practice, this is seldom, if ever.

> how to deal with a free license not in SPDX?

We could provide our own identifiers, such as `Unlicense`, defining them in a `licenses.txt` file, such as:
```
Identifer\tDescription\tURL
Unlicense\tThe Unlicense\thttps://unlicense.org/
...

Comment (linsui) [reply bucket/7zip.json]: I'm affraid I didn't understand what you mean. There are lots of freeware without a license or eula, should I tag them as Non-license or Unknown?

File: bucket/radare2.json [hunk 3]
```diff
@@ -1,7 +1,7 @@
 {
     "homepage": "https://www.radare.org/r/",
     "version": "3.6.0",
-    "license": "GPL-2.0",
+    "license": " LGPL-3.0-only",
```
r15ch13: Random whitespace

bucket/sliksvn.json
```diff
@@ -1,7 +1,7 @@
 {
     "homepage": "https://sliksvn.com/",
     "version": "1.12.0",
```
rasa: Apache-2.0 per https://sliksvn.com/support/license-applies-sliksvn-client/

PR #273, opened 2019-07-27
Process Explorer manifest

alkuzad:
Choco has one: https://chocolatey.org/packages/procexp

We cannot be worse :)


linsui:
It's part of extras/sysinternals

h404bi:
or: https://github.com/h404bi/dorado/blob/master/bucket/process-explorer.json

alkuzad:
@linsui ok so I will use dorado bucket then, do not want to clumsy my env with everything, This can be closed then.

PR #289, opened 2019-08-02
Update PHP versions

phh:
Update PHP from 7.3.7 to 7.3.8.

Also, seems like [the current version](https://windows.php.net/downloads/releases/php-7.3.7-Win32-VC15-x64.zip) is not found.


r15ch13:
Thanks 👍 
It seems our update service got stuck. :grin:

PR #295, opened 2019-08-04
Puppet: fix checkver and autoupdate

linsui:


bucket/puppet.json
```diff
@@ -1,14 +1,15 @@
 {
     "homepage": "https://puppetlabs.com",
-    "version": "5.5.8",
+    "version": "6.7.2",
+    "license": "Apache-2.0",
```
Ash258: No description

PR #299, opened 2019-08-04
pngquant: Fix checkver

rasa:
See #150

### DIFF
--- a/bucket/pngquant.json
+++ b/bucket/pngquant.json
@@ -7,7 +7,7 @@
     "hash": "7ee4f98c7c0dfd23807b624a90a2aa3108ed46bc01f1bbd322bb9e3f3f54e6a6",
     "extract_dir": "pngquant",
     "bin": "pngquant.exe",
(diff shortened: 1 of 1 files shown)


niheaven:
Nope, that version is not Windows release version... Waiting for author's update should be better.

rasa:
https://pngquant.org/pngquant-windows.zip contains pngquant.exe which is version 2.12.0:
```
c:\>pngquant.exe -V
2.12.0 (January 2018)
```

niheaven:
Change its version to 2.12.0?

rasa:
> Change its version to 2.12.0?

I would say no, as there is no way to know what the .exe's version is, other than to download it, unzip it, and run it.

niheaven:
Then left it here, since the lastest version that we get from the website is 2.12.3...

r15ch13:
2.12.1 and newer seem to only contain changes to the build system https://github.com/kornelski/pngquant/issues/312

PR #311, opened 2019-08-09
bulk-rename-command: Add version 1.3.3.0

niheaven:


bucket/bulk-rename-command.json
```diff
@@ -0,0 +1,45 @@
+{
+    "homepage": "https://www.bulkrenameutility.co.uk",
+    "version": "1.3.3.0",
+    "license": {
+        "identifier": "Freeware",
```
rasa: It requires a commercial license at times, so let’s use Proprietary, right?

bucket/bulk-rename-command.json
```diff
@@ -0,0 +1,45 @@
+{
+    "homepage": "https://www.bulkrenameutility.co.uk",
+    "version": "1.3.3.0",
+    "license": {
+        "identifier": "Proprietary",
+        "url": "https://www.bulkrenameutility.co.uk/License.php"
+    },
+    "description": "File renaming software for Windows (Command Line Version)",
```
Ash258: I think `command line version` is redundant as it is in manifest name and also in Main bucket

PR #312, opened 2019-08-09
MEGAcmd: Add version 1.1.0

issaclin32:
close lukesampson/scoop-extras#2597

**MEGAcmd** ([homepage](https://mega.nz/cmd)) is a command line tool to access MEGA(mega.nz) maintained by mega.nz itself.

Notes:
* **installer/uninstaller script**: 
MEGAcmd contains about 50 batch files for commands like `mega-get`, `mega-ls`, ...
These commands can also be executed by `MEGAclient.exe get`, `MEGAclient.exe ls`, and so on.

~There are 3 possible ways (I can think of) to handle this:~
~1. Ignore these commands and add `MEGAclient.exe` only.~
~2. Map all the available commands into `shim` directory.~
~3. use `env_add_path` (adding possibility of conflict with other programs on DLL files)
I think `2.` is the best approach.~

(put every batch file in `bin` instead)


bucket/megacmd.json
```diff
@@ -0,0 +1,33 @@
+{
+    "homepage": "https://mega.nz/cmd",
+    "description": "Command line tool to access MEGA(mega.nz)",
```
Ash258:
```suggestion
    "description": "Manage MEGA.nz account and files from command line.",
```

bucket/megacmd.json
```diff
@@ -0,0 +1,33 @@
...
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://github.com/meganz/MEGAcmd/blob/master/build/installer/terms.txt"
+    },
+    "url": "https://mega.nz/MEGAcmdSetup.exe#/dl.7z",
+    "hash": "9a7ddc65cc6a792af2d5f54d69b2e08a5a48db050c8d3131011805b508a2f20d",
+    "installer": {
+        "script": [
+            "Remove-Item \"$dir\\`$PLUGINSDIR\", \"$dir\\uninst.exe\" -Recurse",
+            "Get-ChildItem \"$dir\" 'mega-*.bat' | ForEach-Object {",
```
Ash258: Just add them all as bin. What's problem?

Ash258: Doing this manually is nonsense and dangerous.

Ash258: `gci . 'mega-*.bat' | Select -expand Basename | % { "\"$_\"," } | clip` and paste it into manifest

issaclin32: There may be more batch file added in the future. 
If I simply put every thing in `bin`, new commands will be ignored. 
(Until some user found out and file an issue)

Ash258: And? That is casual.

issaclin32: You mean just add `gci...` to the install script? Or replacing the whole part after `Get-ChildItem`?

Ash258: No ...

1. Open manifest inside text editor
1. Create bin property as array.
1. Execute `gci . 'mega-*.bat' | Select -expand Basename | % { "\"$_\"," } | clip` inside terminal
1. Press CTRL+V in bin property inside manifest
1. Remove trailing comma
1. Press shortcut for formatting json file
1. Commit changes

issaclin32: Thanks for the instruction.
I found that batch files in `bin` will not work without adding `.bat` in the end.

bucket/megacmd.json
```diff
@@ -0,0 +1,79 @@
...
+    "homepage": "https://mega.nz/cmd",
+    "description": "Manage MEGA.nz account and files from command line.",
+    "version": "1.1.0",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://github.com/meganz/MEGAcmd/blob/master/build/installer/terms.txt"
+    },
+    "url": "https://mega.nz/MEGAcmdSetup.exe#/dl.7z",
+    "hash": "9a7ddc65cc6a792af2d5f54d69b2e08a5a48db050c8d3131011805b508a2f20d",
+    "bin": [
```
Ash258:
```suggestion
    "bin": [
        "MEGAclient.exe",
```

PR #317, opened 2019-08-11
y-cruncher: Add version 0.7.7.9501

issaclin32:
**y-cruncher** ([homepage](http://www.numberworld.org/y-cruncher/)) is a multi-threaded benchmark program that calculate Pi and other constants to trillions of digits.

Notes:
* The website (www.numberworld.org) does not support HTTPS.

### DIFF
--- a/bucket/y-cruncher.json
+++ b/bucket/y-cruncher.json
@@ -1,6 +1,6 @@
 {
     "homepage": "http://www.numberworld.org/y-cruncher/",
-    "description": "Scalable multi-threaded benchmark program that calculate Pi and other constants to trillions of digits",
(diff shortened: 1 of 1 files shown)


bucket/y-cruncher.json
```diff
@@ -0,0 +1,18 @@
+{
+    "homepage": "http://www.numberworld.org/y-cruncher/",
+    "description": "Scalable multi-threaded benchmark program that calculate Pi and other constants to trillions of digits",
```
Ash258:
```suggestion
    "description": "Scalable multi-threaded benchmark calculating multiple mathematical constants to trillions of digits.",
```

PR #322, opened 2019-08-12
Erlang: Extract first to avoid sudo

linsui:


bucket/erlang.json
```diff
@@ -1,17 +1,22 @@
 {
     "homepage": "https://www.erlang.org",
+    "description": "A programming language used to build massively scalable soft real-time systems with requirements on high availability.",
```
rasa: This will conflict with https://github.com/ScoopInstaller/Main/blob/7c1a4b5a681d16b950c4a0776c48e3a94d6f5bfc/bucket/erlang.json#L3

linsui: Merged...

PR #324, opened 2019-08-12
fciv: move from extras

linsui:
https://github.com/lukesampson/scoop-extras/pull/2623


Ash258:
@h404bi Please do not merge manifest moving PRs when you do not have push access to both repositories.

This will cause two same manifests in two repositories when `extras` take precedence as it is alphabetically earlier, which could confuse end users.

PR #329, opened 2019-08-14
R: add au.hash

linsui:


niheaven:
Please use `https://cloud.r-project.org` for better spead.

niheaven:
Use arch spec shortcut that target to `bin\x64\Rgui.exe` or `bin\i386\Rgui.exe`. `bin` should be also target to arch spec dir, I thought.

linsui:
I fix the shortcuts. No need to change bin as it is arch spec.

niheaven:
`bin/R.exe` is a wrapper of `bin/x64/R.exe` and you should add the following bins:

- `bin/*ARCH*/R.exe`
- `bin/*ARCH*/Rcmd.exe`
- `bin/*ARCH*/Rgui.exe`
- `bin/*ARCH*/Rscript.exe`
- `bin/*ARCH*/Rterm.exe`

linsui:
@niheaven Yeah, you are right! Thanks! I forgot to add Rcmd because the installer doesn't create a wrapper for it.

PR #337, opened 2019-08-15
rust, rust-msvc: Move apps out of the only dir

dodorz:


niheaven:
Does this work fine?

dodorz:
It works fine here.

PR #338, opened 2019-08-15
sqlite: add au.hash

linsui:


bucket/sqlite.json
```diff
@@ -12,10 +13,14 @@
     ],
     "checkver": {
         "url": "https://www.sqlite.org/download.html",
-        "re": "sqlite-tools-win32-x86-(?<number>[\\d]+).zip[\\s\\S]+for SQLite version (?<version>[\\d.]+)\\."
+        "regex": "sqlite-tools-win32-x86-(?<number>[\\d]+).zip[\\s\\S]+for SQLite version (?<version>[\\d.]+)\\."
```
Ash258: - `.*?` should work same as `[\\s\\S]+`
- named version group is not needed

```suggestion
        "regex": "sqlite-tools-win32-x86-(?<number>[\\d]+).zip.*?for SQLite version ([\\d.]+)\\."
```

PR #340, opened 2019-08-15
portainer: add au.hash

linsui:


bucket/portainer.json
```diff
@@ -1,5 +1,6 @@
 {
     "homepage": "https://portainer.io/",
+    "description": "Making Docker management easy.",
```
Ash258: Take description from #33

PR #341, opened 2019-08-15
opus tools: add 64bit

linsui:


Ash258:
Why did you take `0.2-opus-1.3`??

On downloads page there is `opus-tools-0.2-win32.zip` without opus number, on ftp there is also `opus-tools-0.2-opus-1.3.1.zip`

image

linsui:
They are different files, and this is the link on the download page, don't know why. There is no 64bit binary for 1.3.1.

PR #354, opened 2019-08-15
allure: add au.hash

linsui:


baev:
@linsui why did you change distribution from Maven Central back to Bintray? 

https://github.com/lukesampson/scoop/pull/2979

Bintray reached EOF by now, and I thought we was all set and well prepared but just found that Bintray url is specified for all the last releases ...

linsui:
I don't know... I can't remember the reason 2 years ago...

PR #366, opened 2019-08-17
flutter: Move from extras

linsui:


bucket/flutter.json
```diff
@@ -0,0 +1,46 @@
...
+        "https://storage.googleapis.com/flutter_infra/releases/stable/windows/flutter_windows_v1.5.4-hotfix.2-stable.zip",
+        "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/flutter-dev-setup.ps1"
+    ],
+    "hash": [
+        "1d5e1779430be6bbb4fa49f4bccc7c5d24a76f3d2a100129e09b4b8eef33b1b4",
+        "bbd8dd269dd70d97e0224025281e55b7e2e32364d5c47e082ca7f45e33d1a613"
+    ],
+    "depends": [
+        "android-sdk",
+        "java/adopt8-hotspot"
```
Ash258: This is special requirement to have this version?

bucket/flutter.json
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "extract_dir": "flutter",
+    "bin": [
+        "bin\\flutter.bat",
+        "flutter-dev-setup.ps1"
+    ],
+    "env_add_path": "bin\\cache\\dart-sdk",
+    "post_install": [
+        "flutter-dev-setup.ps1",
```
Ash258: If you want to execute this I doubt it will work as context of execution will be set to Execution path.

```suggestion
        "& \"$dir\\flutter-dev-setup.ps1\"",
```

bucket/flutter.json
```diff
@@ -0,0 +1,46 @@
...
+        "Write-Host Some licenses need to be accepted before developing. We recommend you do so by running 'flutter doctor --android-licenses'. -ForegroundColor Yellow",
```
Ash258:
```suggestion
        "Write-Host 'Some licenses need to be accepted before developing. We recommend you do so by running ''flutter doctor --android-licenses''.' -ForegroundColor Yellow",
```

scripts/flutter-dev-setup.ps1
```diff
@@ -0,0 +1,169 @@
...
+.NOTES
+  Version:        1.0
+  Author:         Ardelean Călin
+  Creation Date:  14 July 2018
+  Purpose/Change: Initial script development
+#>
+
+
+# Gets the installed packages list from sdkmanager
+function get-installed {
```
Ash258: This script could be rewritten to follow core format. But as it is work it is OK for moving.

bucket/flutter.json
linsui: flutter does not depend on java, but android sdk depends on java.

Ash258: Do not manually mark conversation as resolved when all users who was part of conversation did not read it. Because when you comment on discussion other user will get notification for it. And when you push to repository notification will be replaced with this push.

Which make other user to go through all collapsed conversations and uncollapsed them and see them one by one if there is new message. Instead of seeing it on PR page directly.

Findulas:
Flutter is not upgrade to new versions right now. It seems to be blocked by this move. Ash258 you wrote that it is OK for moving, but this PR is blocked. What needs to be done to get this PR and the one in extras merged?

linsui:
https://github.com/lukesampson/scoop/pull/3613 and https://github.com/lukesampson/scoop/pull/3721 block fultter get update.

PR #375, opened 2019-08-22
CUDA: fix autoupdate

linsui:

Wondering why hash extract not work without au.hash.regex.

### DIFF
--- a/bucket/cuda.json
+++ b/bucket/cuda.json
@@ -1,5 +1,5 @@
 {
-    "version": "10.1.168_425.25",
+    "version": "10.1.243_426.00",
     "description": "A parallel computing platform and programming model invented by NVIDIA",
     "homepage": "https://developer.nvidia.com/cuda-toolkit",
     "license": {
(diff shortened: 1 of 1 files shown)


bucket/cuda.json
```diff
@@ -35,11 +35,12 @@
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://developer.nvidia.com/compute/cuda/$majorVersion.$minorVersion/Prod/local_installers/cuda_$version_win10.exe#/dl.7z"
+                "url": "https://developer.download.nvidia.com/compute/cuda/$majorVersion.$minorVersion/Prod/local_installers/cuda_$version_win10.exe#/dl.7z"
             }
         },
         "hash": {
-            "url": "https://developer.download.nvidia.com/compute/cuda/$majorVersion.$minorVersion/Prod/docs2/sidebar/md5sum.txt"
+            "url": "https://developer.nvidia.com/cuda-downloads?target_os=Windows&target_arch=x86_64&target_version=10&target_type=exelocal",
```
Ash258: What about this?

http://developer.download.nvidia.com/compute/cuda/10.1/Prod/docs/sidebar/md5sum.txt

A

linsui: There is no hash for this version. And the md5 file url is changing...(docs2 ->docs)

Ash258:
Also sidenote. What is difference between `win10` and `windows` installers?

linsui:
`win10` for win10 and windows server 2016/2019, `windows` for older windows version.

Ash258:
Obviously, but what is the actual reason in that binaries?

linsui:
Do not know... I guess it's related to drivers.

PR #378, opened 2019-08-25
Move openvpn from extras and add tap

linsui:
https://github.com/lukesampson/scoop-extras/pull/2734

### DIFF
--- a/bucket/openvpn.json
+++ b/bucket/openvpn.json
@@ -1,0 +1,28 @@
+{
+    "version": "2.4.7-I603",
+    "description": "A flexible virtual private network (VPN) solution to secure your data communications.",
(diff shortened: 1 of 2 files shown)


Ash258:
Why you are moving manifests between buckets when there is no support for it? It is pain for users to find out and fix it, no automatic resolvoing, no command added for this and user do not know about it unless they read log really well.

linsui:
Yeah, maybe I should wait until scoop core support move manifest without bother users? Then I'll just change the manifest without move.

linsui:
Please review.

Ash258:
How `-RunAs` behave for end user? Does it spawn UAC pop, it will fail or ...?

linsui:
It shows a UAC pop.

PR #379, opened 2019-08-25
Latex: Fix env add path

linsui:
https://github.com/lukesampson/scoop/issues/3609
A temporary workaround before https://github.com/lukesampson/scoop/pull/3503 is merged...


bucket/latex.json
```diff
@@ -2,11 +2,11 @@
     "##": "Fails to validate schema, as env_add_path is not allowed in architecture/XXbit",
```
rasa:
```suggestion
```
No longer true

PR #380, opened 2019-08-26
Remove the parameter "site" of "Cygwin Setup" shortcut

KanDisheng:
"mirrors.kernel.org" sometimes is not the fastest site. Users should make their own choices.


KanDisheng:
The installation process should use the default site to prevent blocking it.
Shortcut of "Cygwin Setup" does not recommend using default site. The site that chose or added last time will be the default.

PR #381, opened 2019-08-26
scrcpy: Add shortcut for scrcpy-noconsole.exe

wenmin92:
Add a shortcut may be handy.


wenmin92:
In version 1.10, there's a problem: [scrcpy-noconsole.exe shows a console · Issue #691](https://github.com/Genymobile/scrcpy/issues/691), but does not affect.

PR #392, opened 2019-09-03
Removed godoc binary

shilangyu:
go 1.13 [no longer provides](https://golang.org/doc/go1.13#godoc) one out of the box


kroppt:
For reference: https://golang.org/doc/go1.12
"Go 1.12 is the last release that will include the `godoc` webserver; in Go 1.13 it will be available via go get."

PR #405, opened 2019-09-11
add aliyun-cli

patrick330602:
include Alibaba Cloud commandline interface to the main bucket


bucket/aliyun.json
```diff
@@ -0,0 +1,21 @@
+{
+    "description": "Alibaba Cloud CLI",
```
Ash258:
```suggestion
    "description": "Manage and use Alibaba Cloud resources.",
```

patrick330602: Noted and changed.

PR #425, opened 2019-09-17
traefik: Fix au.url

rasa:
See #150.

### DIFF
--- a/bucket/traefik.json
+++ b/bucket/traefik.json
@@ -26,6 +26,9 @@
             "32bit": {
                 "url": "https://github.com/containous/traefik/releases/download/v$version/traefik_v$version_windows_386.zip"
             }
+        },
+        "hash": {
+            "url": "$baseurl/traefik_v$version_checksums.txt"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


bucket/traefik.json
```diff
@@ -21,10 +21,10 @@
...
         "architecture": {
             "64bit": {
-                "url": "https://github.com/containous/traefik/releases/download/v$version/traefik_windows-amd64.exe#/traefik.exe"
+                "url": "https://github.com/containous/traefik/releases/download/v$version/traefik_v$version_windows_amd64.zip"
             },
             "32bit": {
-                "url": "https://github.com/containous/traefik/releases/download/v$version/traefik_windows-386.exe#/traefik.exe"
+                "url": "https://github.com/containous/traefik/releases/download/v$version/traefik_v$version_windows_386.zip"
             }
         }
```
Ash258:
```suggestion
        },
        "hash": {
            "url": "$baseurl/traefik_v$version_checksums.txt"
        }
```

rasa: But of course!

linsui:
> Please be aware that the old configurations for Traefik v1.X are NOT compatible with the v2.X config as of now.

It seems that they will maintain two versions for some time.

PR #428, opened 2019-09-18
Concurse-fly: Add version 5.5.1

javakhishvili:
Concourse is primarily driven from the command-line; there is no GUI config wizard. So, the first step to getting started with Concourse is to install the fly CLI tool.


bucket/fly.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "5.5.1",
+    "description": "Command-line tool used to control Concourse CI/CD",
+    "license": "Apache-2.0",
+    "homepage": "https://concourse-ci.org/",
```
Ash258:
```suggestion
    "homepage": "https://concourse-ci.org/fly.html",
```

bucket/fly.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "5.5.1",
+    "description": "Command-line tool used to control Concourse CI/CD",
```
Ash258:
```suggestion
    "description": "Concourse CI/CD manager and controller.",
```

linsui:
Should we rename it as concurse-fly? Fly is a common word.

javakhishvili:
> Should we rename it as concurse-fly? Fly is a common word.

Yes, that makes perfect sense. Agreed.

PR #430, opened 2019-09-19
radare2: Fix 'extract_dir'

JFRN:
Extract dir changed and should not contain the _dyn suffix


linsui:
Please also fix extract_dir in autoupdate.

JFRN:
> 
> 
> Please also fix extract_dir in autoupdate.

Fixed.

PR #448, opened 2019-09-23
LLVM: Add remaining binaries as shims

DaanDeMeyer:
This PR adds the remaining binaries in LLVM's bin directory as shims.


Boscop:
@DaanDeMeyer Any idea why lld only shows ELF as supported, not PE?
> C:\Users\me\scoop\apps\llvm\current\bin\ld.lld.exe: supported targets: elf

It means it's not working for me:
https://users.rust-lang.org/t/using-llvm-lld-linker-on-windows-for-faster-linking/37938

Boscop:
Ah, nvm, I had to write `linker = "lld-link.exe"` in my config, now it works..

PR #451, opened 2019-09-26
dos2unix: add autoupdate section

rasa:


bucket/dos2unix.json
```diff
@@ -21,6 +21,16 @@
     ],
     "checkver": {
         "url": "https://waterlan.home.xs4all.nl/dos2unix.html",
-        "re": "Latest version: (?<version>[\\d.]+) \\([\\d\\-]+\\)"
+        "re": "Latest version: ([\\d.]+) \\([\\d\\-]+\\)"
```
Ash258: Do you need to be so strict?

```suggestion
        "regex": "Latest version: ([\\d.]+)\\s+\\(\\d"
```
Should be enough. Or?

bucket/dos2unix.json
```diff
@@ -21,6 +21,16 @@
...
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://waterlan.home.xs4all.nl/dos2unix/dos2unix-$version-win64.zip"
```
Ash258: Why not sourceforge? Current URLs are from sourceforge

rasa: Understood, but the URLs are on https://waterlan.home.xs4all.nl/dos2unix.html, so let's consider these URLs are "authoritative". Also, Sourceforge returns 429 errors during many excavator runs, so minimizing hitting sourceforge will decease these 429s.

Ash258: Then run checkver to change current URLs

PR #454, opened 2019-09-30
halo: Add version 1.1.1

niheaven:


linsui:
Should you suggest java? It seems that in jar file manifest sometimes jave is suggested but sometimes not.

h404bi:
> I am not fan of these non english applications inside official buckets, but manifest is OK

@niheaven You can transfer it to my [bucket](https://github.com/h404bi/dorado) if you're willing to. haha

PR #455, opened 2019-09-30
nu: Add version 0.3.0

niheaven:
> Nu will look for the plugins in your PATH on startup. While nu will have some functionality without them, for full functionality you'll need to copy them into your path so they can be loaded.

### DIFF
--- a/bucket/nu.json
+++ b/bucket/nu.json
@@ -6,7 +6,14 @@
     "url": "https://github.com/nushell/nushell/releases/download/0.3.0/nu_0_3_0_windows.zip",
     "hash": "73f5ac08c3c3fe1f81ec45bcae95985068f859d3aceb1267e88b924317e227f7",
     "extract_dir": "nushell_0_3_0",
-    "env_add_path": ".",
+    "installer": {
+        "script": [
+            "New-Item \"$dir\\Plugins\" -ItemType Directory | Out-Null",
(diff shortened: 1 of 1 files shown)


bucket/nu.json
```diff
@@ -0,0 +1,18 @@
...
+    "version": "0.3.0",
+    "license": "MIT",
+    "description": "A modern shell written in Rust.",
+    "url": "https://github.com/nushell/nushell/releases/download/0.3.0/nu_0_3_0_windows.zip",
+    "hash": "73f5ac08c3c3fe1f81ec45bcae95985068f859d3aceb1267e88b924317e227f7",
+    "extract_dir": "nushell_0_3_0",
+    "env_add_path": ".",
+    "checkver": {
+        "github": "https://github.com/nushell/nushell",
+        "regex": "/tag/([d.]+)"
```
Ash258: ??
Why you need special regex for this?

niheaven: Wo, yes, normal regex works.

Ash258:
> you'll need to copy them into your path so they can be loaded.

Would not make sense to crate some folder **Plugins** folder inside application directory and then add normal bins and plugins folder to path.

niheaven:
That sounds well, but we need to check if there're new ones in new release. I'll change it and check periodically.

PR #466, opened 2019-10-04
nmap-portable: Add version 7.80

r15ch13:
This provides a nmap installation without priviliges.
- Some small fixes for the nmap manifest
- Closes https://github.com/lukesampson/scoop/issues/1799

### DIFF
--- a/bucket/nmap-portable.json
+++ b/bucket/nmap-portable.json
@@ -1,0 +1,44 @@
+{
+    "homepage": "https://nmap.org",
+    "description": "Network exploration and security auditing utility.",
+    "version": "7.80",
+    "license": {
+        "identifier": "GPL-2.0-only",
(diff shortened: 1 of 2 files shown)


r15ch13:
It would be possible to only extract the `npcap-*-oem.exe` and create a separate manifest for it Npcap this way. But I don't think the license would allow this.

https://nmap.org/npcap/#License
https://github.com/nmap/npcap/blob/master/LICENSE

linsui:
If npcap need to be installed manully, why not use the zip package without npcap?
I wonder if it is allowed to install npcap by extracting and installing drivers with pnputil?

r15ch13:
npcap will be installed to `C:\Program Files` and is available under `Add or remove programs`

Ash258:
For me this could go in. Manifests looks good and for behavior I cannot really tell, as I am not consumer.

PR #471, opened 2019-10-07
kustomize: update to latest version

kungfutse:


Ash258:
I do not know if there will be only kustomize releases tagged as latest, but with this weird release process i think latest could be flanky and cause Excavator trying to match non existent releases

Right now there are two newer releases (pluginator, go api), which are not tagged as latest. It could be kept this way or using this checkver to really match only kustomize releases:

```json
    "checkver": {
          "url": "...kustomize/releases",
          "regex": "tree/kustomize/v([\\d.]+)"
   },
```

rasa:
See #150

PR #479, opened 2019-10-12
dotnet-sdk: Set MSBuildSDKsPath environment

johannesegger:
OmniSharp isn't able to find SDKs otherwise (see https://github.com/OmniSharp/omnisharp-vscode/issues/2970#issuecomment-524292461)

### DIFF
--- a/bucket/dotnet-sdk.json
+++ b/bucket/dotnet-sdk.json
@@ -15,7 +15,8 @@
     },
     "bin": "dotnet.exe",
     "env_set": {
-        "DOTNET_ROOT": "$dir"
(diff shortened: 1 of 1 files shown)


wenqiangxie:
It seems that fix not work. I install dotnet-sdk with scoop, but VSCode OmniSharp can't find any sdks. In system environment I can find MSBuildSDKSPath. Then I install with exe file form official site, everything works normal. I think there are others setting for this.

johannesegger:
That's interesting, because I updated dotnet-sdk after the fix and it started to work immediately.

PR #481, opened 2019-10-13
gifski: Fix checkver

h404bi:
From: https://github.com/ScoopInstaller/Main/issues/150

### DIFF
--- a/bucket/gifski.json
+++ b/bucket/gifski.json
@@ -8,7 +8,7 @@
     "extract_dir": "win",
     "bin": "gifski.exe",
     "checkver": {
-        "github": "https://github.com/ImageOptim/gifski/"
+        "regex": "For Windows.*?gifski-([\\d.]+).zip"
     },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/gifski.json
```diff
@@ -8,7 +8,7 @@
     "extract_dir": "win",
     "bin": "gifski.exe",
     "checkver": {
```
Ash258:
```suggestion
    "checkver": "For Windows.*?gifski-([\\d.]+)\\.zip",
```

h404bi: Are you sure? I can't see any tweaks after running `bin\formatjson.ps1`.

Ash258: https://github.com/lukesampson/scoop/blob/3e55a70971c5ff0d035daa54ca5dfab95dfaaa1d/bin/checkver.ps1#L109-L112

rasa:
I ran `checkver -f` on this locally, and it updated fine.

PR #486, opened 2019-10-17
jcli: Add a new cli

LinuxSuRen:
Official Repo is https://github.com/jenkins-zh/jenkins-cli/

### DIFF
--- a/bucket/jcli.json
+++ b/bucket/jcli.json
@@ -1,33 +1,18 @@
 {
     "version": "0.0.21",
+    "description": "Multiple Jenkins servers manager.",
     "homepage": "https://jenkins-zh.github.io/jcli-docs/",
-    "description": "Jenkins CLI allows you manage your Jenkins as an easy way",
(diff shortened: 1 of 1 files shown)


LinuxSuRen:
@Ash258 Thanks for the suggestion. Can I keep the description as the original one? The brew and [gofish](https://github.com/fishworks/fish-food/blob/master/Food/jcli.lua) used the same one.

LinuxSuRen:
@Ash258 Thanks for the suggestion. Can I keep the description as the original one? The brew and [gofish](https://github.com/fishworks/fish-food/blob/master/Food/jcli.lua) used the same one.

Ash258:
Why?

- There is no reason to mention CLI as it main criteria for main bucket.
- Aiming at user directly using `you`, `your` should not be inside generic application descriptions as it has to information value. This is use case for some marketing / PR.
- Your description do not contain main advantage of this utility = Multiple server management

PR #491, opened 2019-10-18
Add butterflow

TheRandomLabs:
I had to use a slightly convoluted checkver because the release tags aren't equal to the version strings used in the file names.

**Copied from README:**

BF is a video manipulation tool that can:

1. Make **motion interpolated videos** (increase a video's frame rate by rendering intermediate frames based on motion using a combination of pixel-warping and blending).
2. Make **smooth motion videos** (do simple blending between frames).
3. Leverage interpolated frames to make **fluid slow motion videos**.

### DIFF
--- a/bucket/butterflow.json
+++ b/bucket/butterflow.json
@@ -6,7 +6,7 @@
     "architecture": {
         "64bit": {
             "url": "https://github.com/dthpham/butterflow/releases/download/0.2.4.alpha2/butterflow-0.2.4a4-win64.zip",
-            "hash": "a77fbbdbdd0d85bb31feac30e53378a36ff4fb2a8a98ba15a6fa06def4b36ad1",
+            "hash": "856ad05bafaeb9d084d6b1cbf0bdec17defeef438e610f6a3a772087acea301b",
             "extract_dir": "butterflow-0.2.4a4-win64"
         }
     },
@@ -20,12 +20,12 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/dthpham/butterflow/releases/download/$matchRel/butterflow-$version-win64.zip",
-                "hash": {
-                    "url": "https://raw.githubusercontent.com/dthpham/butterflow/master/README.md",
-                    "find": "$basename[\\S\\s]+\\* Sha256: $sha256"
-                },
                 "extract_dir": "butterflow-$version-win64"
             }
+        },
+        "hash": {
+            "url": "https://raw.githubusercontent.com/dthpham/butterflow/master/README.md",
+            "regex": "(?sm)$basename.*?\\*\\s+Sha256:\\s+$sha256"
         }
     }
 }
--- a/bucket/butterflow.json
+++ b/bucket/butterflow.json
@@ -25,7 +25,7 @@
         },
         "hash": {
             "url": "https://raw.githubusercontent.com/dthpham/butterflow/master/README.md",
-            "regex": "(?sm)$basename.*?\\*\\s+Sha256:\\s+$sha256"
+            "regex": "$basename\\s+\\* Sha256: $sha256"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


linsui:
```
Windows 10 (Portable): Download the latest releases.

    Preview: butterflow-0.2.4a4-win64.zip
        Sha256: 856ad05bafaeb9d084d6b1cbf0bdec17defeef438e610f6a3a772087acea301b
    Stable: butterflow-0.2.3-win64.zip
        Sha256: a77fbbdbdd0d85bb31feac30e53378a36ff4fb2a8a98ba15a6fa06def4b36ad1
```

Please extract hash from readme. Why not use stable version?

TheRandomLabs:
Would it be worth having a separate manifest for the alpha version?

bucket/butterflow.json
```diff
@@ -20,6 +20,10 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/dthpham/butterflow/releases/download/$matchRel/butterflow-$version-win64.zip",
+                "hash": {
+                    "url": "https://raw.githubusercontent.com/dthpham/butterflow/master/README.md",
+                    "find": "$basename[\\S\\s]+\\* Sha256: $sha256"
```
Ash258: Multilne regex is needed
```suggestion
                    "regex": "(?sm)$basename.*?\\*\\s+Sha256:\\s+$sha256"
```

bucket/butterflow.json
```diff
@@ -6,7 +6,7 @@
     "architecture": {
         "64bit": {
             "url": "https://github.com/dthpham/butterflow/releases/download/0.2.4.alpha2/butterflow-0.2.4a4-win64.zip",
-            "hash": "856ad05bafaeb9d084d6b1cbf0bdec17defeef438e610f6a3a772087acea301b",
+            "hash": "a77fbbdbdd0d85bb31feac30e53378a36ff4fb2a8a98ba15a6fa06def4b36ad1",
```
Ash258: Wrong hash

bucket/butterflow.json
```diff
@@ -20,6 +20,10 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/dthpham/butterflow/releases/download/$matchRel/butterflow-$version-win64.zip",
+                "hash": {
```
Ash258: Move hash out of architecture property.

TheRandomLabs:
@linsui Thanks for the suggestion. I hope I've done it correctly. :) Also, the stable version is over a year older than the preview version, and [butterflow-ui](https://github.com/wagesj45/butterflow-ui) seems to deem it stable enough to use it in their release build. However, I can change it to the stable version if you would like.

bucket/butterflow.json
linsui: `$basename\\s+\\* Sha256: $sha256` is enough.

bucket/butterflow.json
TheRandomLabs: I got the right hash for 0.2.4.alpha2. `a77fbbdbdd0d85bb31feac30e53378a36ff4fb2a8a98ba15a6fa06def4b36ad1` is for 0.2.3.

bucket/butterflow.json
TheRandomLabs: Thanks, I don't have much experience with regular expressions.

linsui:
@TheRandomLabs If it's stable enough, it's Ok 😄

TheRandomLabs:
@Ash258 I've applied your suggestions. :)

TheRandomLabs:
Regarding the description, what was wrong with `a command-line tool for performing motion interpolation on videos`?

Ash258:
- `Command line tool` is obsolete for Main bucket
- Motion interpolation is not the only one feature this application can do

TheRandomLabs:
How about `a tool for making motion interpolated and smooth motion videos`?

TheRandomLabs:
Fair enough.

PR #496, opened 2019-10-21
kubectxwin: Add version 0.1.1

gracefullight:
https://github.com/lukesampson/scoop/issues/3510

- [kubectx](https://github.com/ahmetb/kubectx)
- [kubectxwin](https://github.com/thomasliddledba/kubectxwin)

### DIFF
--- a/bucket/kubectxwin.json
+++ b/bucket/kubectxwin.json
@@ -1,6 +1,6 @@
 {
     "homepage": "https://github.com/thomasliddledba/kubectxwin",
-    "description": "Windows Version of the powerful tool kubectx.",
+    "description": "Kubernetes cluster switcher.",
     "license": "MIT",
     "version": "0.1.1",
     "architecture": {
--- a/bucket/kubectxwin.json
+++ b/bucket/kubectxwin.json
@@ -1,0 +1,27 @@
+{
+    "homepage": "https://github.com/thomasliddledba/kubectxwin",
+    "description": "Windows Version of the powerful tool kubectx.",
+    "license": "MIT",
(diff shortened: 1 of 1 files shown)


bucket/kubectxwin.json
```diff
@@ -0,0 +1,27 @@
+{
+    "homepage": "https://github.com/thomasliddledba/kubectxwin",
+    "description": "Windows Version of the powerful tool kubectx.",
```
Ash258:
```suggestion
    "description": "Kubectl(1) context manager and switcher.",
```

or

```suggestion
    "description": "Kubernetes cluster switcher.",
```

I f

bucket/kubectxwin.json
```diff
@@ -0,0 +1,27 @@
...
+    "license": "MIT",
+    "version": "0.1.1",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/thomasliddledba/kubectxwin/releases/download/0.1.1/kubectxwin.exe",
+            "hash": "sha1:34703BCB9B16865541E58524B596B1AEC0E4E3D0"
```
Ash258: Sha1 hashes are provided somewhere? Or you just generated it using sha1?

gracefullight: I just generated 😂

gracefullight:
> I think it could be named `kubectx`. But I am not kubectx user so I cannot properly tell if this repository is official or it needs to be distinguish.

The official Kubectx provides [only bash script version](https://github.com/ahmetb/kubectx/blob/master/kubectx), so I think both libraries are managed separately

PR #497, opened 2019-10-21
dependency-check: Add version 5.2.2

stevehipwell:
Add [dependency-check](https://www.owasp.org/index.php/OWASP_Dependency_Check) package.

### DIFF
--- a/bucket/dependency-check.json
+++ b/bucket/dependency-check.json
@@ -21,7 +21,6 @@
         "url": "https://dl.bintray.com/jeremy-long/owasp/dependency-check-$version-release.zip",
         "hash": {
             "url": "$url.sha256",
-            "regex": "$sha256"
         }
     }
 }
--- a/bucket/dependency-check.json
+++ b/bucket/dependency-check.json
@@ -1,0 +1,26 @@
+{
+    "homepage": "https://www.owasp.org/index.php/OWASP_Dependency_Check",
+    "description": "Software composition analysis utility that identifies project dependencies and checks if there are any known, publicly disclosed, vulnerabilities.",
+    "license": "Apache-2.0",
+    "version": "5.2.2",
+    "url": "https://dl.bintray.com/jeremy-long/owasp/dependency-check-5.2.2-release.zip",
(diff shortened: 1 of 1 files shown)


bucket/dependency-check.json
```diff
@@ -0,0 +1,26 @@
...
+        ]
+    },
+    "checkver": {
+        "url": "https://jeremylong.github.io/DependencyCheck/current.txt",
+        "re": "(.+)"
+    },
+    "autoupdate": {
+        "url": "https://dl.bintray.com/jeremy-long/owasp/dependency-check-$version-release.zip",
+        "hash": {
+            "url": "$url.asc"
```
Ash258:
```suggestion
            "url": "$url.sha256"
```
I do not see way how could PGP signature provide hash

image

Ash258: I do not know if default regex will cover this. If checkver fallback to downloading add:
```suggestion
            "url": "$url.sha256",
            "regex": "$sha256"
```

bucket/dependency-check.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "checkver": {
+        "url": "https://jeremylong.github.io/DependencyCheck/current.txt",
+        "re": "(.+)"
+    },
+    "autoupdate": {
+        "url": "https://dl.bintray.com/jeremy-long/owasp/dependency-check-$version-release.zip",
+        "hash": {
+            "url": "$url.sha256",
+            "regex": "$sha256"
```
Ash258: Tested and regex is not needed

```suggestion
```

A

bucket/dependency-check.json
```diff
@@ -0,0 +1,27 @@
...
+        ]
+    },
+    "checkver": {
+        "url": "https://jeremylong.github.io/DependencyCheck/current.txt",
+        "re": "(.+)"
+    },
+    "autoupdate": {
+        "url": "https://dl.bintray.com/jeremy-long/owasp/dependency-check-$version-release.zip",
+        "hash": {
+            "url": "$url.sha256",
```
Ash258:
```suggestion
            "url": "$url.sha256"
```

PR #513, opened 2019-10-25
openssl-light: Add version 1.1.1d

linsui:


niheaven:
See https://github.com/ScoopInstaller/Versions/blob/master/bucket/openssl-light.json

linsui:
Thanks!

PR #515, opened 2019-10-28
ffsend: fix openssl dependency

linsui:


bucket/ffsend.json
```diff
@@ -1,15 +1,16 @@
 {
+    "version": "0.2.54",
     "homepage": "https://github.com/timvisee/ffsend",
-    "description": "Easily and securely share files from the command line. A fully featured Firefox Send client.",
+    "description": "Easily and securely share files with Firefox Send.",
     "license": "GPL-3.0-only",
-    "version": "0.2.54",
+    "depends": "version/openssl-light",
```
Ash258: https://github.com/ScoopInstaller/Versions
https://github.com/lukesampson/scoop/blob/master/buckets.json#L4

```suggestion
    "depends": "versions/openssl-light",
```

Ash258:
I cannot objectively review this. Openssl is completely out of my scope and I am avoiding it all the time.
Version bucket depends, and hard linking inside manifest is not something I would like to accept.

linsui:
What's your suggestion? We can
1. suggest openssl and openssl-light or move openssl-light to Main bucket
3. copy instead of make hardlink

rasa:
What about using
https://github.com/ScoopInstaller/Main/blob/master/bucket/openssl-mingw.json ?

Since it adds itself to the path, perhaps you don’t need to link or copy the dis?

linsui:
Good idea! It works well for me. Thanks!

ThatNerdyPikachu:
Can confirm that this fixes the issue, could this be merged?

timvisee:
Developer of `ffsend` here.

I just discovered (https://github.com/timvisee/ffsend/issues/106) a runtime error occurs using the current `ffsend` scoop package manifest because the OpenSSL library can't be found.

Is this just an attempt to fix that issue, or is there more to it?

I've submitted a PR to properly add the `openssl` libraries to `PATH` to fix exactly this, which might be a better solution than switching to `openssl-mingw`: https://github.com/ScoopInstaller/Main/pull/800

linsui:
This is just for that issue.

timvisee:
The `openssl` package has now been updated (#800), and might fix the issue this addresses. I don't have a Windows machine available right now to confirm this. Maybe someone (@linsui) can give it a try?

linsui:
I can't access my Windows box right now, either. But it should work. I'll close this. It's painful for me to download openssl which may takes about an hour so if you can deliver those dlls with ffsend it would be convient. Thanks for your work anyway :joy:

timvisee:
> ... so if you can deliver those dlls with ffsend it would be convient.

Good idea. I'm not sure how to do that though. Nor do I have the time right now to look into it. I believe I did take a look into static compilation for Windows with OpenSSL included, but I believe I didn't get it to work back then.

Even though I don't have plans to work on this, I've opened the following issue for tracking this request:
https://gitlab.com/timvisee/ffsend/issues/96

PR #524, opened 2019-10-31
ntop: add ntop

patrick330602:
add ntop, a htop-like system-monitor with Vi-emulation for Windows

### DIFF
--- a/bucket/ntop.json
+++ b/bucket/ntop.json
@@ -2,7 +2,7 @@
     "description": "htop-like system-monitor with Vi-emulation for Windows",
     "homepage": "https://github.com/Nuke928/NTop",
     "version": "0.3.3",
-    "license": "GPLv3",
+    "license": "GPL-3.0-only",
     "url": "https://github.com/Nuke928/NTop/releases/download/v0.3.3/ntop.exe",
     "hash": "b52babe71f1a654c34fac8f6ef946afa34574f23aeecbfd1ac6023b4e8ed1297",
     "bin": "ntop.exe",
--- a/bucket/ntop.json
+++ b/bucket/ntop.json
@@ -1,0 +1,13 @@
+{
+    "description": "htop-like system-monitor with Vi-emulation for Windows",
+    "homepage": "https://github.com/Nuke928/NTop",
+    "version": "0.3.3",
+    "license": "GPLv3",
+    "url": "https://github.com/Nuke928/NTop/releases/download/v0.3.3/ntop.exe",
+    "hash": "b52babe71f1a654c34fac8f6ef946afa34574f23aeecbfd1ac6023b4e8ed1297",
+    "bin": "ntop.exe",
+    "checkver": "github",
+    "autoupdate": {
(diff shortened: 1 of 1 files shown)


bucket/ntop.json
```diff
@@ -0,0 +1,13 @@
+{
+    "description": "htop-like system-monitor with Vi-emulation for Windows",
+    "homepage": "https://github.com/Nuke928/NTop",
+    "version": "0.3.3",
+    "license": "GPLv3",
```
niheaven:
```suggestion
    "license": "GPL-3.0-only",
```

bucket/ntop.json
```diff
@@ -0,0 +1,13 @@
+{
+    "description": "htop-like system-monitor with Vi-emulation for Windows",
+    "homepage": "https://github.com/Nuke928/NTop",
+    "version": "0.3.3",
+    "license": "GPL-3.0-only",
+    "url": "https://github.com/Nuke928/NTop/releases/download/v0.3.3/ntop.exe",
+    "hash": "b52babe71f1a654c34fac8f6ef946afa34574f23aeecbfd1ac6023b4e8ed1297",
+    "bin": "ntop.exe",
```
niheaven:
```suggestion
    "bin": "ntop.exe",
    "pre_install": [
        "if (!(Test-Path \"$persist_dir\\ntop.conf\")) {",
        "   try {",
        "       Invoke-WebRequest -Uri 'https://raw.githubusercontent.com/Nuke928/NTop/master/ntop.conf' -OutFile \"$dir\\ntop.conf\"",
        "   } catch {",
        "       @(",
        "           '# NTop default color scheme'",
        "           '# See https://msdn.microsoft.com/en-us/library/windows/desktop/ms682013(v=vs.85).aspx for color values (note that they need to be ORed by hand)'",
        "           'FGColor\t\t\t\t\t0x7'",
        "           'BGColor\t\t\t\t\t0x8'",
        "           'FGHighlightColor\t\t0xB'",
        "           'BGHighlightColor\t\t0xB0'",
        "           'MenuBarColor\t\t\t0x10'",
        "           'ProcessListHeaderColor\t0x20'",
        "           'ErrorColor\t\t\t\t0x47'",
        "           ''",
        "           '# Percentage bar colors'",
        "           'CPUBarColor\t\t\t0x4'",
        "           'MemoryBarColor\t\t0x2'",
        "           'PageMemoryBarColor\t0x2'",
        "       ) -join \"`r`n\" | Set-Content \"$dir\\ntop.conf\" -Encoding ASCII",
        "   }",
        "}"
    ],
    "persist": "ntop.conf",
```

niheaven: @patrick330602 You could add it in another PR.

Ash258: I would like to avoid downloading additional files in install blocks. This could be security issue when someone change this file.
Default configuration is not created after executing application?

niheaven:
Yes, the default config file is not created, and for default config, this file is not needed.

bucket/ntop.json
niheaven: Yes, the default config file is not created, and for default config, this file is not needed.

PR #525, opened 2019-10-31
memcached: Add version 1.5.19

### LINKED ISSUE #462: [Request] Add memcached
https://github.com/memcached/memcached

issaclin32:
close #462

**memcached** ([homepage](https://memcached.org/)) is a distributed memory object caching system.

Notes:
* [The official repo](https://github.com/memcached/memcached) only provides source code. This package downloads the pre-compiled binaries from https://github.com/nono303/memcached.

* ~This package has to be updated manually (for now). `Autoupdate` will be working after lukesampson/scoop#3518 get merged.~
(Download the ZIP file instead. Therefore `autoupdate` is not an issue.)

* `checkver` matches `memcache-` instead of `memcached-`. This is not a typo. (see: https://github.com/nono303/memcached/commits/master)

* This package does not require Cygwin to run, according to [nono303's comment on Apache forum](https://www.apachelounge.com/viewtopic.php?t=7919).


Ash258:
Do not use master. Use tagged branches to prevent hash mismatch

https://github.com/nono303/memcached/tree/memcache-1.5.19

image

issaclin32:
I found that I can just download the ZIP file (instead of multiple files).  Sorry I was blind.

r15ch13:
I would add the apache lounge forum post from nono303 as a comment to the manifest

https://www.apachelounge.com/viewtopic.php?t=7919

PR #544, opened 2019-11-06
texlive-x64: Add version 2019-11-03

li-ruijie:
The main texlive distro distributes only 32-bit binaries. As per the texlive documentation, 64-bit binaries are currently distributed and updated separately from the 32-bit binaries.


Ash258:
- Adding only 64bit package with this name is nonsense.
- You are not handlind insatllation, persists
- Since "release" is each 5 day (approx) you will spend about 2 to 4 hours of downloading and installaling packages from slow mirrors on this day.

See lukesampson/scoop-extras#2840

li-ruijie:
This is NOT the full texlive installation. Perhaps it is ill-named to give such an impression.
This only contains the x64 binaries that are not included with a standard texlive installation.

PR #570, opened 2019-11-14
miktex: exract instead of install

linsui:
- It works for me but need more test. I'm not a latex expert. 
- Do not know what are those files in the data directory and if it is needed to persist them.

More info: https://github.com/ScoopInstaller/Main/pull/123


niheaven:
It's a needless change... Installation by its original installer is totally portable, and the script is very simple. Not all install should be transfered to extraction.

linsui:
There is a popup window when running the installer. This is the main reason that I prefer extraction.

niheaven:
I don't like this one, see the diff result: (left is offical installation, right is extraction)

miktex-diff

There are 383 diff items between extraction and install, most of which are exes. Some of them are renamed from `miktex-*` in `install\miktex\bin`, but others aren't (e.g., a2ping.exe, a5toa4.exe, etc.).

linsui:
Thanks! I'll look into it. If I can fix this problem I'll reopen this pr.

PR #578, opened 2019-11-18
dnscrypt-proxy: fix hash

linsui:


linsui:
Fixed in https://github.com/ScoopInstaller/Main/commit/206cef4e1604e7adc84889ed7dd7bab07cc46012

PR #597, opened 2019-11-23
paket: Add version 5.236.6

kostrse:
Paket is a popular alternative package manager for .NET
https://fsprojects.github.io/Paket/

### DIFF
--- a/bucket/paket.json
+++ b/bucket/paket.json
@@ -1,0 +1,13 @@
+{
+    "version": "5.236.4",
+    "description": "Dependency manager for .NET with support for NuGet and Git repositories",
+    "homepage": "https://fsprojects.github.io/Paket/",
+    "license": "MIT",
+    "url": "https://github.com/fsprojects/Paket/releases/download/5.236.4/paket.exe",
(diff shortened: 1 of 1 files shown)


bucket/paket.json
```diff
@@ -0,0 +1,15 @@
...
+    "homepage": "https://fsprojects.github.io/Paket/",
+    "license": "MIT",
+    "url": "https://github.com/fsprojects/Paket/releases/download/5.236.6/paket.exe",
+    "hash": "c97bf9740162a833ccc20ade2bbe6cb79a2778d568663801bf74124b197050bd",
+    "bin": "paket.exe",
+    "checkver": {
+        "github": "https://github.com/fsprojects/Paket"
+    },
+    "autoupdate": {
+        "url": "https://github.com/fsprojects/Paket/releases/download/$version/paket.exe"
```
Ash258:
```suggestion
        "url": "https://github.com/fsprojects/Paket/releases/download/$version/paket.exe",
        "hash": "$baseurl/paket-sha256.txt"
```

kostrse: Updated

kostrse: Schema complains that hash should have an object with url, rather than a string directly.
I changed to object with url field.

PR #604, opened 2019-11-24
ie11webdriver: Update to version 3.150.1

linsui:
[Here](https://github.com/SeleniumHQ/selenium/tree/master/cpp/iedriverserver) is the source code.

### DIFF
--- a/bucket/ie11webdriver.json
+++ b/bucket/ie11webdriver.json
@@ -1,22 +1,22 @@
 {
-    "version": "3.14.0",
-    "homepage": "http://www.seleniumhq.org/",
-    "description": "Selenium WebDriver for Internet Explorer 11.",
-    "license": "Unknown",
-    "bin": "IEDriverServer.exe",
+    "version": "3.150.1",
+    "homepage": "https://selenium.dev",
+    "description": "Selenium WebDriver for Internet Explorer 11",
+    "license": "Apache-2.0",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/ie11webdriver.json
```diff
@@ -1,22 +1,22 @@
...
-            "hash": "29de31f5a3028224e5ae2df5757e112755ef8ccd0a06de8f333b129139647d70"
+            "url": "https://selenium-release.storage.googleapis.com/3.150/IEDriverServer_Win32_3.150.1.zip",
+            "hash": "09263561924a5a222b39bd0a80ef3778b8bf45457e1cf91c79b6a9f65d6230aa"
         }
     },
+    "bin": "IEDriverServer.exe",
     "checkver": {
-        "url": "http://www.seleniumhq.org/download/",
-        "re": "Download version ([\\d.]+).*[^<]+32 bit Windows IE"
+        "url": "https://selenium.dev/downloads/",
```
Ash258: Why you do not check it from raw https://github.com/SeleniumHQ/selenium/blob/master/cpp/iedriverserver/CHANGELOG ?

linsui: Good idea. 👍

linsui: Ping. When revision version is 0 we need to drop it. Also we don't know if we should drop the build version when it's 0. It's difficult to get the right version.

rasa: Is this concern resolved?

PR #606, opened 2019-11-24
axel: Update to version 2.16.1-1, use Unix Utils ported to Windows build

linsui:
There will not be an official build for Windows in a near future. See https://github.com/axel-download-accelerator/axel/issues/108 . 2.16.1-1 is still out of date but much better than 2.4 .


ismaell:
Is the build compiled against cygwin?

linsui:
I do not know more about this build...

PR #613, opened 2019-11-26
cormanlisp: Add version 3.1.2b

bb010g:
Adds in Corman Lisp with all the proper initialization from the MSI installer.

The icon has to be downloaded separately because it's a byte array in the Icon table in the MSI, but the MSI extractor deletes that before we can do anything with it in `pre_` or `post_install`, and it also would be extracted normally via `dark -x`, but #3502 isn't merged yet, and it's present in the `CormanLisp.exe` Portable Executable, but there are no interfaces to `kernel32.dll` FindResource/LoadResource in .NET and extracting and recreating the `.ico` from here would require C#, `DllImport`, and more, or an ungodly section of PowerShell code, and overall we can't have nice things. So we download the `.ico` once again for the REPL shortcut to have the nice icon its supposed to have.

---
This change is [Reviewable](https://reviewable.io/reviews/scoopinstaller/main/613)

### DIFF
--- a/bucket/cormanlisp.json
+++ b/bucket/cormanlisp.json
@@ -1,0 +1,43 @@
+{
+    "version": "3.1.2b",
+    "description": "A Common Lisp development environment integrated with Windows.",
+    "homepage": "https://github.com/sharplispers/cormanlisp",
+    "license": "MIT",
+    "suggest": {
+        "vcredist": "extras/vcredist2015"
+    },
+    "url": "https://github.com/sharplispers/cormanlisp/releases/download/v3.1.2b/CormanLisp-3.1.2b.zip",
+    "hash": "8bf722bd919a69ff743e2ae6448fcb3ccbc8be8b3ef746e6acc5cdcb1853ea95",
+    "extract_dir": "Corman Lisp",
+    "bin": [
+        "clconsole.exe",
+        "clconsoleapp.exe"
+    ],
+    "env_set": {
+        "CORMANLISP_HOME": "$dir"
+    },
+    "shortcuts": [
+        [
+            "CormanLisp.exe",
+            "Corman Lisp IDE"
+        ],
+        [
+            "documentation\\CormanLisp.pdf",
+            "Corman Lisp Manual"
+        ],
+        [
+            "clconsole.exe",
+            "Corman Lisp REPL",
+            "",
+            "CormanLisp.exe"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/sharplispers/cormanlisp",
+        "regex": "tree/v([\\w.]+)/"
+    },
+    "autoupdate": {
+        "url": "https://github.com/sharplispers/cormanlisp/releases/download/v$version/CormanLisp-$version.zip"
+    },
+    "notes": "Please restart your command line for CORMANLISP_HOME to take effect. Consider installing <https://www.quicklisp.org/>."
+}


Ash258:
Why did you add that bins?? What are each executable for?

bb010g:
Thanks for the tip on the icon, didn't think to try that before.

From the User Guide:

> # 6  Files in This Release
>
> When you install Corman Lisp, you selected an installation directory. We refer to this as the Corman Lisp directory. There are subdirectories located within this directory, in which many of the Corman Lisp files are stored.
>
> Files located in the Corman Lisp directory:
>
> - **`CormanLisp.exe`**
>
>   This is the Corman Lisp Integrated Development Environment (IDE) application. Double click on it to launch Corman Lisp.
>
> - **`CormanLispServer.dll`**
>
>   The Corman Lisp kernel. It should be located in the Corman Lisp installation directory. It is a COM server for the Corman Lisp system, and will be registered as such by your system the first time you run Corman Lisp.
>
> - **`CormanLisp.img`**
>
>   This contains the remainder of the Corman Lisp system. It is a Lisp image file, produced by loading the Corman Lisp system and using `SAVE-IMAGE`.
>
> - **`CLConsole.exe`**
>
>   This is a simple console application which allows you to interact with Corman Lisp via a console interface. We expect you will find the IDE much nicer to use than this. However, this is an example of how the Corman Lisp COM server may be used to interact with Corman Lisp from another application. Source (in C++) is provided in the `clconsole` directory. In addition to being used by itself to run a lisp console, this file also serves as a template application to generate console-based applications using `SAVE-APPLICATION`.
>
> - **`CLConsoleApp.exe`**
>
>   This file is similar to `CLConsole.exe`, except that it contains the Corman Lisp kernel (**`CormanLispServer.dll`**) linked as a static library. This file serves as a template application to generate applications using `SAVE-APPLICATION`.
>
> - **`CLBoot.exe`**
>
>   This is a simple Windows application which has no user interface. It may be used to load and run a Lisp image, but the Lisp program must handle all the input and output by itself. This file serves as a template application to generate applications using `SAVE-APPLICATION`.
>
> - **`CLBootApp.exe`**
>
>   This file is similar to `CLBoot.exe`, except that it contains the Corman Lisp kernel (`CormanLispServer.dll`) linked as a static library. This file serves as a template application to generate applications using `SAVE-APPLICATION`, with the standalone option set to true.
>
> - **`Init.lisp`**
>
>   This file always gets executed when the lisp top level loop starts. This typically is used to customize the console or IDE by executing lisp expressions at start. It should be noted that this file gets executed by the every instance of the Corman Lisp on the system. Users are encouraged to use an user specific ***initialization file*** whenever possible (see the next section for details).

I just forgot to add them at pull request time, after dropping in `CormanLisp.exe` for initial testing. These are all user-facing commands that would be launched from the path.

As for the manual PDF shortcut, the upstream installer does this as well. Does scoop normalize on not linking these?

bucket/cormanlisp.json
```diff
@@ -0,0 +1,42 @@
...
+    "license": "MIT",
+    "url": "https://github.com/sharplispers/cormanlisp/releases/download/v3.1.2b/CormanLisp-3.1.2b.zip",
+    "hash": "8bf722bd919a69ff743e2ae6448fcb3ccbc8be8b3ef746e6acc5cdcb1853ea95",
+    "extract_dir": "Corman Lisp",
+    "bin": [
+        "clboot.exe",
+        "clbootapp.exe",
+        "clconsole.exe",
+        "clconsoleapp.exe",
+        "CormanLisp.exe"
```
Ash258: There is no reason to keep this in path since it has no command line handler

```suggestion
```

bucket/cormanlisp.json
```diff
@@ -0,0 +1,42 @@
+{
+    "version": "3.1.2b",
+    "description": "Corman Lisp development environment for Common Lisp",
```
Ash258: Do not mention application name. It is obsolete

```suggestion
    "description": "Common Lisp development environment",
```

bb010g:
@Ash258 This is a runtime, just like SBCL, Python, Perl, or PowerShell. It just happens to come with an IDE because it's a Common Lisp, and they tend to do that, especially the commercial ones. For packages with similar functionality in main, you've got 7-Zip (with 7zFM), Coq (with CoqIDE), LessMSI (with Less MSIérables), MySQL (with MySQLWorkbench), Nmap (with Zenmap), PostgreSQL (with pgAdmin), R (with Rgui), Racket (with DrRacket), and Vim/Neovim (with gVim/nvim-qt). (I'm surprised now you don't package the normal Windows IDLE GUI with Python, given how it's normal for other languages.)

bucket/cormanlisp.json
bb010g: Thanks, missed that originally. (That's what you get for packaging at 2:00 AM.)

bb010g:
Bump?

For an example of why this is `main` and not `extras`:

```
PS C:\Users\bb010g> clconsole.exe -execute '.\Documents\Corman Lisp\foo.lisp'                                                                                                                                                                                                 ;; Corman Lisp 3.1  (Patch level 2)
;; Copyright (c) Corman Technologies Inc. See LICENSE.txt for license information.
;; User: bb010g.
Type :quit to exit.
?Hello, World!"Hello, World!"
```

(The console needs work, but `clboot.exe` and `clbootapp.exe` are (meant to be) equivalent to SBCL's `sbcl.exe`.)

Upon further investigation, `clboot.exe` and `clbootapp.exe` aren't user-facing (in the same way; path resolution is handled by your Corman install normally for those in `CCL:SAVE-APPLICATION`) and have been removed from `bin`.

bucket/cormanlisp.json
```diff
@@ -0,0 +1,43 @@
...
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/sharplispers/cormanlisp",
+        "regex": "tree/v([\\w.]+)/"
+    },
+    "autoupdate": {
+        "url": "https://github.com/sharplispers/cormanlisp/releases/download/v$version/CormanLisp-$version.zip"
+    },
+    "notes": "Please restart your command line for CORMANLISP_HOME to take effect. Consider installing <https://www.quicklisp.org/>."
```
Ash258: > Consider installing <https://www.quicklisp.org/>

Why?

bb010g: This is copied from the SBCL notes. QuickLisp is the primary library manager for Common Lisp (think pip/PyPI or Cargo/crates.io).

PR #566, opened 2019-11-12
starship: Add version 0.26.4

matchai:
Adds starship to Scoop.
Please let me know if anything is missing from the manifest. 👍

PR #615, opened 2019-11-27
starship: Add hash extraction

matchai:
Howdy Scoop team! 👋 

Following the advice given in #566, we have gone ahead and added shasums to our releases in https://github.com/starship/starship/pull/649. This PR uses those hashes for verification.

Cheers

### DIFF
--- a/bucket/starship.json
+++ b/bucket/starship.json
@@ -17,6 +17,9 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/starship/starship/releases/download/v$version/starship-x86_64-pc-windows-msvc.zip"
+            },
+            "hash": {
+                "url": "$url.sha256"
             }
         }
     },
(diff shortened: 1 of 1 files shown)


bucket/starship.json
```diff
@@ -17,6 +17,9 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/starship/starship/releases/download/v$version/starship-x86_64-pc-windows-msvc.zip"
+            },
+            "hash": {
```
Ash258: Move this property one prop above.

```json
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/starship/starship/releases/download/v$version/starship-x86_64-pc-windows-msvc.zip"
            }
        },
        "hash": {
            "url": "$url.sha256"
        }
    },
```

matchai: Whoops! Will do. 👍

PR #629, opened 2019-12-03
psutils: Fix checkver

### LINKED ISSUE #625: Why the ln always update to version 0.2013.08.10 & to version 0.2018.07.25
[History for Main/bucket/ln.json](https://github.com/ScoopInstaller/Main/commits/53b89a3d94ffee23b2054f50c8665fe4e556b163/bucket/ln.json)
Is it a bug?

niheaven:
Fix #625

`shim` is moved from `extras`, ref https://github.com/lukesampson/scoop-extras/pull/3298

### DIFF
--- a/bucket/gitignore.json
+++ b/bucket/gitignore.json
@@ -7,8 +7,8 @@
     "hash": "733e4e2bece6e84ad1d25589cc6d1f852587f2d4a6ffe74b8268435ae229487a",
     "bin": "gitignore.ps1",
     "checkver": {
-        "url": "https://github.com/lukesampson/psutils",
-        "re": "gitignore\\.ps1[\\S\\s]*?/(?<sha>[0-9a-f]{40})[\\S\\s]*?(\\d+)-(\\d+)-(\\d+)",
+        "url": "https://github.com/lukesampson/psutils/commits/master/gitignore.ps1.atom",
+        "regex": "(?s)(\\d+)-(\\d+)-(\\d+).*?(?<sha>[0-9a-f]{40})",
         "replace": "0.${1}.${2}.${3}"
     },
     "autoupdate": {
--- a/bucket/gitignore.json
+++ b/bucket/gitignore.json
@@ -8,7 +8,7 @@
     "bin": "gitignore.ps1",
     "checkver": {
         "url": "https://github.com/lukesampson/psutils/commits/master/gitignore.ps1.atom",
-        "regex": "(?s)(\\d+)-(\\d+)-(\\d+).*?(?<sha>[0-9a-f]{40})",
+        "regex": "(?s)>(\\d+)-(\\d+)-(\\d+)T.*?/(?<sha>[0-9a-f]{40})",
         "replace": "0.${1}.${2}.${3}"
     },
     "autoupdate": {
--- a/bucket/say.json
+++ b/bucket/say.json
@@ -14,7 +14,7 @@
     "bin": "say.ps1",
     "checkver": {
         "url": "https://github.com/lukesampson/psutils/commits/master.atom",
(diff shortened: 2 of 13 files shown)


niheaven:
The atom file is less than 10K while `https://api.github.com/repos/lukesampson/psutils/commits?path=ln.ps1` is more than 30K, so using atom.

bucket/gitignore.json
```diff
@@ -7,8 +7,8 @@
     "hash": "733e4e2bece6e84ad1d25589cc6d1f852587f2d4a6ffe74b8268435ae229487a",
     "bin": "gitignore.ps1",
     "checkver": {
-        "url": "https://github.com/lukesampson/psutils",
-        "re": "gitignore\\.ps1[\\S\\s]*?/(?<sha>[0-9a-f]{40})[\\S\\s]*?(\\d+)-(\\d+)-(\\d+)",
+        "url": "https://github.com/lukesampson/psutils/commits/master/gitignore.ps1.atom",
+        "regex": "(?s)(\\d+)-(\\d+)-(\\d+).*?(?<sha>[0-9a-f]{40})",
```
Ash258: Why `\s`?

niheaven: You mean `(?s)`? This trigger make PowerShell search RegEx in single line mode, so `.` matches new line. Ref: https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference#regular-expression-options

Ash258: What exactly line you want to match in URL?

niheaven: ln.ps1.atom

Ash258: Then edit it to be more strict. This one could get something you do not want. For example if commit message contain `X-X-X`, then it will be matched instead of date

```suggestion
        "regex": ">(?s)(\\d+)-(\\d+)-(\\d+)T.*?/(?<sha>[0-9a-f]{40})",
```

bucket/gitignore.json
```diff
@@ -7,8 +7,8 @@
...
         "replace": "0.${1}.${2}.${3}"
```
Ash258: Just to clarify. Why there is `0.`??

niheaven: Means unreleased version, SemVer's `0.xxx.xxx`.

niheaven:
Add `concfg` and `pshazz`, same `checkver.regex`.

rasa:
Forgive my ignorance, but is the point of the leading `0.`, that we expect to someday move away from date based versions, and have human generated version numbers, `1.`, `2.`, etc?

bucket/say.json
```diff
@@ -13,8 +13,8 @@
     ],
     "bin": "say.ps1",
     "checkver": {
-        "url": "https://api.github.com/repos/lukesampson/psutils/commits",
-        "re": "\"(\\d+)-(\\d+)-(\\d+)",
+        "url": "https://github.com/lukesampson/psutils/commits/master.atom",
```
Ash258: Should not be checked for file directly?

bucket/shasum.json
```diff
@@ -13,8 +13,8 @@
     ],
     "bin": "shasum.ps1",
     "checkver": {
-        "url": "https://api.github.com/repos/lukesampson/psutils/commits",
-        "re": "\"(\\d+)-(\\d+)-(\\d+)",
+        "url": "https://github.com/lukesampson/psutils/commits/master.atom",
```
Ash258: Should not be checked for file directly?

bucket/say.json
niheaven: This one has assistant script, so checking file may cause error.

bucket/shasum.json
niheaven: This one has assistant script, so checking file may cause error.

PR #634, opened 2019-12-03
gnuplot: Update to 5.2.8 (manually)

rasa:


Ash258:
Is there any post about dropping 32bit version? Is it permanent or it is one time change?

rasa:
Nope, just 
https://sourceforge.net/projects/gnuplot/files/gnuplot/5.2.8/
saying
`Check back later for Windows binary packages`

PR #648, opened 2019-12-13
httpie: Add version 2.1.0

Restia666Ashdoll:
https://httpie.org/
https://github.com/jakubroztocil/httpie

### DIFF
--- a/bucket/httpie.json
+++ b/bucket/httpie.json
@@ -1,0 +1,36 @@
+{
+    "homepage": "https://httpie.org/",
+    "description": "HTTPie: a CLI, cURL-like tool with wget-like downloads and more",
+    "version": "1.0.3",
+    "license": "BSD 3-Clause New or Revised",
+    "url": "https://github.com/jakubroztocil/httpie/archive/1.0.3.zip",
+    "hash": "5c51d78f65003c028e6b5699207ab5035731de8fd41b5b363a453f37d6cef2b5",
+    "depends": "main/python",
+    "post_install": [
+        "$currentdir = (Get-Item -Path \".\\\").FullName",
+        "Move-Item $dir\\httpie-$version\\* $dir",
+        "Remove-Item $dir\\httpie-$version",
+        "Set-Location $dir",
+        "python setup.py install",
+        "Set-Location $currentdir"
+    ],
+    "uninstaller": {
+        "script": [
+            "$pythonpath = scoop which python",
+            "$pythondir = $pythonpath -Replace \"python.exe$\", \"\"",
+            "Get-Childitem -path $pythondir\\lib\\site-packages -Filter *.egg -Recurse | where-object {$_.Name -ilike \"httpie*\"} | Remove-Item -Force",
+            "Remove-Item $pythondir\\Scripts\\http-script.py -Force",
+            "Remove-Item $pythondir\\Scripts\\http.exe -Force",
+            "Remove-Item $pythondir\\Scripts\\pygmentize-script.py -Force",
+            "Remove-Item $pythondir\\Scripts\\pygmentize.exe -Force",
+            "Remove-Item $pythondir\\Scripts\\chardetect-script.py -Force",
+            "Remove-Item $pythondir\\Scripts\\chardetect.exe -Force"
+        ]
+    },
+    "checkver": {
+        "github": "https://github.com/jakubroztocil/httpie"
+    },
+    "autoupdate": {
+        "url": "https://github.com/jakubroztocil/httpie/archive/$version.zip"
+    }
+}
--- a/bucket/httpie.json
+++ b/bucket/httpie.json
@@ -1,19 +1,21 @@
 {
     "homepage": "https://httpie.org/",
     "description": "HTTPie: a CLI, cURL-like tool with wget-like downloads and more",
-    "version": "2.0.0",
+    "version": "2.1.0",
     "license": "BSD 3-Clause New or Revised",
-    "url": "https://github.com/jakubroztocil/httpie/archive/2.0.0.zip",
-    "hash": "0c09f7b19b7704c0bf5ab12401327a32d95383b87d57dc7326fb24dcbdbd7025",
+    "url": "https://github.com/jakubroztocil/httpie/archive/2.1.0.zip",
+    "hash": "0ed60fbded523ce48575c722ce68dffbce69681dbfc8c743ff292a11807de50d",
     "depends": "main/python",
-    "post_install": [
-        "$currentdir = (Get-Item -Path \".\\\").FullName",
-        "Move-Item $dir\\httpie-$version\\* $dir",
-        "Remove-Item $dir\\httpie-$version",
-        "Set-Location $dir",
-        "python setup.py install",
-        "Set-Location $currentdir"
-    ],
+    "installer": {
+        "script": [
+            "$currentdir = (Get-Item -Path \".\\\").FullName",
+            "Move-Item $dir\\httpie-$version\\* $dir",
+            "Remove-Item $dir\\httpie-$version",
+            "Set-Location $dir",
+            "python setup.py install",
+            "Set-Location $currentdir"
+        ]
+    },
     "uninstaller": {
         "script": [
             "$pythonpath = scoop which python",
@@ -21,6 +23,8 @@
             "Get-Childitem -path $pythondir\\lib\\site-packages -Filter *.egg -Recurse | where-object {$_.Name -ilike \"httpie*\"} | Remove-Item -Force",
             "Remove-Item $pythondir\\Scripts\\http-script.py -Force",
             "Remove-Item $pythondir\\Scripts\\http.exe -Force",
+            "Remove-Item $pythondir\\Scripts\\https-script.py -Force",
+            "Remove-Item $pythondir\\Scripts\\https.exe -Force",
             "Remove-Item $pythondir\\Scripts\\pygmentize-script.py -Force",
             "Remove-Item $pythondir\\Scripts\\pygmentize.exe -Force",
             "Remove-Item $pythondir\\Scripts\\chardetect-script.py -Force",
--- a/bucket/httpie.json
+++ b/bucket/httpie.json
@@ -1,10 +1,10 @@
 {
     "homepage": "https://httpie.org/",
     "description": "HTTPie: a CLI, cURL-like tool with wget-like downloads and more",
-    "version": "1.0.3",
+    "version": "2.0.0",
     "license": "BSD 3-Clause New or Revised",
-    "url": "https://github.com/jakubroztocil/httpie/archive/1.0.3.zip",
-    "hash": "5c51d78f65003c028e6b5699207ab5035731de8fd41b5b363a453f37d6cef2b5",
+    "url": "https://github.com/jakubroztocil/httpie/archive/2.0.0.zip",
+    "hash": "0c09f7b19b7704c0bf5ab12401327a32d95383b87d57dc7326fb24dcbdbd7025",
     "depends": "main/python",
     "post_install": [
         "$currentdir = (Get-Item -Path \".\\\").FullName",


sinapis:
What needs to be done to get this PR approved?

niheaven:
Why not use `pip` to install `httpie`?

equinox:
I think we're all aware scoop is a convenient way to install and keep software up to date with little effort. Why does the fact it's available via pip matter here? Why slap on the label "wontfix" when it's not even clear what issue you (the scoop maintainers) have with it are?

Restia666Ashdoll:
pip is inconvenient. Worst package manager I have ever used. They dont even have an update all command.

Ash258:
When you accept this, there will be other users requesting other "applications" from other package managers. 

Maintaining wrapper around pip, npm, ... is completely waste of time, since they are specially made for that language. 

So this should never been in any official bucket.

Want it? Create custom bucket

If you want it in official bucket, requests compiled binary from author.

sinapis:
> If you want it in official bucket, requests compiled binary from author.

Is this one of the conditions to get something inside the main bucket? I failed to find the list of conditions. Can you kindly refer me to them? What else is needed?

Ash258:
https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket

Technically this could be added into main bucket. But as stated above, supplying/mimicking package managers functionally for specific languages is complete nonsense.

- You are relying on lots of manual checking
	- Already provided by the language specific installers
- You have to be sure to everything works well with all of the possible versions of language
- Pip is still leading standard for Python packages
	- There is anything more scoop could provide to enhance pip

Accepting this will flood main bucket with dozens of other "applications", which are already available in language specific managers.

The time spent on custom wrapper about pip-like installation is not worth the time and sanity of collaborators.
The project owner (of httpie in this case) already support pip out of the box. No reason to waste time on `scoop install httpie`, which could require lots of manual work to be compatible with all python versions, updates, updates of python is completely waste of time when the author itself provide top grade support for `pip install httpie` with all needed configuration already happening.

Same for other LS package managers (lukesampson/scoop#3924). NPM, gem, pip, cargo, cpan, pub.

As I said. You want it? Start a new bucket for it.

Starting own bucket with autoudpate and everything, consists of two clicks.
Create bucket `Scoop-pip` and add it there.

Restia666Ashdoll:
Its not the same thing. pip uses some whl files, this one uses `python setup.py install`
Homebrew seems to do this
https://docs.brew.sh/Python-for-Formula-Authors
They explain the difference between pip and setup.py(they use the latter)
It downloads this
https://files.pythonhosted.org/packages/e2/79/da6aec7b4356e8b325561b987c940e5b1e4de1200a5c3db7c57a97d61ca1/httpie-2.1.0.tar.gz
And invokes setup.py.

sinapis:
I think scoop-python will be a great idea. Specifically, httpie is an awesome developer tool and the fact that scoop does not support it out of the box is a huge miss IMHO.

Someone built an exe from httpie. he's not the original author though... Is there any way we can include that in the main bucket? I can host it on Azure Blob or some other cloud provider.

https://github.com/jftuga/HTTPie-Windows

TheRandomLabs:
I've taken a shot at a [Scoop-Python](https://github.com/TheRandomLabs/Scoop-Python) bucket. I've [rewritten](https://github.com/TheRandomLabs/Scoop-Python/blob/master/bucket/httpie.json) @Restia666Ashdoll's manifest and made some improvements. I've also added some other packages I found as a proof of concept. :)

Restia666Ashdoll:
@TheRandomLabs Thanks for creating the bucket
shouldnt there be a '-py' at the end for things like this
https://gist.github.com/Restia666Ashdoll/ff0404f93ac26eade6e2b581e0e3a2d2
Also, considering adding setuptools so that we can update it
https://github.com/pypa/setuptools
And virtualenv
https://github.com/pypa/virtualenv
But the most important point - Scoop does not retain python packages after update
https://github.com/lukesampson/scoop/issues/2180
So, maybe we should limit this bucket to users only.

TheRandomLabs:
I believe the `-np` at the end of the package names in the `nonportable` bucket is there to differentiate them from the portable versions. I think such a suffix would be redundant in this bucket.
I am aware of lukesampson/scoop#2180, but considering that installations are local by default anyway, I don't think it hurts to have a global installation mode, especially if the issue is eventually fixed. The only problem with local installations is #772.
I'll look at setuptools and virtualenv. :)

Restia666Ashdoll:
@TheRandomLabs Those who have a laptop/pc with no gaming related stuffs and dont want VCRedist in their PC will have to go with `pip install youtube-dl`. This python dependent version of youtube-dl is the one present in pip, OS X and Linux.
Setuptools is required for `python setup.py install`. It comes built in with Python but is updated with pip.

TheRandomLabs:
Understood. I've added youtube-dl to the bucket and changed the naming scheme as you suggested, and I am working on setuptools now.

Restia666Ashdoll:
[#772](https://github.com/ScoopInstaller/Main/issues/772) links to the user Site-Packages path rather than the user Scripts path (`(Split-Path -Parent (python -m site --user-site)) + "\Scripts"`).
To use this bucket the user scripts path has to be first added to the path(otherwise it wont install properly).
To add the user scripts directory to path and create it(if it does not exist)
```
$env:userpath = [System.Environment]::GetEnvironmentVariable("Path","User")
$scriptspath = (Split-Path -Parent (python -m site --user-site)) + "\Scripts"
$env:userpath += ";$scriptspath"
[environment]::setEnvironmentVariable("Path",$env:userpath,"User")
mkdir $scriptspath
```
Conversely to remove it
```
 $env:userpath = [System.Environment]::GetEnvironmentVariable("Path","User")
$scriptspath = (Split-Path -Parent (python -m site --user-site)) + "\Scripts"
$env:userpath = ($env:userpath.Split(';') | Where-Object { $_ -ne $scriptspath }) -Join ';'
[environment]::setEnvironmentVariable("Path",$env:userpath,"User")
```
As of Python 3.8.3 some other variables are
```
$pythonpath = (Split-Path -Parent (python -m site --user-site))
Returns 'C:\Users\<username>\AppData\Roaming\Python\Python38'
$versionshort = (Split-Path -Leaf (Split-Path -Parent (python -m site --user-site)))
Returns 'Python38'
$version = (python --version) -Replace "Python ", ""
Returns '3.8.3'
$pythondir = (Split-Path -Parent (Split-Path -Parent (python -m site --user-site)))
Returns 'C:\Users\<username>\AppData\Roaming\Python'
```
And here comes the main issue. When a completely new Python version(like 3.9) comes out, you will have to create a new Scripts directory. If it depends simply on python, then transfer it to the new Scripts directory. If it depends on the older version(3.8), then keep it.
Unless Scoop lets you transfer modules for global installs this has to be figured out.

TheRandomLabs:
You're right, Python updates will be problematic one way or another. If only there were a way to hook them...

TheRandomLabs:
Could migration of user-specific modules be put in the post-install script for Python? And could global installs also be persisted?

Restia666Ashdoll:
It can be done in the python manifest, but only for user installs. You know the $pythonpath variable I defined. The uninstaller script will record it as $pythonpath1 and post-install as $pythonpath2. If $pythonpath1 and $pythonpath2 are not same, a major Python version change has occurred and everything in $pythonpath1 has to be transferred to $pythonpath2. Packages meant for specific Python version should be restricted to global install. 
But this creates the issue with uninstall ? How do you uninstall now ?

TheRandomLabs:
It would be kinda cool if there were a way to differentiate between updates and actual uninstalls in the uninstaller script.

Restia666Ashdoll:
There is a way for uninstallation. As I specified in my previous comment, contents in $pythonpath will be moved during major version update and PATH will be updated. Packages which depend on an older Python version can only be globally installed to that Python version.
There are two components we need to remove. First is the site-package. This requires using the name Python used for the package. For setuptools its setuptools, for youtube-dl its youtube_dl. This can be found in setup.py.
```
setup(
    name='youtube_dl',
```
Its stored as $packagename
```
$packagename = "setuptools"
$packagepath = (python -c "import $packagename as _; print(_.__path__)")
$packagepath = $packagepath.Substring(2)
$packagepath = $packagepath.Substring(0,$packagepath.Length-2)
$packagepath = (Split-Path -Parent $packagepath)
$packagepath = $packagepath.Substring(0,$packagepath.Length-1)
```
This can be a file or a folder. It has to be removed
The next are the executables in Scripts directory(`(Split-Path -Parent (python -m site --user-site)) + "\Scripts"`) and the python script they come with. For this
We use `(Get-Content ".\installed_files.txt") -Match ".exe$" | Split-Path -LeafBase`. For httpie it returns.
```
.\
❯ (Get-Content ".\installed_files.txt") -Match ".exe$" | Split-Path -LeafBase
http
https
pygmentize
chardetect
```
Everything in Scripts directory(`(Split-Path -Parent (python -m site --user-site)) + "\Scripts"`) containing this in the filename has to be removed.

TheRandomLabs:
I don't believe you're taking into account package dependencies. Would it not be easier to simply copy or move everything in `installed_files.txt`?

Restia666Ashdoll:
I dont think Scoop handles dependencies the same way pip does.
All I mentioned was to uninstall after a major version change. You can move everything `installed_files.txt`, but what about uninstall or update ? We just have to change the way to look for site-package and executables. This assumes that during major Python version update we move all packages in user directory to the new user directory(say Python39). So, I was not really talking about moving but uninstalling or updating henceforth without depending on the full file locations provided by `installed_files.txt`.
So, to summarise
1. Move every user packages from Python 3X to Python3Y during a major upgrade.
2. Uninstall without depending on the file path specified in `installed_files.txt`.
Ofcourse, packages depending on older python versions should be global install.

TheRandomLabs:
I think an alternative approach should be taken, especially taken into consideration [this](https://stackoverflow.com/a/1315764/5076824). Perhaps the `python` bucket should provide a script that automatically checks to see if any packages need to be reinstalled?

Restia666Ashdoll:
I guess you do need to run `python setup.py install` on all packages again.

TheRandomLabs:
I don't see any better solution. :P

TheRandomLabs:
Can I have your thoughts on my [attempt at a migration script](https://github.com/TheRandomLabs/Scoop-Python/blob/master/scripts/migrate-python-packages.ps1)?

Restia666Ashdoll:
- If Python updates, run setup-py on global apps and rewrite `installed_files.txt`
- If `(Split-Path (python -m site --user-site))`  changes, update all user apps and rewrite `installed_files.txt`.
This is what we need. btw this wont run on Scoop's recommended execution policy. It needs to be passed with `-ExecutionPolicy Bypass`.

TheRandomLabs:
Yep, and that's what my script does. A shim would fix the execution policy, right?

Restia666Ashdoll:
I guess it would.

TheRandomLabs:
I've added a separate manifest for it as well. :)

Restia666Ashdoll:
To be frank `& "$(Split-Path (Split-Path (Get-Command scoop).Source))\buckets\python\scripts\migrate-python-packages.ps1"` does work perfectly fine. I guess it existing as a separate package makes the command shorter(and easier to remember). My comment on execution policy was because I didnt check your readme and tried to download and then run the script.

Restia666Ashdoll:
@TheRandomLabs With spotdl-py you have now run into the issue of packages that need older Python versions. spotdl needs Python 3.7 and it wont uninstall. It should depend on python@3.7.7.

TheRandomLabs:
Strange, I've got spotdl running fine on 3.8.

Restia666Ashdoll:
It wont uninstall at all, keeps throwing an error. I thought it was python 3.7 only because pip does not mention 3.8 for spotdl.
Also, try adding all the dependencies to the bucket. since they wont uninstall otherwise. For httpie
https://github.com/pygments/pygments
https://github.com/tartley/colorama
https://github.com/chardet/chardet
https://github.com/andrewp-as-is/request.py
For spotdl
```
    install_requires=[
        "pathlib >= 1.0.1",
        "youtube_dl >= 2017.9.26",
        "pytube3 >= 9.5.5",
        "spotipy >= 2.12.0",
        "mutagen >= 1.41.1",
        "beautifulsoup4 >= 4.6.3",
        "unicode-slugify >= 0.1.3",
        "coloredlogs >= 14.0",
        "lyricwikia >= 0.1.8",
        "PyYAML >= 3.13",
        "appdirs >= 1.4.3",
        "tqdm >= 4.45.0"
    ],
```

TheRandomLabs:
What error is it throwing? Also, I think it'd be easier to just use `pip` to uninstall the dependencies.

Restia666Ashdoll:
```
❯ scoop uninstall spotdl-py
Uninstalling 'spotdl-py' (2.0.5).
Running uninstaller script...
Option --match requires an argument.
Usage: uninstall-python-package <name> [options]
uninstall-python-package --help for for more information
Unlinking C:\Users\<username>\Scoop\apps\spotdl-py\current
'spotdl-py' was uninstalled.
```
If all dependencies are present in Scoop, it will be easier to manage and uninstall.

TheRandomLabs:
Sorry, that was my bad. Uninstallation should be fixed now for `spotdl-py`.
It might be *marginally* easier, but it would result in the addition of too many manifests. The removal of dependencies is probably best left to `pip`.

Restia666Ashdoll:
How about `pip uninstall <dependency package>`  in the uninstall script ? Ofcourse this should exclude the packages present in bucket like `youtube-dl-py` and `pip-py` will now be a dependency for every package.

TheRandomLabs:
I don't think it's viable to uninstall dependencies. What happens if another package is installed that also depends on those dependencies? `pip` also does it this way.

Restia666Ashdoll:
Thanks for all the efforts. Hope it eventually becomes an official bucket.

Restia666Ashdoll:
@TheRandomLabs Have you taken a look at tensorflow ?
https://github.com/tensorflow/tensorflow
Github releases are a mess, so latest version number has to be picked from pypi page
https://pypi.org/project/tensorflow/
setup.py is in "tensorflow\tools\pip_package"

TheRandomLabs:
I'll look at it, but I think it would be a good idea to open issues in my repo instead of spamming this thread.

Restia666Ashdoll:
You have closed the issue tab in your repo.

TheRandomLabs:
Oh, damn, I didn't realise!

TheRandomLabs:
I've enabled the Issues tab, but I can't find any convenient way to add Tensorflow to the bucket.

Restia666Ashdoll:
No problem. Its mostly used with pip anyway. I just wondered if it was possible. Its not that important and not worth the effort.

PR #658, opened 2019-12-19
aws-iam-authenticator: Fix checkver

niheaven:
#150

### DIFF
--- a/bucket/aws-iam-authenticator.json
+++ b/bucket/aws-iam-authenticator.json
@@ -1,12 +1,15 @@
 {
-    "homepage": "https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html",
+    "homepage": "https://github.com/kubernetes-sigs/aws-iam-authenticator",
     "description": "A tool to use AWS IAM credentials to authenticate to a Kubernetes cluster",
     "license": "Apache-2.0",
     "version": "2019-08-22",
     "url": "https://amazon-eks.s3-us-west-2.amazonaws.com/1.14.6/2019-08-22/bin/windows/amd64/aws-iam-authenticator.exe",
     "hash": "9014193536bc97b79bf8d5540954548704453648bba63d74ac63e0b0f32aab49",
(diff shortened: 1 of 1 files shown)


tynril:
**TL;DR**: The releases are also on GitHub: https://github.com/kubernetes-sigs/aws-iam-authenticator/releases -- this might allow simplifying the version check.

The current version on Scoop (`2019-08-22`) seems to map to `0.4.0`: running `aws-iam-authenticator version` returns `{"Version":"v0.4.0","Commit":"c141eda34ad1b6b4d71056810951801348f8c367"}`, which is the correct commit for release `v0.4.0` on the aforementioned GitHub.

That being said, the executable that Scoop currently downloads is not the same as the one the GitHub release provides:
```
λ ls | Get-FileHash -Algorithm SHA256 | Format-Table

Algorithm       Hash                                                                   Path
---------       ----                                                                   ----
SHA256          9014193536BC97B79BF8D5540954548704453648BBA63D74AC63E0B0F32AAB49       aws-iam-authenticator.exe
SHA256          D26CCD0E83667F79DEA88A72072371FD87C2023895417B54C2BD0E4C55D7D5FA       aws-iam-authenticator_0.4.0_windows_amd64.exe
```

And while they are both `0.4.0`, they use two different commits (with `9999cd` being the one right before `c141eda`):
```
λ .\aws-iam-authenticator.exe version
{"Version":"v0.4.0","Commit":"c141eda34ad1b6b4d71056810951801348f8c367"}
~\Desktop
λ .\aws-iam-authenticator_0.4.0_windows_amd64.exe version
{"Version":"0.4.0","Commit":"9999cd8047782c94c08e2d24b77dc9488947aa03"}
```

Presumably, Amazon uploaded a different build made on the previous commit to the repository for their documentation. I've used the GitHub version for a bit and everything behaves as expected (which is to say, the same way as the version currently on Scoop).

tynril:
I just made a separate PR for the same issue using GitHub instead of a regex: #668

PR #667, opened 2019-12-20
b2sum: Add version 20130305

niheaven:


linsui:
> All code is triple-licensed under the CC0, the OpenSSL Licence, or the Apache Public License 2.0, at your choosing.

See https://github.com/BLAKE2/BLAKE2

niheaven:
> BLAKE2 is specified in RFC 7693, and our code and test vectors are available on GitHub, licensed under CC0 (public domain-like). BLAKE2 is also described in the 2015 book The Hash Function BLAKE. 

See homepage.

PR #680, opened 2019-12-26
neofetch: remove dependency on git installed through scoop

xeijin:
currently, if I don't have `git` or `git-with-openssh` installed through scoop I get a confusing error indicating `git` couldn't be found, even though it is right there on my path

there are many apps that come bundled with git now, including github desktop, Cmder etc. it would be better for check for these on path as a last resort, and work out the location to `bash.exe` from there.


rasa:
This patch is valuable, as I had to recently uninstall scoop's git, as when `git --exec-path` was run, it spawned 1000's of gits until the system came to a grind. When I uninstalled scoop's git, and installed git directly, the problem went away, so I'm assuming the shim call was the issue.

xeijin:
who can actually review it then?

rashil2000:
Actually, my bad, this was already fixed in #2729

PR #685, opened 2019-12-29
haxe, neko: fix env path

linsui:
fix https://github.com/lukesampson/scoop/issues/3808


linsui:
Should we remove haxe-dev? It only matches pre-release versions which are sometimes even older than latest release versions.

rishavs:
Can we tag the reviewers? I need to install Haxe and would love to do it via scoop.

rishavs:
Thank you! <3

rishavs:
@linsui This did not fix the original issue #3808

Haxelib still cannot find neko.dll

linsui:
It works for me. Did you reinstall neko?

rishavs:
i didn't. I will try right away and report back

rishavs:
ok. Can confirm that its working fine. Thanks!

PR #687, opened 2019-12-29
BatCodeCheck: Add version 0.32.1

issaclin32:
**BatCodeCheck** ([manual](https://www.robvanderwoude.com/helptext.php?src=batcodecheck_cs_help)) is a linter for Windows batch file **(.bat)**.

Notes:
* **License** info can be found in the *Disclaimer* section of the homepage(https://www.robvanderwoude.com/).
> Unless stated otherwise, these pages, the scripts presented and their sources are copyrighted freeware.
You may modify them, as long as a reference to the original code is included in the modified code.

>However, it is not allowed to publish (copies of) my scripts on your own site, or distribute them on paper, CD or by whatever medium, without my prior written consent (hence copyrighted freeware).

### DIFF
--- a/bucket/batcodecheck.json
+++ b/bucket/batcodecheck.json
@@ -1,0 +1,23 @@
+{
+    "version": "0.32.1",
+    "description": "Batch file (.bat) linter",
+    "homepage": "https://www.robvanderwoude.com",
+    "license": {
+        "identifier": "Freeware",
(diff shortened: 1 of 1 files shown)


issaclin32:
> * Based on words of author, this is not suitable for main bucket

I will re-open this PR in the *extras* bucket.

PR #702, opened 2020-01-04
gnuplot: Add shortcut

dodorz:


linsui:
https://spdx.org/licenses/gnuplot.html gnuplot is in spdx list, no need to add url.

PR #707, opened 2020-01-05
swi-prolog: Add version 8.0.3-1

ddavness:
SWI-Prolog is the most popular implementation and runtime for the Prolog programming language.

The package comes with a GUI tool (`swipl-win.exe`), but is not essential (comparable to Python's IDLE). 
In the same spirit of Python's manifest for this bucket, a shortcut was not added.


niheaven:
See https://github.com/lukesampson/scoop-extras/blob/master/bucket/swipl.json

PR #712, opened 2020-01-06
fix: change gopath to $persist_dir\.gopath

deeyi2000:


h404bi:
This could break previously installed packages. I don't see any important reason to change go's default GOPATH. I think you want a fully portable installation but this is not a **fix**.

deeyi2000:
I think GOPATH is important for golang,it has global tools and packages.this is similar to GEM_PATH for ruby, or CARGO_HOME for rust.
In older go project, GOPATH is also used as workspace,(in different project change $GOPATH=xxx;$GOPATH).
so,I think GOPATH is persistent data

PR #715, opened 2020-01-06
gcc: fix checkver/add env_add_path

linsui:


bucket/gcc.json
```diff
@@ -58,8 +58,9 @@
         "bin\\windmc.exe",
         "bin\\windres.exe"
     ],
+    "env_add_path": "bin",
```
h404bi: If you add bin directory the path, there is no need to shim all those bins.

PR #720, opened 2020-01-08
dotnet-sdk: Add path and suggest

myty:
Add env_add_path and suggest values to include required dependencies.

This a duplicate of #253 since it was closed and old.

### DIFF
--- a/bucket/dotnet-sdk.json
+++ b/bucket/dotnet-sdk.json
@@ -14,6 +14,10 @@
         }
     },
     "bin": "dotnet.exe",
+    "suggest": {
+        "Visual C++ Redistributable 2015": "extras/vcredist2015"
+    },
+    "env_add_path": ".",
     "env_set": {
         "DOTNET_ROOT": "$dir",
         "MSBuildSDKsPath": "$dir\\sdk\\$version\\Sdks"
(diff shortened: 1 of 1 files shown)


bucket/dotnet-sdk.json
```diff
@@ -14,6 +14,10 @@
         }
     },
     "bin": "dotnet.exe",
+    "suggest": {
+        "Visual C++ Redistributable 2015": "extras/vcredist2015"
```
niheaven:
```suggestion
        "Visual C++ Redistributable 2019": "extras/vcredist2019"
```

myty: The 2015 version was only included because of this: https://docs.microsoft.com/en-us/dotnet/core/install/dependencies?pivots=os-windows&tabs=netcore31#net-core-dependencies.  Not sure the 2019 version is the right choice here.

niheaven: https://support.microsoft.com/help/2977003/the-latest-supported-visual-c-downloads

myty: Ah. Got ya. So the vcredist2019 will cover the minimum requirements for dotnet sdk. 👍

PR #750, opened 2020-01-24
vector: Add version 0.7.0

Hoverbear:
Adds Vector 0.7.0. 0.6.0 added [Windows support](https://vector.dev/blog/windows-support/).

Unfortunately, 32 bit support is not available.

Tested with `scoop install bucket/vector.json`, `vector`, then finally `scoop uninstall vector`


Hoverbear:
Hey @Ash258 !

Thanks for your feedback, I improved the manifest (I hope) based on your feedback.

It now:

* Uses zips
* Has a bin alias
* Uses persist
* References the config on startup

Sorry for the rocky first commits. :) First manifest for scoop.

bucket/vector.json
```diff
@@ -0,0 +1,30 @@
...
+        "64bit": {
+            "url": "https://github.com/timberio/vector/releases/download/v0.7.0/vector-x86_64-pc-windows-msvc.zip",
+            "hash": "70f5d130b1afe46ddbbe30c979627bd147919bd5bb3eda628448d820b43e1a84"
+        }
+    },
+    "bin": [
+        [
+            "bin\\vector.exe",
+            "vector",
+            "--config $persist_dir\\config\\vector.toml"
```
Ash258:
```suggestion
            "--config",
            "\"$persist_dir\\config\\vector.toml\""
```

Hoverbear:
Hmmm, yeah, you're totally right! Commonly users seem to be deploying them as agents/sidecars so I think it's better not to use this shim!

LucioFranco:
👍 Generally the user should provide the path to the config and we will infer the path if they don't provide it. If that fails then it friendly errors out :) So this change seems good to me!

Hoverbear:
@Ash258 could you take another look?

Ash258:
Cannot do 100% review right now, but for the quick look you can change:

```diff
-    "bin": [
-        "bin\\vector.exe"
-    ]
+    "bin": "bin\\vector.exe",
```

I guess there should not be any issue with installation itself. I only remember that with previous state installation (at first version of manifest with my changes) when I did not feed the `config` parameter it was loading from APPDATA. Is it true or I mistook application?

Hoverbear:
@Ash258 At the moment Vector searches for the `/etc` directory by default, and needs to be configured for Windows. I expect this will improve in a future version. So my goal was just to preserve the default config folder since folks will likely run one or multiple vector instances using files from there.

PR #774, opened 2020-02-02
nodejs: Use bin instead of adding new PATH

BasixKOR:
While this PR doesn't give any new features, I think using `bin` over adding PATH is generally better in this kind of cases. (i.e. Scoop would be able to generate shims for Node.js)

### DIFF
--- a/bucket/nodejs.json
+++ b/bucket/nodejs.json
@@ -3,6 +3,11 @@
     "description": "As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications.",
     "version": "13.7.0",
     "license": "MIT",
+    "bin": [
+        "node.exe",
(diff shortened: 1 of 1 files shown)


BasixKOR:
I'm not sure whether it should keep `nodevars` and `install_tools` or not, so I haven't cleared up `env_add_path` yet. I'd happily remove it if it won't cause any problems!

BasixKOR:
Um, it seems like only node is working with this patch applied. Closing the PR.

Asfolny:
Since this one works, even if the yarn one doesn't, why was this closed exactly?

BasixKOR:
I was unable to run npm with this patch applied, so I thought this wasn't working. Does this patch work for you?

PR #775, opened 2020-02-02
yarn: Use bin property instead of PATH

BasixKOR:
Like #774, this PR adds new `bin`s instead of PATH. But since yarn has only two binaries, I removed `Yarn\\bin` in `env_add_path`.


BasixKOR:
Yarn seems like can't find its path with this. Closing the PR.

Issue #760, opened 2020-01-29
tor@0.4.2.5: download via aria2 failed


konovalov-:
```txt
tor: 0.4.1.6 -> 0.4.2.5
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
Updating one outdated app:
Updating 'tor' (0.4.1.6 -> 0.4.2.5)
Downloading new version
Starting download with aria2 ...
Download: 01/29 04:12:33 [ERROR] CUID#8 - Download aborted. URI=https://dist.torproject.org/torbrowser/9.0.3/tor-win64-0.4.2.5.zip
Download: Exception: [AbstractCommand.cc:351] errorCode=3 URI=https://dist.torproject.org/torbrowser/9.0.3/tor-win64-0.4.2.5.zip
Download:   -> [HttpSkipResponseCommand.cc:219] errorCode=3 Resource not found
Download: Download Results:
Download: gid   |stat|avg speed  |path/URI
Download: ======+====+===========+=======================================================
Download: 0456d5|ERR |       0B/s|/scoop/cache/tor#0.4.2.5#https_dist.torproject.org_torbrowser_9.0.3_tor-win64-0.4.2.5.zip
Download: Status Legend:
Download: (ERR):error occurred.
Download: aria2 will resume download if the transfer is restarted.
Download: If there are any errors, then see the log file. See '-l' option in help/man page for details.
ERROR Download failed! (Error 3) Resource was not found
ERROR https://dist.torproject.org/torbrowser/9.0.3/tor-win64-0.4.2.5.zip
    referer=https://dist.torproject.org/torbrowser/9.0.3/
    dir=\scoop\cache
    out=tor#0.4.2.5#https_dist.torproject.org_torbrowser_9.0.3_tor-win64-0.4.2.5.zip

ERROR & '\scoop\apps\aria2\current\aria2c.exe' --input-file='\scoop\cache\tor.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/6.2 (Windows NT 10.0; Win64; x64; Core)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn --enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=9516 --continue
[...]

Issue #730, opened 2020-01-13
tor@0.4.2.5: download via aria2 failed


tayrayb:
ERROR Download failed! (Error 3) Resource was not found
ERROR https://dist.torproject.org/torbrowser/9.0.3/tor-win64-0.4.2.5.zip
    referer=https://dist.torproject.org/torbrowser/9.0.3/
    dir=C:\Users\Tayray\scoop\cache
    out=tor#0.4.2.5#https_dist.torproject.org_torbrowser_9.0.3_tor-win64-0.4.2.5.zip

ERROR & 'C:\Users\Tayray\scoop\apps\aria2\current\aria2c.exe' --input-file='C:\Users\Tayray\scoop\cache\tor.txt' --user-agent='Scoop/1.0 (+http://scoop.sh/) PowerShell/5.1 (Windows NT 10.0; Win64; x64; Desktop)' --allow-overwrite=true --auto-file-renaming=false --retry-wait=2 --split=5 --max-connection-per-server=5 --min-split-size=5M --console-log-level=warn -
-enable-color=false --no-conf=true --follow-metalink=true --metalink-preferred-protocol=https --min-tls-version=TLSv1.2 --stop-with-process=8964 --continue

Ash258:
08f96585485d10358b0401e31648e4f01ab0db6b

PR #788, opened 2020-02-05
tor@0.4.2.5: Fix url

MCOfficer:
This PR fixes - for the time being - the download URL for tor (see #760, #730 ).

The issue will occur again when the browser is updated, but not tor. In this case, one has to force an Autoupdate to fix the URL.


MCOfficer:
08f96585485d10358b0401e31648e4f01ab0db6b

PR #800, opened 2020-02-11
openssl: Add to PATH to make libraries discoverable

timvisee:
This adds the `bin` directory to `PATH` to make provided libraries discoverable.

This is according to https://github.com/timvisee/ffsend/issues/106#issuecomment-584661412, and is needed because `ffsend` currently fails to find the installed OpenSSL libraries.

I'm not sure whether to increase a version number or anything in this manifest as well, to force-update and apply the `PATH` variable on current systems.

### DIFF
--- a/bucket/openssl.json
+++ b/bucket/openssl.json
@@ -21,6 +21,9 @@
     "env_set": {
         "OPENSSL_CONF": "$dir\\bin\\cnf\\openssl.cnf"
     },
+    "env_add_path": [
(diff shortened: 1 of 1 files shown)


rasa:
Generally, we want to shim exes,, instead of a PATH fox, but here, we have .dlls that other apps need, so we need to add to the PATH, right?

timvisee:
@rasa That is correct.

PR #808, opened 2020-02-14
migrate: Add version 4.9.1

martinlindhe:
This is a very useful CLI when working with databases.
It is used to migrate/rollback database changes over time.

~~The current version (4.8.0) is only released as a 64-bit windows exe, so this bucket only supports 64-bit installs.~~
~~I asked if upstream could provide 32-bit windows binaries in the future in golang-migrate/migrate#339~~


martinlindhe:
migrate 4.9.1 has been released, now with a 32-bit exe. Formula updated

PR #826, opened 2020-02-20
v: Add version 0.2.4

lukeb2e:
Hi,

I would like to add https://vlang.io to the bucket.

There are still some questions with this PR that need to be verified:

- [ ] Name: v is very short. I followed the scheme used by golang here, but vlang might be a saner option (especially while searching).
- [ ] Bucket: is this the correct bucket or should this be added to the extras bucket.

Let me know if one of the above should be changed.

Kind regards


chawyehsu:
> Name: v is very short. I followed the scheme used by golang here, but vlang might be a saner option (especially while searching).

For me, prefer `vlang`, two reasons: 1) better for searching 2) the same as Homebrew (`brew install vlang`)

> Bucket: is this the correct bucket or should this be added to the extras bucket.

There is no doubt it should be in this bucket.

lukeb2e:
Changed the file name as recommended.

wstein (changes requested):
I have install MinGW-w64 like mentioned in notes:

`scoop install MinGW-w64`

I have observed two issues:

1. Please fix bin entry to v/v.exe as v.exe does not exists in base folder
2. After locally fixing issue 1, I am getting this error:
```
cannot compile `D:\scoop\apps\vlang\current\v\cmd\tools\vrepl.v`:
V panic: ls() couldnt open dir "C:\Program Files (x86)\Windows Kits\10\Lib": directory does not exist
print_backtrace_skipping_top_frames is not implemented
```

issaclin32:
> 2. After locally fixing issue 1, I am getting this error:
> 
> ```
> cannot compile `D:\scoop\apps\vlang\current\v\cmd\tools\vrepl.v`:
> V panic: ls() couldnt open dir "C:\Program Files (x86)\Windows Kits\10\Lib": directory does not exist
> print_backtrace_skipping_top_frames is not implemented
> ```

"Windows Kits" (actually named as *Windows SDK* by Microsoft) can be installed by either:
(1) Using the installer [from this page](https://developer.microsoft.com/en-us/windows/downloads/sdk-archive/)
(2) Installing Visual Studio

issaclin32:
I updated the manifest by myself because the PR is pretty outdated.

`checkver` currently fails because there are too many weekly releases on [the release page](https://github.com/vlang/v/releases). This should not be a problem when a new stable version is released.

rashil2000:
I think we can keep the name v.json? There are already 2 manifests like that - r.json (RLang) and s.json (a CLI web search tool).

Actually, the current situation is a bit haphazard. Many manifests do have the -lang suffix (beeflang.json, ziglang.json), and many don't (go.json, rust.json, nim.json).

What do you think?

issaclin32:
> I think we can keep the name v.json? There are already 2 manifests like that - r.json (RLang) and s.json (a CLI web search tool).
> 
> Actually, the current situation is a bit haphazard. Many manifests do have the -lang suffix (beeflang.json, ziglang.json), and many don't (go.json, rust.json, nim.json).
> 
> What do you think?

I agree. We cannot say `v.json` is too short when there is an obvious example of `r.json`.

rashil2000 (changes requested):
Should use the arch 64bit block - the compiler is 64bit.

rashil2000:
> I agree. We cannot say v.json is too short when there is an obvious example of r.json.

Do you think we should rename beeflang and ziglang to beef and zig, for standardization?

issaclin32:
> Do you think we should rename beeflang and ziglang to beef and zig, for standardization?

I think it the right thing to do.

issaclin32:
> Should use the arch 64bit block - the compiler is 64bit.

Done.

rashil2000 (approved):

PR #842, opened 2020-02-26
uutils-coreutils: Add version 0.0.2

### LINKED ISSUE #838: [Request] Add Uutil Coreutils
Cross-platform Rust rewrite of the GNU coreutils

Source: https://github.com/uutils/coreutils
Out-of-date yet functioning binaries: https://github.com/rivy/rust.coreutils/releases

issaclin32:
close #838

**uutils coreutils**([homepage](https://github.com/uutils/coreutils)) is a Rust implementation of GNU coreutils.

Notes:
* https://github.com/rivy/rust.coreutils/releases provides binaries compiled with *GNU GCC* and *MSVC*. ~I seperate them into 2 packages (`-gnu` and `-msvc`).~ use the *MSVC* version for this package.


rasa:
@issaclin32 I approved this, but perhaps one should simply be named uutils-coreutils? Not sure which one makes more sense to do that with?

issaclin32:
> perhaps one should simply be named uutils-coreutils? Not sure which one makes more sense to do that with?

I am not sure about the difference (or pros/cons) between the *msvc* and *gnu* version. 
Since https://github.com/rivy/rust.coreutils/releases only provides the *msvc* version at version **0.01**, I will name the *msvc* version as *uutils-coreutils*.

PR #846, opened 2020-02-28
python: Add registry configuration

### LINKED ISSUE #128: install python with windows registry
reference from [pip-514](https://www.python.org/dev/peps/pep-0514).

maybe:

``` registry
[HKEY_CURRENT_USER\Software\Python\Scoop]
(Default) = (value not set)
DisplayName = "Scoop"
SupportUrl = "https://github.com/lukesampson/scoop"

[HKEY_CURRENT_USER\Software\Python\Scoop\3.7.3]
(Default) = (value not set)
DisplayName = "Python 3.7.3"
SupportUrl = "https://github.com/lukesampson/scoop"
Version = "3.0.12345.0"
SysVersion = "3.7.3"
SysArchitecture = "64bit"

[HKEY_CURRENT_USER\Software\Python\Scoop\3.7.3\InstallPath]
(Default) = "C:\Users\<User>\scoop\apps\python\3.7.3"
ExecutablePath = "C:\Users\<User>\scoop\apps\python\3.7.3\python.exe"
WindowedExecutablePath = "C:\Users\<User>\scoop\apps\python\3.7.3\pythonw.exe"
```

idbrii:
- Closes #128

PEP 514 specifies how python should be added to the registry so
third-party installers, binary packages depending on python (vim), and
pre-compiled binary extension packages (kivy, Pillow, PyGame) can find
python.

I think it also helps resolve python finding 32-bit vs 64-bit
executables.

Seems that python requires the "PythonCore" key. Testing with gvim:
    gvim +"py3 print(3)"
crashes with "Scoop" but works with "PythonCore". We're installing
python built by the Python Software Foundation, so it might be matching
that name somewhere inside of python.exe.

Since the manifest uses github urls, it won't work until merged into master. If you want to check it out, I made [a testing branch that points to itself](https://github.com/idbrii/scoop-bucket-main/tree/py-registry-testing-against-mine).

### DIFF
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -3,16 +3,46 @@
     "description": "A programming language that lets you work quickly and integrate systems more effectively.",
     "license": "Python-2.0",
     "version": "3.8.2",
+    "notes": "Allow applications and third-party installers to find python by running: \"$dir\\install-pep-514.reg\"",
     "architecture": {
         "64bit": {
-            "url": "https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe",
-            "hash": "8e400e3f32cdcb746e62e0db4d3ae4cba1f927141ebc4d0d5a4006b0daee8921"
+            "url": [
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/install-pep-514.reg",
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/uninstall-pep-514.reg",
+                "https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe"
+            ],
+            "hash": [
+                "5d9d7a604c057e67973e1d4b3b488ae20a0e5ee5496d03b5cf091410e33e39ed",
+                "d9309423b693ed63aea6b1fbfc0c34f16842cdca22a1b3edef283b87567a53b9",
+                "8e400e3f32cdcb746e62e0db4d3ae4cba1f927141ebc4d0d5a4006b0daee8921"
+            ]
         },
         "32bit": {
-            "url": "https://www.python.org/ftp/python/3.8.2/python-3.8.2.exe",
-            "hash": "03ac5754a69c9c11c08d1f4d694c14625a4d27348ad4dd2d1253e2547819db2c"
+            "url": [
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/install-pep-514.reg",
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/uninstall-pep-514.reg",
+                "https://www.python.org/ftp/python/3.8.2/python-3.8.2.exe"
+            ],
+            "hash": [
+                "5d9d7a604c057e67973e1d4b3b488ae20a0e5ee5496d03b5cf091410e33e39ed",
+                "d9309423b693ed63aea6b1fbfc0c34f16842cdca22a1b3edef283b87567a53b9",
+                "03ac5754a69c9c11c08d1f4d694c14625a4d27348ad4dd2d1253e2547819db2c"
+            ]
         }
     },
+    "pre_install": [
+        "'install-pep-514.reg', 'uninstall-pep-514.reg' | ForEach-Object {",
+        "    $py_root = \"$dir\".Replace('\\', '\\\\')",
+        "    $py_version = ($version -split '\\.')[0..1] -join '.'",
+        "    $content = Get-Content \"$dir\\$_\"",
+        "    $content = $content.Replace('$py_root', $py_root)",
+        "    $content = $content.Replace('$py_version', $py_version)",
+        "    if ($global) {",
+        "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "    }",
+        "    Set-Content \"$dir\\$_\" $content -Encoding UTF8",
+        "}"
+    ],
     "installer": {
         "script": [
             "Expand-DarkArchive \"$dir\\$fname\" \"$dir\\_tmp\"",
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -40,7 +40,7 @@
         "    if ($global) {",
         "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
         "    }",
-        "    Set-Content \"$dir\\$_\" $content -Encoding UTF8",
+        "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
         "}"
     ],
     "installer": {
--- a/scripts/python/install-pep-514.reg
+++ b/scripts/python/install-pep-514.reg
@@ -1,0 +1,17 @@
+Windows Registry Editor Version 5.00
+
+[HKEY_CURRENT_USER\Software\Python\PythonCore]
+"DisplayName"="Official Python installed with Scoop"
+"SupportUrl"="https://www.python.org/"
+
+[HKEY_CURRENT_USER\Software\Python\PythonCore\$py_version]
+"DisplayName"="Python $py_version"
+
+[HKEY_CURRENT_USER\Software\Python\PythonCore\$py_version\InstallPath]
+@="$py_root"
+"ExecutablePath"="$py_root\\python.exe"
+"WindowedExecutablePath"="$py_root\\pythonw.exe"
+
+[HKEY_CURRENT_USER\Software\Python\PythonCore\$py_version\PythonPath]
+@="$py_root\\Lib\\;$py_root\\DLLs\\"
+
--- a/scripts/python/uninstall-pep-514.reg
+++ b/scripts/python/uninstall-pep-514.reg
@@ -1,0 +1,3 @@
+Windows Registry Editor Version 5.00
+
+[-HKEY_CURRENT_USER\Software\Python\PythonCore]


idbrii:
Works great on my testing branch:

```
> scoop uninstall python & scoop install python
Uninstalling 'python' (3.8.2).
'python' was uninstalled.
Installing 'python' (3.8.2) [64bit]
Loading install-pep-514.reg from cache
Checking hash of install-pep-514.reg ... ok.
Loading uninstall-pep-514.reg from cache
Checking hash of uninstall-pep-514.reg ... ok.
Loading python-3.8.2-amd64.exe from cache
Checking hash of python-3.8.2-amd64.exe ... ok.
Running pre-install script...
Running installer script...
Linking ~\scoop\apps\python\current => ~\scoop\apps\python\3.8.2
Creating shim for 'python'.
Creating shim for 'pythonw'.
Creating shim for 'python3'.
Creating shim for 'idle'.
Creating shim for 'idle3'.
'python' (3.8.2) was installed successfully!
Notes
-----
Allow applications and third-party installers to find python by running: C:\Users\idbrii\scoop\apps\python\current\install-pep-514.reg

> gvim +"py3 print(3)"
# 3 is printed to gvim's message area.
```

bucket/python.json
```diff
@@ -3,14 +3,31 @@
...
     "version": "3.8.2",
+    "notes": "Allow applications and third-party installers to find python by running: $dir\\install-pep-514.reg",
     "architecture": {
         "64bit": {
-            "url": "https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe",
-            "hash": "8e400e3f32cdcb746e62e0db4d3ae4cba1f927141ebc4d0d5a4006b0daee8921"
+            "url": [
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/install-pep-514.reg",
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/uninstall-pep-514.reg",
+                "https://www.python.org/ftp/python/3.8.2/python-3.8.2-amd64.exe"
```
Ash258: First item in array

bucket/python.json
```diff
@@ -3,14 +3,31 @@
...
+                "8e400e3f32cdcb746e62e0db4d3ae4cba1f927141ebc4d0d5a4006b0daee8921"
+            ]
         },
         "32bit": {
-            "url": "https://www.python.org/ftp/python/3.8.2/python-3.8.2.exe",
-            "hash": "03ac5754a69c9c11c08d1f4d694c14625a4d27348ad4dd2d1253e2547819db2c"
+            "url": [
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/install-pep-514.reg",
+                "https://raw.githubusercontent.com/ScoopInstaller/Main/master/scripts/python/uninstall-pep-514.reg",
+                "https://www.python.org/ftp/python/3.8.2/python-3.8.2.exe"
```
Ash258: first item

bucket/python.json
```diff
@@ -28,6 +45,19 @@
             "}"
         ]
     },
+    "pre_install": [
+        "'install-pep-514.reg', 'uninstall-pep-514.reg' | ForEach-Object {",
+        "    $py_root = \"$dir\".Replace('\\', '\\\\')",
+        "    $py_version = [regex]::match($version,'(\\d+.\\d+)').Groups[0].Value",
```
Ash258: You need only `3.8` for example?

bucket/python.json
```diff
@@ -28,6 +45,19 @@
...
+        "    $content = Get-Content \"$dir\\$_\"",
+        "    $content = $content.Replace('$py_root', $py_root)",
+        "    $content = $content.Replace('$py_version', $py_version)",
+        "    if ($global) {",
+        "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "    }",
+        "    Set-Content \"$dir\\$_\" $content",
```
Ash258:
```suggestion
        "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
```

bucket/python.json
```diff
@@ -28,6 +45,19 @@
             "}"
         ]
     },
+    "pre_install": [
```
Ash258: Move pre_install above installer

bucket/python.json
idbrii: Yes. py_version needs to be only major.minor. If it includes micro, then it doesn't work. Is there a smarter way to get that?

bucket/python.json
idbrii: [`installer` requires](https://github.com/lukesampson/scoop/wiki/App-Manifests) either the installer is the last URL or to explicitly specify the name:

> file: The installer executable file. For installer this defaults to the last URL downloaded. Must be specified for uninstaller.


The installer name is different for each architecture and each version. Instead of trying to write confusing logic to make that work, I made it the last one. Since that's a general scoop app restriction, I assumed this would be a common convention?

bucket/python.json
idbrii: Why ascii?

Powershell defaulted to exporting with utf-8 no bom and that correctly imported into regedit. If we switched to ascii, then what would happen if someone had unicode in their username (and scoop install path)?

regedit exports with utf-16le with BOM. The closest [matches I see](https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Management/Set-Content?view=powershell-7) are UTF8BOM and BigEndianUnicode.

I'd prefer to stick to the more common and functioning utf8 instead of something more esoteric.

Ash258: Powershell 6 and 7 default is UTF8 without BOM

Not powershell 5

bucket/python.json
```diff
@@ -3,16 +3,46 @@
...
+                "5d9d7a604c057e67973e1d4b3b488ae20a0e5ee5496d03b5cf091410e33e39ed",
+                "2d637182b7bd4536e70e9bf62ecc59d78ce2427046cb47a45777f449148b8f88",
+                "03ac5754a69c9c11c08d1f4d694c14625a4d27348ad4dd2d1253e2547819db2c"
+            ]
         }
     },
+    "pre_install": [
+        "'install-pep-514.reg', 'uninstall-pep-514.reg' | ForEach-Object {",
+        "    $py_root = \"$dir\".Replace('\\', '\\\\')",
+        "    $py_version = [regex]::match($version,'(\\d+.\\d+)').Groups[0].Value",
```
Ash258:
```suggestion
        "    $py_version = ($version -split '\\.')[0..1] -join '.'",
```

bucket/python.json
```diff
@@ -3,16 +3,46 @@
...
+        "    $content = Get-Content \"$dir\\$_\"",
+        "    $content = $content.Replace('$py_root', $py_root)",
+        "    $content = $content.Replace('$py_version', $py_version)",
+        "    if ($global) {",
+        "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "    }",
+        "    Set-Content \"$dir\\$_\" $content",
```
Ash258:
```suggestion
        "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
```

scripts/python/install-pep-514.reg
```diff
@@ -0,0 +1,17 @@
+Windows Registry Editor Version 5.00
+
+[HKEY_CURRENT_USER\Software\Python\PythonCore]
+"DisplayName"="Official Python installed with Scoop"
```
Ash258: Where is this displayed?

scripts/python/uninstall-pep-514.reg
```diff
@@ -0,0 +1,6 @@
+Windows Registry Editor Version 5.00
+
+[-HKEY_CURRENT_USER\Software\Python\PythonCore]
```
Ash258: Is not this enough?

bucket/python.json
```diff
@@ -3,16 +3,46 @@
     "description": "A programming language that lets you work quickly and integrate systems more effectively.",
     "license": "Python-2.0",
     "version": "3.8.2",
+    "notes": "Allow applications and third-party installers to find python by running: $dir\\install-pep-514.reg",
```
Ash258:
```suggestion
    "notes": "Allow applications and third-party installers to find python by running: '$dir\\install-pep-514.reg'",
```

scripts/python/install-pep-514.reg
idbrii: I'm following [PEP-514]( https://www.python.org/dev/peps/pep-0514/) and including a minimal set of the fields. In theory it's displayed by tools that look up python location in the registry. In practice, I don't have tools that display this name.

bucket/python.json
idbrii: Why wouldn't `-Encoding UTF8` be sufficient?
(I'm using powershell 5.1 and utf8nobom was default.)

idbrii: I see. ps5 outputs bom for `-Encoding utf8`. That's fine. I changed it to that and tested that it correctly imports to regedit.

bucket/python.json
```diff
@@ -3,16 +3,46 @@
...
+        "'install-pep-514.reg', 'uninstall-pep-514.reg' | ForEach-Object {",
+        "    $py_root = \"$dir\".Replace('\\', '\\\\')",
+        "    $py_version = ($version -split '\\.')[0..1] -join '.'",
+        "    $content = Get-Content \"$dir\\$_\"",
+        "    $content = $content.Replace('$py_root', $py_root)",
+        "    $content = $content.Replace('$py_version', $py_version)",
+        "    if ($global) {",
+        "       $content = $content.Replace('HKEY_CURRENT_USER', 'HKEY_LOCAL_MACHINE')",
+        "    }",
+        "    Set-Content \"$dir\\$_\" $content -Encoding UTF8",
```
Ash258: For the third and last time.

Ascii Encoding is best practice in scoop overall as it is compatible and tested for all cultures and all scripts are able to load ascii files without any issues and without hackourounds.
Not all scripts are handling BOM correctly. Just use Ascii and everything will be fine.

```suggestion
        "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
```

idbrii: Okay.

I don't understand what you mean by "all scripts are able to load ascii files" since it's not a script that would be loading this file: It's the Windows Registry. But you have more experience with scoop, so I'll take your advice.

PR #847, opened 2020-02-28
vim-nightly: Advertise install-context script

idbrii:
Inform users how to add vim to the context menu.


bucket/vim-nightly.json
```diff
@@ -3,6 +3,7 @@
     "description": "A highly configurable text editor for efficiently creating and changing any kind of text.",
     "license": "Vim",
     "version": "8.2.0324",
+    "notes": "Add gVim as a context menu option by running: $dir\\install-context.reg",
```
Ash258:
```suggestion
    "notes": "Add gVim as a context menu option by running: '$dir\\install-context.reg'",
```

idbrii: I didn't add quotes because it makes it harder to copy paste -- double clicking selects the quotes.
Is that acceptable or do you still want me to change?

Ash258: Not all terminals linkify paths without quotes

idbrii: Double quotes still copy-paste correctly, so I used them. (Single quotes produce an error when copy pasted in cmd.exe).

idbrii:
Also added the same note to the vim json.

idbrii:
To address the "1 change requested" that I can't seem to mark as resolved:

> I do not think this is needed as the file is print while downloading, but if you cannot find it, keep it then.

I use scoop to install tons of packages at once, so I often miss some of the output pointing to these files. By putting this in notes, I can find the information when I need it -- when I later notice that the context menu isn't setup.

PR #878, opened 2020-03-13
coursier: Set COURSIER_HOME

nightscape:


Ash258:
- Why?
- Why to localappdata?

nightscape:
Why: Mostly because of this SBT bug https://github.com/scala/scala-dev/issues/666 and because many other packages do the same thing.
Why to LOCALAPPDATA: That's where Coursier puts its files unless told otherwise.
I was also considering setting it to the corresponding `persist` folder but wasn't sure about it because it could lead to people having two Coursier caches in different locations if they e.g. first used SBT without explicitly installing Coursier via scoop (SBT bundles Coursier) and later installed Coursier in scoop.

PR #909, opened 2020-03-21
pwsh: Make shortcut start in home folder

Livven:
Currently the start menu shortcut for PowerShell Core uses the install folder as its working directory, i.e. `~/scoop/apps/pwsh/current`. This change makes it start in the user home folder instead, as you would normally expect.

### DIFF
--- a/bucket/pwsh.json
+++ b/bucket/pwsh.json
@@ -17,7 +17,8 @@
     "shortcuts": [
         [
             "pwsh.exe",
-            "PowerShell Core"
+            "PowerShell Core",
+            "-WorkingDirectory ~"
         ]
     ],
     "checkver": "github",


Livven:
@Ash258 Any comment? Seems like this would be a minor but obviously useful improvement.

Ash258:
Personal preference configuration should not affect official bucket.

Livven:
This is hardly personal preference... start menu shortcuts for pretty much all shells/terminals will normally launch in the home directory and not in the folder where their executable is located.

The default shortcuts for Windows PowerShell and Command Prompt do this, `git-bash` does this [via its manifest](https://github.com/ScoopInstaller/Main/blob/43e9b40213abcabaf24da81f26c3e1022a58c70d/bucket/git.json#L27), others like ConEmu do it by default so no command line arguments are necessary.

The start menu shortcut created by the [PowerShell 7 MSI installer](https://github.com/PowerShell/PowerShell/releases/tag/v7.0.0) also uses the exact same argument as in this PR to start in the home folder.

Livven:
@Ash258 Any chance this could get merged? Again, this is not personal preference, but literally the exact same thing the _official_ PowerShell installer does.

rashil2000:
No worries @Livven, this was indeed an important fix (and certainly not a personal preference).

PR #910, opened 2020-03-21
git: Add bin/bash.exe to path

Livven:
Unlike the already existing git-bash.exe, this allows you to properly launch bash in a console without opening a new terminal window.

I think this might be a somewhat opinionated change, even though I personally find it very useful. Maybe this could alternatively be a separate preset `git-with-bash` instead?

### DIFF
--- a/bucket/git-with-openssh.json
+++ b/bucket/git-with-openssh.json
@@ -17,6 +17,7 @@
         "cmd\\git.exe",
         "cmd\\gitk.exe",
         "cmd\\git-gui.exe",
+        "bin\\bash.exe",
         "git-bash.exe",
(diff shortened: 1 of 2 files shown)


ddavness:
**Careful!** Windows uses `bash` to launch WSL.
```
$ scoop which bash
C:\Windows\system32\bash.exe
```

Livven:
Good point, would override that. Definitely not a good idea for a default then.

Alternatively then I'd prefer `git-bash` to point to `bash` so that it does not open a new terminal window but instead opens bash in an existing one. That would stray from what the standard Git installer does though so probably not appropriate here as well.

I'll simply close this then since no one else seems to feel strongly about this.

PR #916, opened 2020-03-22
feat(github): Add GitHub Actions

Ash258:


niheaven:
Does this work fine now? Just as the one in extras bucket?

Ash258:
Yes. Its configuration is same as in Extras.

- Issues
	- Automatically on issue creation
	- labeled with 'verify' label.

PR #921, opened 2020-03-24
python: Add user base scripts to path

### LINKED ISSUE #772: [Request] Handle Python user base scripts directory
The Scoop manifest for Python ([python.json](https://github.com/ScoopInstaller/Main/blob/3a7112ebdef3ed0ab4725926a75cf72fe143558c/bucket/python.json)) does not handle the Python [user base](https://docs.python.org/3/library/site.html#site.USER_BASE) scripts directory, so CLI scripts from packages installed in user mode (`pip install --user`) are not available on the PATH. Performing [user installations](https://pip.pypa.io/en/stable/user_guide/#user-installs) of packages is considered best practice, and many packages (such as the popular [Pipenv](https://pipenv.readthedocs.io/en/latest/install/#pragmatic-installation-of-pipenv)) recommend doing them.

Two possible solutions might be:

1. Construct the user base scripts directory and add it to PATH. The path can be constructed as follows: `Join-Path -Path (Split-Path -Path (python -m site --user-site) -Parent) -ChildPath Scripts`
1. Point the [PYTHONUSERBASE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUSERBASE) environment variable to a directory within the Python install directory and persist it.

Related issue: ScoopInstaller/Versions#196

bryanburke:
This PR modifies the Python app manifest to add the Python user base scripts directory to PATH. Now, when installing Python packages with `pip install --user`, any CLI scripts are immediately accessible. For example:

```powershell
PS C:\Users\burke> pip install --user virtualenv
Collecting virtualenv
...
Successfully installed ... virtualenv-20.0.13
PS C:\Users\burke> pip list -v
Package    Version Location                                                     Installer
---------- ------- ------------------------------------------------------------ ---------
...
virtualenv 20.0.13 c:\users\burke\appdata\roaming\python\python38\site-packages pip
PS C:\Users\burke> scoop which virtualenv
C:\Users\burke\AppData\Roaming\Python\Python38\Scripts\virtualenv.exe
```

Additionally, an added benefit of this change is that any Python packages installed this way persist between updates of the same major Python version.

### DIFF
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -43,6 +43,7 @@
         "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
         "}"
     ],
+    "post_install": "add_first_in_path (Join-Path -Path (Split-Path -Path (& $dir\\python.exe -m site --user-site) -Parent) -ChildPath Scripts) $global",
     "installer": {
         "script": [
             "Expand-DarkArchive \"$dir\\$fname\" \"$dir\\_tmp\"",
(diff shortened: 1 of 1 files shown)


bryanburke:
Closing as never reviewed.

PR #932, opened 2020-03-27
Fix .editorconfig

ycrack:
This looks like a mistake

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,7 +12,7 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -12,7 +12,7 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{bat,cmd,[Bb][Aa][Tt],[Cc][Mm][Dd]]
+[*.{bat,cmd,[Bb][Aa][Tt],[Cc][Mm][Dd]}]
```
Ash258:
```suggestion
[*.{[Bb][Aa][Tt],[Cc][Mm][Dd]}]
```

ycrack: That's right 👍

PR #935, opened 2020-03-27
dvc: Update to version 0.91.1

Ash258:


niheaven:
@Ash258 Could you update these outdated github based manifests? I'll review and merge them.

Ash258:
I should now have admin so I will do some casual auto-pr later today. I am thinking of some solution to this. Because this should be avoidable with multiple checkver calls in auto-pr.

Instead of one `bin\checkver.ps1 *` we could do a batches. First get list of all manifests, and iterate in some simple manner.

checkver '[0-9]*'
checkver '[0-9]*'
checkver '[a-c]*'

But I need to test how many requests is too much for github. This change was over night without any announcement and I do not know why it was changed. Or maybe we can contact github directly if they can whitelist specific useragent.

PR #946, opened 2020-03-28
Added Post Install Script to 7zip json to symbolic link

skyhoshi:
This Symbolic Link is to the known default location of 7zip that some scripts and applications search for when testing to see if 7zip is installed. 

I've included 32bit and 64 bit but was only able to fully test 64 bit. 

this also deals with a condition where if you already have 7zip installed it doesn't update the Default install location. in other words if you install 7zip via Global and then attempt to install it again via scoop it will not update the default's location symbolic link. as only one instance is needed. 

if you install 7zip from the MSI Manually and don't change the default location the application will update the existing folder if it exists. which would alter the scoop's version but would be update by scoop when 7zip is updated via scoop. 

that's all the usecases I can think to test for.

### DIFF
--- a/bucket/winpython.json
+++ b/bucket/winpython.json
@@ -1,13 +1,13 @@
 {
-    "version": "3.8.1.0",
+    "version": "3.8.2.0",
     "description": "Free, open-source and portable Python distribution for Windows",
     "homepage": "https://winpython.github.io",
     "license": "MIT",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


skyhoshi:
found an issue with the script going to put this on hold for a min

skyhoshi:
Updated and fixed a testing issue. would return null when directory existed but was not a symbolic link.

PR #952, opened 2020-03-31
php: Add base dir to path

viewable:
PHP install includes cURL support by default (via the included `ext/php_curl.dll` module) and it's dependency OpenSSL dlls (`./libcrypto-*.dll` and `./libssl-*.dll`). 

But because the PHP install base dir is not present in PATH, the OpenSSL dlls may not be found (unless installed elsewhere and are accessible via PATH), and the cURL module will throw errors when trying to load any time `php` is run.


r15ch13:
Which errors are cURL showing?
Installed `php cacert extras/vcredist`, enabled `curl` extension and added `cacert.pem` to `php.ini`. I can't reproduce it.

image

```php
<?php
error_reporting(E_ALL);
$ch = curl_init();

curl_setopt($ch, CURLOPT_URL,"https://www.apache.org/");
curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
$server_output = curl_exec($ch);

var_dump(curl_errno($ch));
var_dump(curl_error($ch));
curl_close($ch);
echo substr($server_output, 0, 1000);
```

viewable:
```
PS C:\> php -m
PHP Warning:  PHP Startup: Unable to load dynamic library 'curl' (tried: ext\curl (The specified module could not be found.), ext\php_curl.dll (The specified module could not be found.)) in Unknown on line 0
[PHP Modules]
bcmath
calendar
Core
ctype
date
dom
filter
hash
iconv
json
libxml
mysqlnd
pcre
PDO
Phar
readline
Reflection
session
SimpleXML
SPL
standard
tokenizer
xml
xmlreader
xmlwriter
zip
zlib
```

Specifically `PHP Warning:  PHP Startup: Unable to load dynamic library 'curl' (tried: ext\curl (The specified module could not be found.), ext\php_curl.dll (The specified module could not be found.)) in Unknown on line 0` and that the curl module doesn't load.

It does exist at `ext\php_curl.dll`, and it is enabled and pathed correctly in `php.ini`, but the module will not load unless certain SSL libs are available on PATH.

Seems to be a common issue w/ PHP's curl module on Windows. Some SSL libs have to be accessible by PATH for it to work.

Specific note from the PHP installation docs for the curl module (https://www.php.net/manual/en/curl.installation.php):

>    Note: Note to Win32 Users
>    In order to enable this module on a Windows environment, libeay32.dll and ssleay32.dll, or, as of OpenSSL 1.1 libcrypto-*.dll and libssl-*.dll, must be present in your PATH. Also libssh2.dll must be present in your PATH. You don't need libcurl.dll from the cURL site.

rashil2000:
The error is no longer present in the latest release of php

image

PR #971, opened 2020-04-02
Add "py" alias shim for launch python from cmd.

leni8ec:
It's a useful cmd for launch python!

### DIFF
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -73,6 +73,10 @@
             "python.exe",
             "python3"
         ],
+        [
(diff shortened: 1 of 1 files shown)


rasa:
I’m not sure this will work, as the py command is designed to launch different versions of python: For example: `py -2` runs python 2, and `py -3` runs python3. If you hardcode py to run a specific version, this functionality is lost.

niheaven:
Use `python` or `python2` for scoop-installed python, and `py` is just a wrapper.

PR #978, opened 2020-04-04
megacmd: Fix background server and persist settings

Dabombber:
The background server auto-start location is hardcoded to `%LOCALAPPDATA%\MEGAcmd`, so to get it working it's linked to the current install.

The `.megaCmd` folder is where settings and download info is stored so that should be persisted. There is a `megacmd.version` file which is generated there, but I think that's only used by the updater so it should be safe to ignore.

On first run a task is scheduled for updates, and updates are set to enabled so the cfg file is written to stop those.

The batch files are just wrappers for `MEGAclient.exe`, so it'd be better to just link them directly. `mega-cmd`, `mega-cmd-server` and `mega-exec` added for parity with linux versions. `mega-exit` is an alias for `mega-quit` which seems to have been forgotten.


Ash258 (dismissed):
- SymbolicLink requires admin privileges
- Revert the binaries changes. Use bats

@ScoopInstaller/maintainers Someone please review this. I really do not want to custom symbolic links and junctions in official buckets

Dabombber:
Might need to take a look at git-filter-repo then, although the links in that are all inside `$dir`.

issaclin32 (approved):

PR #1019, opened 2020-04-23
mill: Support auto-update for in-between versions.

heksesang:
The current version does not support auto-updating to "in-between" versions like 0.6.2-7-f10b89. This change allows such versions to be installed as well.

### DIFF
--- a/bucket/mill.json
+++ b/bucket/mill.json
@@ -10,7 +10,7 @@
         "github": "https://github.com/lihaoyi/mill"
     },
     "autoupdate": {
-        "url": "https://github.com/lihaoyi/mill/releases/download/$version/$version-assembly#/mill.bat"
+        "url": "https://github.com/lihaoyi/mill/releases/download/$matchHead/$version-assembly#/mill.bat"
     },
     "suggest": {
         "JDK": [


niheaven:
`checkver` couldn't find the version number `0.6.2-11-f97b9c` under tag `0.6.2`.

heksesang:
This runs fine for me locally with the change:
```
> scoop install mill-nightly@0.6.2-11-f97b9c
WARN  Given version (0.6.2-11-f97b9c) does not match manifest (0.6.2)
WARN  Attempting to generate manifest for 'mill-nightly' (0.6.2-11-f97b9c)
Autoupdating mill-nightly
Downloading 0.6.2-11-f97b9c-assembly to compute hashes!
0.6.2-11-f97b9c-assembly (40,1 MB) [==========================================================================] 100%
Computed hash: d56a71d91f0c3651b749c7190388e499e3a87af6901e25064bd5fdc1db332ca5
Writing updated mill-nightly manifest
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
Installing 'mill-nightly' (0.6.2-11-f97b9c) [64bit]
Loading 0.6.2-11-f97b9c-assembly from cache.
Checking hash of 0.6.2-11-f97b9c-assembly ... ok.
Linking ~\scoop\apps\mill-nightly\current => ~\scoop\apps\mill-nightly\0.6.2-11-f97b9c
Creating shim for 'mill'.
'mill-nightly' (0.6.2-11-f97b9c) was installed successfully!
'mill-nightly' suggests installing 'java/oraclejdk' or 'java/openjdk'.
```

Where does the error you talk about come from?

Ash258:
This change could go in, but it will not work for Excavator.
There is no way how to handle this in checker the 100% way.

Assets sort in /releases differ from /releases/latest

So excavator would be updating to older releases.

image

heksesang:
Not sure what Excavator is, but I assume it's what auto-updates the buckets in this repo. So if we take this change as it is, that will keep just updating the repo with the official version (0.6.1, 0.6.2, 0.6.3 etc), but it's still possible to install 0.6.2-14-029675 locally. Isn't that the best approaching, keeping just stable versions in the manifest in the repo and locally you can install whatever version you'd want?

Issue #1027, opened 2020-04-26
cmder/cmder-full: the folder "bin" should be persist


vuwjzl17985:
The folder "bin" is designed to be injected into the PATH environment variable at runtime. Thus it should be included in "persist".

PR #1028, opened 2020-04-26
cmder: Persist "bin"

vuwjzl17985:
Move the folder "bin" into persist.
Fix https://github.com/ScoopInstaller/Main/issues/1027.

### DIFF
--- a/bucket/cmder-full.json
+++ b/bucket/cmder-full.json
@@ -18,6 +18,7 @@
         ]
     ],
     "persist": [
+        "bin",
         "config",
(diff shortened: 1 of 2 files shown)


Ash258:
How is persist related to PATH?

vuwjzl17985:
> 
> 
> How is persist related to PATH?

The executable files in *%cmder_root%\bin* would be injected into PATH only when *cmder* starts. In another words, these executable files are invisible to other programs.

Reference: [cmder hompeage](https://github.com/cmderdev/cmder/blob/master/README.md#single-user-portable-config).

PR #1031, opened 2020-04-26
gcc: Update to version 9.3.0-2

niheaven:


rasa (approved):
Tested and working locally. Later versions on http://repo.msys2.org/mingw/x86_64/ are .zst files, such as http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-gcc-10.2.0-4-any.pkg.tar.zst

PR #1056, opened 2020-05-05
minio: Use archive URL

janicmikes:
I updated the url accordingly to the min.io servers.

I also updated minio-client to the latest version while I was on it.

Cheers


janicmikes:
The minio-client was updated using the checkver.ps1 script to ensure the autoupdate works as intended.

janicmikes:
The installer failed with a 404 if the current version in the manifest is not the most recent.

if you use the /archive/ path then the previous releases still work as well as the most up to date one.

PR #1062, opened 2020-05-07
powercli: Add version 12.0.0-15947286

sinloss:
Introduce the `VMware PowerCLI` to this `main` bucket.


Ash258:
That is not how scoop powershell modules works.

https://github.com/lukesampson/scoop/wiki/PowerShell-Modules

> The name property is required if you use psmodule, and it should match the name of the .psd1 manifest for the PowerShell module for PowerShell to consider it "well-formed" and automatically detect the module (see here for more.)

It would work only if you use something like this:

```json
...
    "url": "https://vdc-download.vmware.com/vmwb-repository/dcr-public/2802fdb1-bca3-4875-b76f-901cb15c2b2a/509b4a98-5ee5-4048-9407-28b9f85ba65c/VMware-PowerCLI-12.0.0-15947286.zip",
    "hash": "md5:ad7b7116e65d277c129e5e9d92fceeb4",
    "extract_dir": "VMware.PowerCLI",
    "psmodule": {
        "name": "VMWare.PowerCLI"
    },
...
```

But then it will not work as it require other modules to be loaded.

There is no reason to prefer scoop module over native PowerShell module, where you can install any specific version.

sinloss:
It has the same structure as the [azure-ps](https://github.com/ScoopInstaller/Main/blob/master/bucket/azure-ps.json). And it does the same type of things as the `azure-ps` I think.

PR #1073, opened 2020-05-10
legit: Add version 1.2.0

TheRandomLabs:
[Legit](https://frostming.github.io/legit/) is a complementary command-line interface for Git optimized for workflow simplicity.

### DIFF
--- a/bucket/legit.json
+++ b/bucket/legit.json
@@ -20,7 +20,7 @@
     "autoupdate": {
         "url": "https://github.com/frostming/legit/releases/download/$version/legit.exe",
         "hash": {
-            "url": "$baseurl/legit.exe.sha256"
+            "url": "$url.sha256"
         }
     }
 }
--- a/bucket/legit.json
+++ b/bucket/legit.json
@@ -1,0 +1,26 @@
+{
+    "version": "1.2.0",
+    "description": "Complementary command-line interface for Git.",
+    "homepage": "https://frostming.github.io/legit/",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://github.com/frostming/legit/blob/master/LICENSE"
+    },
+    "depends": "git",
+    "url": "https://github.com/frostming/legit/releases/download/1.2.0/legit.exe",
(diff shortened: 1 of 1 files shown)


bucket/legit.json
```diff
@@ -0,0 +1,26 @@
...
+    "description": "Complementary command-line interface for Git.",
+    "homepage": "https://frostming.github.io/legit/",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://github.com/frostming/legit/blob/master/LICENSE"
+    },
+    "depends": "git",
+    "url": "https://github.com/frostming/legit/releases/download/1.2.0/legit.exe",
+    "hash": "2c4aa8a777141cad9171a7aec9b91d6570443644f7ca040903773620074d361b",
+    "post_install": "'y' | legit --install | Out-Null",
```
r15ch13, line 12:
```suggestion
    "installer": {
        "script": "'y' | legit --install | Out-Null"
    },
```

bucket/legit.json
```diff
@@ -0,0 +1,26 @@
...
+        "script": "legit --uninstall | Out-Null"
+    },
+    "bin": "legit.exe",
+    "checkver": {
+        "github": "https://github.com/frostming/legit"
+    },
+    "autoupdate": {
+        "url": "https://github.com/frostming/legit/releases/download/$version/legit.exe",
+        "hash": {
+            "url": "$baseurl/legit.exe.sha256"
```
r15ch13, line 23:
```suggestion
            "url": "$baseurl/$basename.sha256"
```

bucket/legit.json
r15ch13: Hm, this doesn't seem to work:
image

TheRandomLabs: Interesting, it works for me:
image
Is there a better way to pass in that input?

r15ch13: Do you use bash or pwsh/powershell? :smile:

TheRandomLabs: PowerShell. :)

r15ch13: Okay, it works fine in Windows Sandbox. (PS 5 and 7)

image

PR #1098, opened 2020-05-17
ffsend: Update to version 0.2.62

timvisee:
The `openssl` dependency can now be removed, so here is a manual PR.


timvisee:
Superseded by #1106

PR #1107, opened 2020-05-20
fix: WixToolPath not added to env Variables

iamsilvio:
environment variable for WixToolPath added


h404bi:
Is there any reference related to this env?

iamsilvio:
Sorry for the lag of context:
If you install the wixtoolset to a custom location almost all build environment require you to set the WixToolPath to this location, otherwise you have to script around it.
It's not mentioned in the WixToolset Documentation but there is also no documentation for Installation at a custom location!

PR #1120, opened 2020-05-25
mercurial: Update to version 5.4

tukanos:
… on Windows.  Python2.7 support will be probably dropped during 2020.

### DIFF
--- a/bucket/mercurial.json
+++ b/bucket/mercurial.json
@@ -17,7 +17,7 @@
     "bin": "hg.exe",
     "checkver": {
         "url": "https://www.mercurial-scm.org/release/windows/latest.dat",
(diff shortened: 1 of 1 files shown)


rasa:
Superceded by https://github.com/ScoopInstaller/Main/commit/1c885e4803500a85f262b9ba6c2306b3d585277a

tukanos:
> Superceded by [1c885e4](https://github.com/ScoopInstaller/Main/commit/1c885e4803500a85f262b9ba6c2306b3d585277a)

The issue here is that your PR uses Python 3 on Windows that is still experimental and the developers are still recommending using Python 2.7 till they drop the support.  Python 3 should be 100% supported on Windows.

PR #995, opened 2020-04-15
kubenswin: Add version 0.1.1

tynril:
[kubenswin](https://github.com/thomasliddledba/kubenswin) is a Windows port of `kubens`, which works alongside [kubectxwin](https://github.com/thomasliddledba/kubectxwin), which [is already in the main bucket](https://github.com/ScoopInstaller/Main/blob/master/bucket/kubectxwin.json).

I am creating this PR now that [kubenswin/#1](https://github.com/thomasliddledba/kubenswin/issues/1) has been closed. :)

PR #1127, opened 2020-05-26
Added kubenswin

smakhtin:
Added kubenswin (kubens) app — same author as kubectxwin, but allows people to change namespaces instead of Kubernetes contexts. I copied manifest from kubectxwin, works on my machine (installed from the local bucket).


Ash258:
Duplicate of #995

smakhtin:
> Duplicate of #995

Which you strangely don't want to merge, judging by the number of code lines.

PR #1137, opened 2020-06-02
shadowsocks-rust: Add version 1.8.11

linsui:


Ash258:
Move this PR into versions. There is no reason to keep both in extras .

linsui:
This is not in extras. It's a cli application and it icontains a server and a client while shadowsocks-windows is noly a gui client.

Ash258:
I see. Sorry my bad.

The `-rust` suffix is needed?

linsui:
There are some other implementations of the shadowsocks protocol with some different features. Without `-rust` users may feel confused. Also the name `shadowsocks`  is not available.

PR #1147, opened 2020-06-04
vswhere: Add version 2.8.4

heaths:
vswhere is a tool commonly with over 6M tracked downloads used in both on-prem and cloud pipelines like Azure DevOps, Visual Studio-related GitHub Actions, boost build pipelines, and more. It is used to locate a suitable Visual Studio 2017 or newer install with required features, and find toolsets within it or set up PATH. It's already in chocolatey but some users have asked for scoop support.

Fixes microsoft/vswhere#122

### DIFF
--- a/bucket/vswhere.json
+++ b/bucket/vswhere.json
@@ -1,0 +1,18 @@
+{
+    "version": "2.8.4",
+    "description": "Locate Visual Studio 2017 and newer installations",
+    "homepage": "https://github.com/microsoft/vswhere",
+    "license": "MIT",
+    "url": "https://www.nuget.org/api/v2/package/vswhere/2.8.4#/vswhere.zip",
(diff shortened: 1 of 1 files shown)


chawyehsu:
It's already in the [extras](https://github.com/lukesampson/scoop-extras/blob/master/bucket/vswhere.json) bucket, though I think it should be in the main bucket.

heaths:
This also has the autoupdate authoring (I'm the writer/maintainer as well), but at 6M+ installs and as much as its used in scripts, I would argue minimal setup required to install it would be beneficial to devs looking to use it in build automation.

rasa:
Is there a consensus on moving to the main bucket?

heaths:
Anything more required or to discuss to move this along?

PR #1154, opened 2020-06-07
uwpdumper: Add version 0.1.50

GalaxySnail:


Ash258:
There is no release. I am against adding this using appveyor artifacts.

GalaxySnail:
> There is no release. I am against adding this using appveyor artifacts.

OK, I see.

PR #1155, opened 2020-06-07
vim: Best effort to rename '$' files accurately

xfournet:
When installing vim with scoop the vim app directory doesn't fit as much as possible the standard vim installation.
Among of that the gvimext.dll is not available since it is still named $R0 after the scoop installation. This prevent to use the complete shell integration. I noticed the install-context.reg but it only cover the simple "open with" case, not all the features provided by Vim shell integration such as diff, open multiple file, etc...
This patch permit to have an almost identical vim app directory than with the standard installation, with the exception of the 32-bit shell integration : it would be tricky to do it since the 32 bit gvimext.dll is overridden by another file with the same name (i.e. $0\GvimExt32\$R0 is present twice in the archive, unfortunately the one that remain after the extraction is not the one corresponding to gvimext.dll). So i completely erased the GvimExt32 that is not usable.
However at least the 64-bit integration is ok with the patch.


xfournet:
Look likes there are problems with appveyor that are not related to the fix ?

PR #1211, opened 2020-06-25
pstools: Add pstools

Summon528:
Pstools is a set of windows process management tools from Microsoft.


Summon528:
I fixed the first four issues. Unfortunately, the version problem is a bit tricky as every tool has its own different version but Microsoft only provides the tools as a pack and the pack does not actually have a "version". I chose the version "20160629" because the MS post was published at that date, however that date might not be updated if they were to update the software pack.

Ash258:
You have version. 2016.July.4

image

Do not use proprietary license without providing URL. it gives anything to user and only create hustle for them.

https://docs.microsoft.com/en-us/sysinternals/license-terms

Summon528:
fixed

rasa:
We appreciate the contribution, but I am going to reject, as:

1. We have removed du and whois from main: https://github.com/ScoopInstaller/Main/search?q=sysinternals&type=commits
2. All of the sysinternal executables are available via [extras/sysinternals.json](https://github.com/lukesampson/scoop-extras/blob/master/bucket/sysinternals.json):
```
scoop bucket add extras
scoop install sysinternals
```
3. Any individual tool, can be installed via:
```
scoop bucket add sysinternals https://github.com/Ash258/Scoop-Sysinternals
scoop install appname
```
4. Pstools hasn't been updated in over four years.

NatoBoram:
>   3. Any individual tool, can be installed via:
> 
> ```
> scoop bucket add sysinternals https://github.com/Ash258/Scoop-Sysinternals
> scoop install appname
> ```

This is false, it has been migrated to a fork of Scoop (Shovel) and is no longer compatible with Scoop.

> Checking repo... OK
> The sysinternals bucket was added successfully.
> Couldn't find manifest for 'psexec'.

rasa:
https://github.com/mogeko/scoop-sysinternals looks like it has all the utilities, but it hasn't been updated since June 2021.

niheaven:
I've add a Sysinternals bucket, please use it:

https://github.com/niheaven/scoop-sysinternals

PR #1216, opened 2020-06-27
elm: Change depends to nodejs-lts

f-akmal:
To fix this annoying message:

Missing runtime dependencies:
    'elm' requires 'nodejs'


linsui:
We should use `suggest` instead. Elm do not need nodejs to work. Only the REPL relies on node.js.

rasa (approved):

rasa:
Installed fine locally

PR #1228, opened 2020-06-30
Add multidumper

mrpapersonic:


Ash258:
I am strongly against adding application using some random commit and published into random repository not associated with original author.

PR #1240, opened 2020-07-04
msys2: Update to version 20200629

codgician:
This PR updates msys2 to version 20200629.
~~This PR updates msys2 to version 20200517, which is the last version that comes with both `i686` and `x86_64` packages.~~

### DIFF
--- a/bucket/msys2.json
+++ b/bucket/msys2.json
@@ -2,25 +2,15 @@
     "homepage": "http://msys2.github.io",
     "description": "A software distro and building platform for Windows.",
     "##": "64-bit version is able to build both 32-bit and 64-bit packages",
-    "version": "20200517",
+    "version": "2020-06-29",
     "license": "GPL-2.0-only|BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


rasa:
> This PR updates msys2 to version 20200517, which is the last version that comes with both `i686` and `x86_64` packages.

Then let’s move this app to versions, and remove the 32bit references.

codgician:
> > This PR updates msys2 to version 20200517, which is the last version that comes with both `i686` and `x86_64` packages.
> 
> Then let’s copy this app to versions, and remove the 32bit references in this bucket.

How should it be named in versions? Like `msys2-20200517` or `msys2-i686`?

codgician:
The last commit removed 32bit references and bumped version to `2020-06-29`.

rasa:
> How should it be named in versions? Like `msys2-20200517` or `msys2-i686`?

msys2-20200517 would be consistent with the other apps in versions.

PR #1256, opened 2020-07-08
gitui: Add version 0.8.1

TheRandomLabs:


slaughtering:
Quite desirable though TheRandomLabs' bucket is already added on my host.

PR #1283, opened 2020-07-24
graphviz: Update to version 2.44.1

AntonOks:


Ash258:
Why do you prefer cmake over msbuild??

https://www2.graphviz.org/Packages/stable/windows/10/msbuild/Release/Win32/

Msbuild have much more executables

image

AntonOks:
> Why do you prefer cmake over msbuild??
> 
> https://www2.graphviz.org/Packages/stable/windows/10/msbuild/Release/Win32/
> 
> Msbuild have much more executables
> 
> image

because MSbuild is only 32bit... and cmake has both, 32 and 64bit...
also if the graphviz dev's offer differen binaries within those packages, it is for a reason, right? maybe the additional MSbuild 32bin binaries are deprecated?!

Ash258:
Then investigate why it is this way.

AntonOks:
> 
> 
> Then investigate why it is this way.

sorry, will not. If you don't want this update, I'm fine with your decision.

AntonOks:
> 
> 
> > Then investigate why it is this way.
> sorry, will not. If you don't want this update, I'm fine with your decision.

Btw., to be more precise... there is no reason to discuss it with the dev's... their explanation here makes it clear... MSbuild is the "graphviz debug build" and not "recommended for the most users":
https://forum.graphviz.org/t/new-simplified-installation-procedure-on-windows/224#cmake-or-msbuild

rasa (approved):

PR #1313, opened 2020-08-05
ensubst: Add version 1.2.0

stevehipwell:
This PR adds the [envsubst](https://github.com/a8m/envsubst) package.

From what I can from the project this package is x64 and I've added it accordingly.


stevehipwell:
@rasa, I've updated this could you please `/verify` it again?

PR #1328, opened 2020-08-13
Add conemu app manifest

ketan:
Conemu is a popular terminal replacement for windows. I realized that this was not available on scoop. So adding it here.


chawyehsu:
It's in the extras bucket: https://github.com/lukesampson/scoop-extras/blob/master/bucket/conemu.json

PR #1394, opened 2020-09-08
imagemagick@7.0.10-29: Add missing bin

jannick0:


niheaven:
Could you please remove `"depends": "ffmpeg",` and test the manifest again?

jannick0:
Thanks. Done in force push.

@niheaven However, I was fiddling around locally a bit. So, what is the best way to test a manifest when sitting on a bucket branch other than `master`? Is there any test ps1 to simply run? Info or link would be appreciated.

rasa:
I would argue against this change, per the discussion in https://github.com/lukesampson/scoop/pull/1192 . Feedback?

jannick0:
Plugging in the latest ffmpeg version into imagemagick, i.e. using a dependency to ffmpeg, might be reasonable, whereas 
* currently the ffmpeg version lag is not that terrible (4.2 from 2019 shipped with imagemagick vs. 4.3.1 from 2020 in its own package);   
* compatibility issues between imagemagick and the latest ffmpeg might arise.

Out of curiosity: Would scoop be able to handle two different versions of ffmpeg.exe at all - one from imagemagick and one from its own package?  Or is the dependency trick a way to work around this issue?

rasa:
> Would scoop be able to handle two different versions of ffmpeg.exe at all - one from imagemagick and one from its own package? Or is the dependency trick a way to work around this issue?

The ffmpeg.exe shim could only point to one exe at a time, so you could merge this PR, run `scoop install ffmpeg`, and then switch the "active" ffmpeg to imagemagick's via:
`scoop reset imagemagick`
to switch back to the ffmpeg package's version:
`scoop reset ffmpeg`

I think the best solution might be to add a `imagemagick-ffpeg` package, which is this PR's .json. This will allow users who want imagemagick's version of ffmpeg, to have it, while not changing the last three years of installations.

niheaven:
So rename this PR's manifest to `imagemagick-ffmpeg` and move it to https://github.com/ScoopInstaller/Versions sounds like a good idea, isn't it?

jannick0:
Umm, this might be beyond my solid scoop knowledge (which effectively is zero as an ordinary user), so I leave it to you guys to handle this issue and - if appropriate - close it now.

niheaven:
@jannick0 Use `scoop install imagemagick-ffmpeg` to install the ffmpeg one.

PR #835, opened 2020-02-24
aws: Update to version 2.0.23

xfournet:
AWS CLI v2 is now official

PR #1395, opened 2020-09-09
aws: Update to version 2.0.50

dennisroche:
- Closes #835  

> the AWS CLI V1 will continue to receive regular updates
that are published to PyPI and pip will continue to be a supported
installation mechanism for AWS CLI V1.
\- joguSD  https://github.com/aws/aws-cli/issues/4947#issuecomment-586046886

AWS CLI v1 will be maintained for sometime and there are valid reasons why a user will want to continue to use v1 and **not** migrate to v2, e.g. breaking changes for existing scripts, no SSO configured etc.

**Proposed:** Add AWS CLI v2 as separate install manifest. I understand that this might not a typical approach for `scoop` but I view aws cli v1 and aws cli v2 as separate tools.


dennisroche:
Renamed title to match `master` pattern of `Add {tool} {version}`.

Ash258:
#835

dennisroche:
This PR replaces #835. 

AWS CLI v1 and v2 are different products with v1 to continue receiving updates for some time.

rasa (approved):

Livven:
Isn't this what the [versions](https://github.com/ScoopInstaller/Versions) bucket is for? I mean, other applications have breaking changes too, e.g. [nodejs](https://github.com/ScoopInstaller/Main/commits/master/bucket/nodejs-lts.json), yet that only has one entry that is updated with new versions. So #835 should be merged instead.

Ultimately though I don't really care about which approach is chosen here as long as AWS CLI v2 finally becomes available via Scoop.

dennisroche:
> Isn't this what the [versions](https://github.com/ScoopInstaller/Versions) bucket is for? I mean, other applications have breaking changes too, e.g. [nodejs](https://github.com/ScoopInstaller/Main/commits/master/bucket/nodejs-lts.json), yet that only has one entry that is updated with new versions. So #835 should be merged instead.
> 
> Ultimately though I don't really care about which approach is chosen here as long as AWS CLI v2 finally becomes available via Scoop.

I wasn't aware of the [versions](https://github.com/ScoopInstaller/Versions) bucket and agree that looks like a better approach. 

If we all agree, let's:

- [ ] change this PR to upgrade `aws` to latest `2.x` (PR #835 appears stalled)
- [ ] add `aws1` to [versions](https://github.com/ScoopInstaller/Versions)

rasa (approved):

PR #1409, opened 2020-09-14
Add beeflang-nightly to scoop

mundusnine:
Tested autoupdate and installation.


mundusnine:
Just fixed issues with checkver regex to replace adequatly.

mundusnine:
@rasa  just compared both hashes and they are the same on my end... is there a way to see the output of the checks ? Can pushing after the pull request was created be an issue, were the test only takes the first commit ?

niheaven:
Please move this PR to https://github.com/ScoopInstaller/Versions/

PR #1410, opened 2020-09-14
Test beeflang-nightly

mundusnine:


Ash258:
Stop wasting actions running time and PRs. If you want test it test in your repo...


That prs is not working because it is not valid manifest as it has no json extension...

PR #1417, opened 2020-09-17
Update bind.json

tymes:
it already includes vcredist2017 (newer than scoop knows, but I've updated that), but it could simply just use latest vcredist2019 instead


Ash258:
9.17.5 is not stable release

image

tymes:
sry, thanx.

Issue #1415, opened 2020-09-16
ffmpeg@4.3.1: Download failed


sjmurdoch:
Scoop will not install ffmpeg 4.3.1 due to a 404 error. I verified using my normal web browser that indeed this results in a 404, and the same problem occurs with Tor Browser (so it is not my ISP blocking this file).

```PS > scoop update
Updating Scoop...
Updating 'extras' bucket...
Updating 'main' bucket...
Scoop was updated successfully!
PS > scoop install ffmpeg
Installing 'ffmpeg' (4.3.1) [64bit]
The remote server returned an error: (404) Not Found.
URL https://ffmpeg.zeranoe.com/builds/win64/static/ffmpeg-4.3.1-win64-static.zip is not valid
PS C:\Users\sjmur>
```

sjmurdoch:
e.g. download fom Tor
<img width="458" alt="Annotation 2020-09-16 110110" src="https://user-images.githubusercontent.com/283486/93323206-37f01b80-f80c-11ea-8630-db2949d17384.png">

and direct
<img width="584" alt="Annotation 2020-09-16 110242" src="https://user-images.githubusercontent.com/283486/93323236-44747400-f80c-11ea-9a44-ff300d575dfb.png">

sjmurdoch:
It could be that the download is blocked, but given that I see the same error from multiple network connections I think that it could be an intermittent error with the site and/or blocking performed by the site itself.

sjmurdoch:
Ah, I think it could be the website shutting down, see https://ffmpeg.zeranoe.com/builds/

> ffmpeg.zeranoe.com will close on Sep 18, 2020, and all builds will be removed. If you're using Zeranoe FFmpeg Builds in your product, please ensure that you have a hosting strategy in place by the 18th. Thank you to everyone who supported the project along the way!

arinoki:
For 64bit i think we can use this. https://github.com/AnimMouse/ffmpeg-stable-autobuild

chawyehsu:
![20200920133356](https://user-images.githubusercontent.com/5764917/93695136-f8943880-fb45-11ea-900d-2f3d1847e2f2.png)

I would suggest to use the one listed on FFmpeg website: https://github.com/BtbN/FFmpeg-Builds/releases

lando-cal:
I can't seem to find an official source for the shared version. If anyone knows where that's at, let me know and I'll attempt to update that one as well.

PR #1424, opened 2020-09-20
ffmpeg: Use alternative download URLs

lando-cal:
This addresses #1415 for 64-bit ffmpeg. It still needs 32 bit, but this got me through for now. Figured I'd contribute as a starting point for anyone else who wants to pick it up.

close #1415 
close #1406
close #1379
close https://github.com/lukesampson/scoop/issues/4105
close https://github.com/lukesampson/scoop/issues/4078

PR #1425, opened 2020-09-21
temporary ffmpeg.json patch.

KScottDB:
idk much about making scoop packages, but this will work until the next version of ffmpeg for sure.

if you don't understand *why* this patch is required, it's because zeranoe shut down recently, making the current ffmpeg package broken. this fixes issue #1415 .


KScottDB:
oops! #1424 already fixes this!

PR #1428, opened 2020-09-23
csview: Add version 0.3.5

wfxr:
Added [csview](https://github.com/wfxr/csview).


rasa:
This tool seems somewhat well established, so I think it can live in our main bucket. If anyone thinks otherwise, please chip in.

rasa (approved):

PR #1466, opened 2020-10-03
grep: add egrep and fgrep to shim

### LINKED ISSUE #1460: egrep and fgrep are not included in shims
Hi there,
I'm new to scoop. I find there are three exe for grep: grep.exe, egrep.exe, fgrep.exe (in the bin folder). However, only grep.exe is created in the shims folder.
I hope the other two exe are also created so that we can directly use egrep/fgrep in command line.
The installation json file is https://github.com/ScoopInstaller/Main/blob/master/bucket/grep.json.

Thanks.

wstein:
fixes #1460

### DIFF
--- a/bucket/grep.json
+++ b/bucket/grep.json
@@ -11,5 +11,9 @@
         "sha1:56f41d351b3ed8ac671df4dd3bbd4c4d3b9190a2",
         "sha1:6dc3a0d1a1751c731fb680a01650a1396c76648c"
     ],
-    "bin": "bin\\grep.exe"
+    "bin": [
+        "bin/grep.exe",
+        "bin/egrep.exe",
+        "bin/fgrep.exe"
+    ]
 }


rasa (approved):

tylerszabo:
It looks like #3082 reverted this fix

rashil2000:
Yes, because the updated source does not have those binaries

tylerszabo:
That's unfortunate; this update is a bit more destructive than I was expecting. Is there a package source that still contains `fgrep` and `egrep`?

rashil2000:
The gnuwin32 builds are still available, but they're like 15-16 year old

wstein:
I think it should be enough to to create the two shims egrep and fgrep using the grep executable.
grep automatically sets the necessary option dependent of the filename. At least this is the standard behavior on linux.

rashil2000:
I don't use grep so I'm not familiar, can you make a PR for the same?

tylerszabo:
Who's the maintainer for the package?

rashil2000:
Manifests don't have maintainers here

tylerszabo:
Gotcha.

If I understand the behavior correctly then perhaps the `grep.exe` binary could be copied to `egrep.exe` and `fgrep.exe` and then respective shims created but I'm not sure if scoop supports that behavior.

I can otherwise see the dilemma of wanting to use a more recent version build. Are there things the new build fixes?

rashil2000:
https://github.com/ScoopInstaller/Main/commit/7209d4ad1d98e3e5a9a4f35a873ff97f6281770d

tylerszabo:
I put a comment in the commit; adding args in the shim is a slightly different behavior (and one that can cause confusion). So it might be better to just note the deprecation of `egrep` and `fgrep` in this manifest for now. I do appreciate the attempts you're making to find a solution - I don't want to burn too much of your time hunting down this thing (since I can work around it on my end and the discussions recorded here make it easy to read about if anyone else hits the same issue).

If I come up with anything I'll hit you with a PR :)

rashil2000:
> adding args in the shim is a slightly different behavior

Can you indicate how so? I'll add a note in the manifest.

tylerszabo:
For the `grep` binary it should behave the same; but I it adds fragility to alter the behavior in a shim. I did some more reading and found that `egrep`, `fgrep`, and `rgrep` are deprecated (I just hadn't noticed the deprecation warning in the official documentation). So I think I'm the one that needs to make changes here and the change that took the new build without the new builds is the correct way to go.

PR #1467, opened 2020-10-04
cuda: fix autoupdate

jat001:


jat001:
I just noticed the directory structure of cuda is very complicated. this manifest does not work on cuda 11. we should deprecate it

PR #1482, opened 2020-10-07
Latest PyPy 3 release is not 3.6 but 3.7

cprieto:
I will expect PyPy3 to be 3.7 but it is installing latest 3.6 version.


cprieto:
Someone already applied the change directly

PR #1483, opened 2020-10-08
python: persist 'Scripts' folder

Vixb1122:
python: persist 'Scripts' folder to avoid re-downloading scripts after every update.


dwettstein:
Closed by #2144

Issue #1372, opened 2020-08-28
Update gdb to use newer releases


Lunchb0ne:
So, the package that gets downloaded for gdb is pretty old now, newer binaries have since been moved to GitHub releases at https://github.com/jmeubank/tdm-binutils-gdb

It would be nice to get that updated to reflect the changes.

Saggiehaim:
Hey, 
I created a PR for that :)

GitMensch:
_Originally posted by @rasa in https://github.com/ScoopInstaller/Main/issues/1496#issuecomment-713686092_
> Obsolete. If http://tdm-gcc.tdragon.net/ releases are needed, it should be added to the versions bucket.

The GDB version that is currently in the [gdb bucket](https://github.com/ScoopInstaller/Main/blob/master/bucket/gdb.json) is 9.1 which cannot be considered "pretty old" - so I think this issue can be closed.

GitMensch:
Note: GDB is used from the MSYS2 repo and _now_ that is at GDB 11.1 (with a new optional dependency to `python-pygments` which I'd suggest to include), so it actually would be nice to have this included (maybe even with gdb-multiarch?)

PR #1496, opened 2020-10-11
Update gdb to use newer releases #1372

Saggiehaim:
Hey,

As issue #1372 state,
I've updated the link to the newest version and updated the hash.

If possible, I will be happy if you add the "hacktoberfest-accepted" label to this PR.

Thanks!


rasa:
Obsolete. If http://tdm-gcc.tdragon.net/ releases are needed, it should be added to the versions bucket.

Issue #1364, opened 2020-08-25
puppet-bolt


dsisnero:
puppet-bolt installs but is not able to run

F:\programming\source\repos\rails\autocomplete_stimulus_example>bolt --help
Traceback (most recent call last):
        2: from F:/windows/scoop/apps/puppet-bolt/current/bin/bolt:23:in `<main>'
        1: from F:/windows/scoop/apps/puppet-bolt/2.24.1/lib/ruby/2.5.0/rubygems.rb:303:in `activate_bin_path'
F:/windows/scoop/apps/puppet-bolt/2.24.1/lib/ruby/2.5.0/rubygems.rb:284:in `find_spec_for_exe': can't find gem bolt (>= 0.a) with executable bolt (Gem::GemNotFoundException)

dsisnero:
I think it is because of ruby scoop manifest setting the GEM_HOME because of versions.  need to UNSET THIS

jfastnacht:
Related: https://github.com/puppetlabs/bolt/issues/2222

jfastnacht:
The missing files are now within the tarball at `share\install-tarballs\ruby-2.5.8.tgz`. The `2.5.8` is a varying version number.

According to the answers within the issue in the puppet bolt repository, the bolt.cmd is not the correct way of using Bolt. There is a PowerShell module which should be installed as well.

dsisnero:
The GEM_HOME command is still causing issues:

Updating 'puppet-bolt' (2.32.0 -> 2.33.1)
Downloading new version
Starting download with aria2 ...
Download: Download Results:
Download: gid   |stat|avg speed  |path/URI
Download: ======+====+===========+=======================================================
Download: 2dba1e|OK  |   3.6MiB/s|F:/windows/scoop/cache/puppet-bolt#2.33.1#https_downloads.puppetlabs.com_windows_puppet6_puppet-bolt-2.33.1-x64.msi
Download: Status Legend:
Download: (OK):download completed.
Checking hash of puppet-bolt-2.33.1-x64.msi ... ok.
Uninstalling 'puppet-bolt' (2.32.0)
Removing shim for 'bolt'.
Unlinking F:\windows\scoop\apps\puppet-bolt\current
Installing 'puppet-bolt' (2.33.1) [64bit]
Loading puppet-bolt-2.33.1-x64.msi from cache.
Extracting puppet-bolt-2.33.1-x64.msi ... done.
[...]

PR #1497, opened 2020-10-13
puppet-bolt: fix issue with missing gems

jfastnacht:
Add post-install script for tarball of gems to fix the issue https://github.com/ScoopInstaller/Main/issues/1364

The PowerShell module relies on a registry entry which can't be set without elevated privileges. Therefore it won't be part of this fix.

### DIFF
--- a/bucket/puppet-bolt.json
+++ b/bucket/puppet-bolt.json
@@ -9,6 +9,12 @@
             "hash": "82471b668eb9f8c43cf62e3a9184fdca903e2d657d78c0d702b15716cc5f6789"
         }
     },
+    "post_install": [
+        "if (Test-Path \"$dir\\share\\install-tarballs\\ruby*.tgz\") {",
+        "   Write-Host -F yellow \"Extract tarball of gems...\"",
+        "   Expand-7zipArchive \"$dir\\share\\install-tarballs\\ruby*.tgz\" \"$dir\"",
+        "}"
+    ],
     "extract_dir": "Puppet Labs\\Bolt",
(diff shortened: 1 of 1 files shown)


dsisnero:
please review this!

jfastnacht:
@dsisnero you can install it directly

```
scoop uninstall puppet-bolt
scoop install https://raw.githubusercontent.com/jfastnacht/Main/fix/puppet-bolt/bucket/puppet-bolt.json
```

until it is merged/fixed.

rasa (approved):
Installed fine locally.

dsisnero:
still not find the gems

C:\Users\Dominic E Sisneros>bolt
Traceback (most recent call last):
        2: from F:/windows/scoop/apps/puppet-bolt/current/bin/bolt:23:in `<main>'
        1: from F:/windows/scoop/apps/puppet-bolt/2.32.0/lib/ruby/2.5.0/rubygems.rb:303:in `activate_bin_path'
F:/windows/scoop/apps/puppet-bolt/2.32.0/lib/ruby/2.5.0/rubygems.rb:284:in `find_spec_for_exe': can't find gem bolt (>= 0.a) with executable bolt (Gem::GemNotFoundException)

jfastnacht:
I see, my fault. Should have read the original issue more thoroughly. The issue is indeed the `GEM_HOME` variable in `env_set` within the [ruby manifest](https://github.com/ScoopInstaller/Main/blob/master/bucket/ruby.json#L24).
This PR is merged though, so please continue within #1364 and I (or someone else) will open up a new PR when it is fixed.

PR #1500, opened 2020-10-14
LLVM: Update to version 11.0.0

pratikpc:
Remove clang-import-test shim because it is not distributed anymore

### DIFF
--- a/bucket/llvm.json
+++ b/bucket/llvm.json
@@ -1,16 +1,16 @@
 {
-    "version": "10.0.0",
+    "version": "11.0.0",
     "description": "Collection of modular and reusable compiler and toolchain technologies.",
     "homepage": "https://www.llvm.org",
     "license": "NCSA",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe#/dl.7z",
-            "hash": "893f8a12506f8ad29ca464d868fb432fdadd782786a10655b86575fc7fc1a562"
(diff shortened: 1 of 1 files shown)


chawyehsu:
Please also fix the checkver

pratikpc:
I would love to help out with the checkver @chawyehsu but I have no idea how to

pratikpc:
@chawyehsu the issue lies with Scoop's parsing of the webpage or maybe with LLVM's webpage

I presume it's down to them not specifying Content-Type in HEAD

Using checkver.ps1 file, and printing $page on the console I saw gibberish, unicode esque gibberish on the screen while the same doesn't occur for 7zip

To face the same issue, add `echo $page` to https://github.com/lukesampson/scoop/blob/731d878490a7bb1aa9d2bf1c15cd27ef2c31c2fb/bin/checkver.ps1#L257

batkiz:
Maybe we could use
```json
"checkver": {
    "url": "https://llvm.org/",
    "regex": "([\\d.]+):"
},
```

to fix the `checkver`, because every release would be listed in the *Release Emails* section on homepage.

I've tried the `checkver` code above and it actually works
```
> cat .\bucket\llvm.json
{
    "version": "10.0.0",
    ...
    "checkver": {
        "url": "https://llvm.org/",
        "regex": "([\\d.]+):"
    },
    ...
}

> .\bin\checkver.ps1  -dir .\bucket\ -app llvm
llvm: 11.0.0 (scoop version is 10.0.0) autoupdate available
```

chawyehsu (approved):

PR #1501, opened 2020-10-14
bit: Add version 0.5.14

JanPokorny:
Add [bit](https://github.com/chriswalz/bit) -- a git CLI with some goodies (incl. some commands from git-extras)


JanPokorny:
@chawyehsu It's just a version mismatch issue (new version came in the meantime), no reason to fix now since dude's pushing new versions like crazy, three in the last 24h, even if I updated it here, it would be out of date before you merge 😄

chawyehsu (approved):

PR #1504, opened 2020-10-15
udpate to pandoc 2.11.0.1

cderv:
This will update the manifest to new release of Pandoc

There is no more 32bits version BTW.

### DIFF
--- a/bucket/pandoc.json
+++ b/bucket/pandoc.json
@@ -1,19 +1,15 @@
 {
-    "version": "2.10.1",
+    "version": "2.11.0.1",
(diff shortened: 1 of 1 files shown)


chawyehsu (approved):

AntonOks:
hmm, my scoop update didn't work with this error message "**Can't shim 'pandoc-citeproc.exe': File doesn't exist.**" so I guess CI/tests shouldn't have passed...

see also #1507

cderv:
Oh thanks! Yes I forgot to remove the bin 🤦‍♂️

PR #1509, opened 2020-10-17
ccextractor: Add version 0.88

spongbros:


rasa:
As a GUI app, this really belongs in the scoop-extras bucket. See https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket
Please open a PR there as we can', move this one. Thanks for your contribution.

PR #1539, opened 2020-10-29
yosys: Add version 0.9

LukBukkit:
Hello,

[yosys](https://github.com/YosysHQ/yosys) is a tool for logic synthesis which can be used for the hardware description langauge language (System)Verilog.
Combined with [graphviz](https://github.com/ScoopInstaller/Main/blob/master/bucket/graphviz.json) you can use it as a toolchain to synthesize and visualize your SystemVerilog code as a schematic.


rasa (changes requested):
Please reorder the attributes per 
https://github.com/Ash258/GenericBucket/blob/master/.vscode/Template.jsonc
Thanks.

rasa (approved):

PR #1541, opened 2020-10-29
git: add less.exe bin

daniel-liuzzi:
`less` is relied upon by popular diff viewers such as [delta] or [diff-so-fancy], so including it makes sense. Furthermore, less v551
bundled with Git fixes issues in older versions. For instance, v394 bundled with Gow has [problems with UTF-8] and [an update seems unlikely], sadly.

[delta]: https://github.com/dandavison/delta
[diff-so-fancy]: https://github.com/so-fancy/diff-so-fancy
[problems with UTF-8]: https://github.com/dandavison/delta/issues/271
[an update seems unlikely]: https://github.com/bmatzelle/gow/issues/250

### DIFF
--- a/bucket/git.json
+++ b/bucket/git.json
@@ -17,6 +17,7 @@
         "cmd\\git.exe",
         "cmd\\gitk.exe",
         "cmd\\git-gui.exe",
+        "usr\\bin\\less.exe",
         "usr\\bin\\tig.exe",
         "git-bash.exe"
     ],


chawyehsu:
the git manifest should expose git related commands only. Exposing other binaries bundled in git can confuse other packages. You can use `scoop install less` if you want the `less` command.

rasa:
I would not add less to git, unless git itself has less as a dependency.

But I would support a git-plus or git-all package that exposes all of the exe's in the release

rasa:
See https://github.com/rasa/scoops/blob/master/bucket/git-all.json per above.

akirayamamoto:
As per discussion above, can we close this PR?

dwettstein:
Closed as a result of the discussion.

PR #1551, opened 2020-11-03
so: Add version 0.4.2

littleli:
### A terminal interface for Stack Overflow
- well-maintained project
- nice utility


Ash258:
https://github.com/lukesampson/scoop-extras/pull/4376

https://github.com/lukesampson/scoop-extras/blob/master/bucket/so.json

PR #1494, opened 2020-10-11
openssh: Update to version 8.3p1-1

wstein:
msys2 has purged many older file version!
Now many packages of **msys2** are only available as **zstd** compressed files!
The standard 7zip package does not support zstd, but the 7zip-zstd package from versions bucket.
To be able to install there is a workaround necessary:
- integrate this patch https://github.com/lukesampson/scoop/pull/4133/files
- replace 7zip package:
```
scoop bucket add versions
scoop install 7zip-zstd
scoop uninstall 7zip
scoop reset 7zip-zstd
```
please refer also: lukesampson/scoop#3990 which needs to be fixed, or msys2 based packages should be dropped.

This is necessary for most msys2 based packages!

Closes #1492

PR #1555, opened 2020-11-04
openssh: Fixed broken links

sanjifr3:
Updated broken links and changed corresponding hash codes.


rasa:
Effective duplicate of https://github.com/ScoopInstaller/Main/pull/1494/files . Blocked as scoop does not yet support unzipping of .zst files.

wumo:
Cannot install `openssh` now because of broken links of `heimdal`. It seems that only `zstd` file is supported for `heimdal`.
```powershell
PS C:\Users\***> scoop install openssh
Installing 'openssh' (8.2p1-1) [64bit]
Loading bash-4.4.023-2-x86_64.pkg.tar.xz from cache
Checking hash of bash-4.4.023-2-x86_64.pkg.tar.xz ... ok.
Loading gcc-libs-9.3.0-1-x86_64.pkg.tar.xz from cache
Checking hash of gcc-libs-9.3.0-1-x86_64.pkg.tar.xz ... ok.
The remote server returned an error: (404) Not Found.
URL http://repo.msys2.org/msys/x86_64/heimdal-7.7.0-1-x86_64.pkg.tar.xz is not valid
```

PR #1556, opened 2020-11-05
Add vcpkg

devwolf75:
VCPKG is a C++ package manager for Windows and other OS. It downloads sources instead of binaries and compiles them with the local Visual Studio instance. Integration with Visual Studio works per user or project.

### DIFF
--- a/bucket/vcpkg.json
+++ b/bucket/vcpkg.json
@@ -9,10 +9,16 @@
         "`vcpkg integrate install`",
         "The best way to use installed libraries with cmake is via the toolchain file scripts\\buildsystems\\vcpkg.cmake.",
         "To use this file, you simply need to add it onto your CMake command line as:",
-        "`-DCMAKE_TOOLCHAIN_FILE=D:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake`"
+        "`-DCMAKE_TOOLCHAIN_FILE=D:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake`",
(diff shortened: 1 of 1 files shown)


bucket/vcpkg.json
```diff
@@ -0,0 +1,42 @@
...
+    "homepage": "http://aka.ms/vcpkg",
+    "license": "MIT",
+    "notes": [
+        "vcpkg is installed with telemetry disabled flag (-disableMetrics) for privacy sake.",
+        "In order to use vcpkg with Visual Studio, run the following command (may require administrator elevation):",
+        "`vcpkg integrate install`",
+        "The best way to use installed libraries with cmake is via the toolchain file scripts\\buildsystems\\vcpkg.cmake.",
+        "To use this file, you simply need to add it onto your CMake command line as:",
+        "`-DCMAKE_TOOLCHAIN_FILE=D:\\src\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake`",
+        "Packages will be installed on $env:USERPROFILE\\.vcpkg\\."
```
rgozim, line 13: Is `$env:USERPROFILE` the best place to put vcpkg packages? Scoop has the _persist_ directory which might be better suited?

pratikpc:
Could we add NASM and other VCPKG app dependencies etc. as suggested dependencies that one could install via Scoop considering if VCPKG does not find them in the path, then it downloads them anyways on it's own.

Etlas suggests potential dependencies:- https://github.com/lukesampson/scoop-extras/blob/master/bucket/etlas.json

PR #1560, opened 2020-11-06
rakudo-star: Change description and add ZEF to bin

AntonOks:
taking the the description from https://rakudo.org/star and adding the **_zef_** tool to the bin / path to enable RAKU module management.

### DIFF
--- a/bucket/rakudo-star.json
+++ b/bucket/rakudo-star.json
@@ -1,7 +1,7 @@
 {
     "version": "2020.05.1-01",
-    "description": "Raku compiler for the MoarVM and JVM",
-    "homepage": "https://rakudo.org",
+    "description": "The Rakudo Star Bundle contains the Rakudo Compiler, a collection of modules from the Raku ecosystem, and the language documentation.",
+    "homepage": "https://rakudo.org/star",
     "license": "Artistic-2.0",
     "architecture": {
         "64bit": {
@@ -13,7 +13,9 @@
     "bin": [
         "bin\\perl6.exe",
         "bin\\raku.exe",
-        "bin\\rakudo.exe"
+        "bin\\rakudo.exe",
+        "share\\perl6\\site\\bin\\zef",
+        "share\\perl6\\site\\bin\\zef.bat"
     ],
     "checkver": {
         "url": "https://rakudo.org/downloads/star",
--- a/bucket/rakudo-star.json
+++ b/bucket/rakudo-star.json
@@ -14,8 +14,7 @@
         "bin\\perl6.exe",
         "bin\\raku.exe",
         "bin\\rakudo.exe",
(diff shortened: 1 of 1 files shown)


bucket/rakudo-star.json
```diff
@@ -13,7 +13,9 @@
     "bin": [
         "bin\\perl6.exe",
         "bin\\raku.exe",
-        "bin\\rakudo.exe"
+        "bin\\rakudo.exe",
+        "share\\perl6\\site\\bin\\zef",
```
rasa, line 17: This line will shim to the same name as the next line. One will need to be renamed. Is this a perl script without an extenstion?

AntonOks:
Yes, one is a PERL script and the other a bat...I can check how those scripts work without having PERL installed... if at all.Am 07.11.20 um 16:26 schrieb Ross Smith II
                
            
            
                
                    Von: "Ross Smith II" <notifications@github.com>Datum: 7. November 2020An: "ScoopInstaller/Main" <Main@noreply.github.com>Cc: "Anton Oks" <notna@ist-einmalig.de>,"Author" <author@noreply.github.com>Betreff: Re: [ScoopInstaller/Main] rakudo-star: Change description and add ZEF to bin (#1560)
                
                
@rasa commented on this pull request.



In bucket/rakudo-star.json:
> @@ -13,7 +13,9 @@
     "bin": [
         "bin\\perl6.exe",
         "bin\\raku.exe",
-        "bin\\rakudo.exe"
+        "bin\\rakudo.exe",
+        "share\\perl6\\site\\bin\\zef",

This line will shim to the same name as the next line. One will need to be renamed. Is this a perl script without an extenstion?

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/ScoopInstaller/Main/pull/1560#pullrequestreview-525660368",
"url": "https://github.com/ScoopInstaller/Main/pull/1560#pullrequestreview-525660368",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

bucket/rakudo-star.json
AntonOks: yes, it's a PERL6 script.
I guess I'll remove the zef.bat. Would that be OK for you?

AntonOks: Ok, tested now. If only one of those is allowed then if has to be zef.bat. Both Powershell and CMD work only with zef.bat.
CMD doesn't like the plain zef :|

chawyehsu (approved):

PR #1574, opened 2020-11-11
fnm: Add version 1.22.8

Armaldio:


bucket/fnm.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "1.22.8",
+    "description": "Cross-platform Node.js version switcher",
+    "homepage": "https://github.com/Schniz/fnm",
+    "license": "GPL-3.0-only",
+    "notes": "Add \"fnm env --use-on-cd | Out-String | Invoke-Expression\" to your powershell profile to enable automatic version switch",
```
Schniz, line 6: `fnm env | Out-String | Invoke-Expression` is needed even if you don't want automatic version switching — the `--use-on-cd` allows the automatic version switching. `fnm env` is a must to make fnm work, just like Starship's init:

https://github.com/ScoopInstaller/Main/blob/2f021cf8ad1eb726f93e9f32668606151899f8df/bucket/starship.json#L7

Armaldio: Thanks !

chawyehsu (approved):

PR #1580, opened 2020-11-12
git-with-openssh: add ssh-copy-id

### LINKED ISSUE #1458: git-with-openssh: add `ssh-copy-id`
It's not currently included as an executable, making it necessary to install `git` and `openssh` separately to access the command.

Teomik129:
Resolves #1458

### DIFF
--- a/bucket/git-with-openssh.json
+++ b/bucket/git-with-openssh.json
@@ -25,6 +25,11 @@
         "usr\\bin\\sftp.exe",
         "usr\\bin\\ssh-add.exe",
         "usr\\bin\\ssh-agent.exe",
+        [
+            "bin/sh.exe",
+            "ssh-copy-id",
+            "ssh-copy-id"
+        ],
         "usr\\bin\\ssh-keygen.exe",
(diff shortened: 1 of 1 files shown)


rasa (approved):

bucket/git-with-openssh.json
```diff
@@ -25,6 +25,11 @@
         "usr\\bin\\sftp.exe",
         "usr\\bin\\ssh-add.exe",
         "usr\\bin\\ssh-agent.exe",
+        [
+            "bin/sh.exe",
+            "ssh-copy-id",
+            "ssh-copy-id"
```
akirayamamoto, line 31: Is there any reason to have this line duplicated?

Teomik129: Yes; this is using the alias shim format per the [manifest docs](https://github.com/lukesampson/scoop/wiki/App-Manifests#:~:text=bin%3A), which is of the form `[ "program.exe", "alias", "--arguments" ]`. Without the duplicate `ssh-copy-id`, this would call `bin/sh.exe` with no arguments which isn't the desired behavior.

PR #1619, opened 2020-11-27
gping: Add version 0.1.7

wenmin92:
[gping](https://github.com/orf/gping), is a ping tool with graph.

See https://github.com/orf/gping/issues/106

### DIFF
--- a/bucket/gping.json
+++ b/bucket/gping.json
@@ -1,0 +1,21 @@
+{
+    "version": "0.1.7",
+    "description": "Ping, but with a graph",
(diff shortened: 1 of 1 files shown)


rasa:
Current version is `v1.0.1-post2` which fails thr internal github version regex.

wenmin92:
> Current version is `v1.0.1-post2` which fails thr internal github version regex.

Thanks, will use explict checkver url and regex to fix it.

chawyehsu (approved):

PR #1641, opened 2020-12-02
gcc: Update to version 10.2.0 and add mingw64 build of llvm

omerakgoz34:
"make" shim has been removed due to conflicts with CMake.

### DIFF
--- a/bucket/gcc-llvm
+++ b/bucket/gcc-llvm
@@ -1,0 +1,19 @@
+{
+    "version": "10.2.0-11.0.0-9.0.0-r3",
+    "description": "winlibs standalone build of GCC-MinGW-w64 and LLVM-Clang compilers)",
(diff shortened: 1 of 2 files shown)


chawyehsu:
winlibs-mingw is a personal standalone build of gcc and mingw64, therefore it's not a good idea to replace msys2's build of gcc with that. There are so many different gcc and mingw builds on the Internet, such as nuwen's mingw, mstorsjo's llvm-mingw, etc.

The better way to support winlibs-mingw is adding a new manifest for it like [this](https://github.com/chawyehsu/dorado/blob/master/bucket/winlibs-mingw.json) and [this](https://github.com/chawyehsu/dorado/blob/master/bucket/llvm-mingw.json). You should find the advantages of creating different manifests.

PR #1665, opened 2020-12-09
puppet-bolt: Set GEM_PATH

dsisnero:
Like GEM_HOME, GEM_PATH needs to be reset to use the puppet-bolt gems.   

TODO
(can we have multiple pre_install keys?)
We also should create a $dir\\gem.bat file that resets GEM_HOME and GEM_PATH


dsisnero:
can we get this merged and approved.  puppet-bolt doesn't work without this change.  Thanks

chawyehsu (approved):

PR #1705, opened 2020-12-24
Update ffmpeg.json

MercuryOoO:
Update to 2020-12-23-12-35 build release

### DIFF
--- a/bucket/ffmpeg.json
+++ b/bucket/ffmpeg.json
@@ -5,8 +5,8 @@
     "license": "GPL-3.0-or-later",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


MercuryOoO:
@Ash258 What the hell just with thumbs down? What you mean? I'm confused.

Ash258:
It does not make sense. The download works and update stable build of applicaiton every day is ...

MercuryOoO:
> It does not make sense. The download works and update stable build of applicaiton every day is ...

Well, the understanding of the update mechanism may be different between me and you. Whether to choose to update is a matter for the user, and whether to reflect the update is a matter of Scoop. However, manual updates are really cumbersome, and perhaps there should be a mechanism for automatically tracking upstream version updates.

PR #1714, opened 2020-12-26
volta: Add version 1.0.0

myty:


batkiz:
Check this: https://github.com/chawyehsu/dorado/blob/master/bucket/volta.json

myty:
> Check this: https://github.com/chawyehsu/dorado/blob/master/bucket/volta.json

I'm cool with using this manifest.  It seems to be closely following what the volta installer does.

felipeqq2:
Will this be merged? 😕

PR #1727, opened 2021-01-03
Add gcc arm none eabi

uyha:


Ash258:
https://github.com/lukesampson/scoop-extras/blob/master/bucket/gcc-arm-none-eabi.json

Issue #1492, opened 2020-10-11
openssh fails to install, heimdall fails to download


Merith-TK:

NemoStein:
Still having the problem... Any news on the topic?

PR #1733, opened 2021-01-05
Update openssh package

jackdown:
Try to fix the heimdal issue mention in #1492


rasa:
Scoop doesn't current support .zst files.

igitur:
> Scoop doesn't current support .zst files.

@rasa Are you opposed to adding support for .zst files? I see there are 7zip plugins available at https://github.com/mcmilk/7-Zip-zstd . If you're OK with that, I can look into a submitting a PR for this support.

PR #1741, opened 2021-01-08
Change LLVM update URL to GitHub Releases and update to 11.0.1

pratikpc:
https://github.com/llvm/llvm-project/releases/latest

LLVM is quite possibly now that they have completed a move to GitHub are not maintaining their website.

LLVM 11.0.1 while is listed as a release on GitHub Releases, is not listed as a release on their website

Since they are using GitHub releases, I think it would be best to update the project via GitHub Releases as well.

### DIFF
--- a/bucket/llvm.json
+++ b/bucket/llvm.json
@@ -1,16 +1,16 @@
 {
-    "version": "11.0.0",
+    "version": "11.0.1",
     "description": "Collection of modular and reusable compiler and toolchain technologies.",
     "homepage": "https://www.llvm.org",
     "license": "NCSA",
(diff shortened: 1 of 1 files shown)


pratikpc:
I think checkver is failing because there is an update available which is why I made the change.

I think updating the version is also handled by Scoop automatically?

So should I run it on my end and paste the updated JSON for LLVM 11.0.1 or should I let it be?

pratikpc:
I updated it to LLVM 11.0.1 to get rid of the package-fix-needed label

rasa (approved):

PR #1743, opened 2021-01-08
Add mob 1.1.0

WtfJoke:
Adds [mob](https://github.com/remotemobprogramming/mob) to scoop. 

Mob is a tool for fast git handover while mob programming.
I hope I included everything which is needed. I also verified that autoupdate works using `checkver.ps1`
In case something is off, let me know.

### DIFF
--- a/bucket/mob.json
+++ b/bucket/mob.json
@@ -1,0 +1,18 @@
+{
+    "version": "1.1.0",
+    "description": "Swift git handover with mob",
+    "homepage": "https://mob.sh",
+    "license": "MIT",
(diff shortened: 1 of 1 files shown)


bucket/mob.json
```diff
@@ -0,0 +1,16 @@
+{
+    "version": "1.1.0",
+    "description": "Swift git handover with mob",
+    "homepage": "https://github.com/remotemobprogramming/mob",
```
pratikpc, line 4: Mob seems to have a homepage here

https://mob.sh/

You might have to update checkver here.
Refer to https://github.com/ScoopInstaller/Main/blob/master/bucket/isx.json if required

WtfJoke: Thanks for the hint. 
I updated the homepage and checkver

rasa (approved):

PR #1750, opened 2021-01-12
objconv: Add version 2020-Jun-08

YenForYang:
https://www.agner.org/optimize/#objconv.
`pre_install` is for flattening a deeply nested folder.

### DIFF
--- a/bucket/objconv.json
+++ b/bucket/objconv.json
@@ -1,0 +1,16 @@
+{
+    "version": "2020-Jun-08",
+    "description": "Object file converter and dissassembler",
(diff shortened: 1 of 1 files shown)


rasa (approved):

rasa:
@ash258 I merged this, but is there a package fix that needs to be made?

Ash258:
It was done by my bot, which I accidently enabled again on Main.

But because this PR is in horrible state. 

- Version used is not the real version 
- Properties are sorted random
- Regex is ridiculously complex and ugly
- pre_install script should be solved with one simple Move-Item

Issue #1086, opened 2020-05-14
gcc 9.3.0-1 install failed


kurtlau:
```powershell
Windows PowerShell
版权所有 (C) Microsoft Corporation。保留所有权利。

尝试新的跨平台 PowerShell https://aka.ms/pscore6

PS C:\Users\kurtliu> scoop install gcc
Installing 'gcc' (9.3.0-1) [64bit]
mingw-w64-x86_64-binutils-2.34-1-any.pkg.tar.xz (13.9 MB) [===================================================] 100%
Checking hash of mingw-w64-x86_64-binutils-2.34-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-crt-git-8.0.0.5685.ef310786-1-any.pkg.tar.xz (3.1 MB) [======================================] 100%
Checking hash of mingw-w64-x86_64-crt-git-8.0.0.5685.ef310786-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-expat-2.2.9-1-any.pkg.tar.xz (141.3 KB) [====================================================] 100%
Checking hash of mingw-w64-x86_64-expat-2.2.9-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-gcc-9.3.0-1-any.pkg.tar.xz (33.0 MB) [=======================================================] 100%
Checking hash of mingw-w64-x86_64-gcc-9.3.0-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-gcc-fortran-9.3.0-1-any.pkg.tar.xz (9.6 MB) [================================================] 100%
Checking hash of mingw-w64-x86_64-gcc-fortran-9.3.0-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-gcc-libgfortran-9.3.0-1-any.pkg.tar.xz (627.3 KB) [==========================================] 100%
Checking hash of mingw-w64-x86_64-gcc-libgfortran-9.3.0-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-gcc-libs-9.3.0-1-any.pkg.tar.xz (687.8 KB) [=================================================] 100%
Checking hash of mingw-w64-x86_64-gcc-libs-9.3.0-1-any.pkg.tar.xz ... ok.
mingw-w64-x86_64-gettext-0.19.8.1-8-any.pkg.tar.xz (3.1 MB) [=================================================] 100%
[...]

PR #1761, opened 2021-01-17
gcc: expat package update to 2.2.10-1 and zst compression?!?

dionysius:
> URL http://repo.msys2.org/mingw/x86_64/mingw-w64-x86_64-expat-2.2.9-1-any.pkg.tar.xz is not valid

Hi, couldn't install `gcc` because package `...2.2.9-1...tar.xz` is no longer available. I've updated it to `2.2.10-1`, but the file extension is tar.zst!? Do we support that?

While at it, I was looking through all the files. The same issue also with:
- gettext
- libiconv
- pkg-config
- zlib


dionysius:
Just found matching issue #1086 , I have to add 2 more i686 updates, one second

dionysius:
done, those two i686 packages according to the issue has been added as well

chawyehsu:
> and zst compression?!?

zst is not supported by scoop's decompression functionality currently, thus zst-related updates will not be merged until the support launch, there is a hanging [PR](https://github.com/lukesampson/scoop/pull/4137) waiting for discussions and reviews.

For the time being, I suggest using alternative gcc packages like [llvm-mingw](https://github.com/chawyehsu/dorado/blob/master/bucket/llvm-mingw.json), [nuwen-mingw-gcc](https://github.com/chawyehsu/dorado/blob/master/bucket/nuwen-mingw-gcc.json), or [winlibs-mingw](https://github.com/chawyehsu/dorado/blob/master/bucket/winlibs-mingw.json).

dionysius:
Thank you, one can come back to this PR once the support has been added. So far I helped as far as I could.

PR #1774, opened 2021-01-23
Quick fix for gcc and gdb x64 and https

yostane:
- Fixed failed links for gcc and gdb x64
- Replaced http by https for x64
- Fixes
  - https://github.com/lukesampson/scoop/issues/4235
  - https://github.com/lukesampson/scoop/issues/4224


yostane:
I just noted that Scoop does not handle .tar.zst

BurnoutAg7 (approved):
I' m not a experienced developer, but I can approve that it worked as I successfully installed gcc & gdb through this fix. :P

yostane:
Hello, thanks for the review. However, I think that one fo those PR are also required for the fix to be complete:  https://github.com/lukesampson/scoop/pull/4137 or https://github.com/lukesampson/scoop/pull/4240

yostane:
> I' m not a experienced developer, but I can approve that it worked as I successfully installed gcc & gdb through this fix. :P

Even though the install succeeds, some components will not be decompressed and will not be usable.

AwesomeCronk:
I saw that there was a scoop package that installs the program for zst archives. What if that package was added to the gcc package’s dependencies and the after install section had the extraction command added to it?

starptr:
In case anyone else stumbles on this, we need to wait for the extraction script to support zstd first (see lukesampson/scoop#3990) before this manifest can work again. The pr lukesampson/scoop#4372 seems to fix this.

rashil2000:
Blocked via #1761

PR #1776, opened 2021-01-24
adjust env_add_path field to solve some problems.

HUMORCE:
duplicate paths when use dot(.) to point to the app directory.

but using \ does not have the problem.

example ( python / pycharm)
before:
https://imgur.com/47LjPSn
after:
https://imgur.com/yiBTYHy

### DIFF
--- a/bucket/dotnet-sdk.json
+++ b/bucket/dotnet-sdk.json
@@ -17,7 +17,7 @@
         }
     },
     "bin": "dotnet.exe",
-    "env_add_path": ".",
+    "env_add_path": "\\",
     "env_set": {
         "DOTNET_ROOT": "$dir",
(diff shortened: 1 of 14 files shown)


bucket/tesseract-languages.json
```diff
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/tesseract-ocr/tessdata_fast",
     "license": "Apache-2.0",
     "notes": [
-        "This packages has overwritten the 'TESSDATA_PREFIX' environment variable with \"$dir\".",
+        "This packages has overwritten the 'TESSDATA_PREFIX' environment variable with \"$dir\"\\",
```
r15ch13, line 7: That's the end of a sentence :grin:

PR #1786, opened 2021-01-28
Add aws_completer.exe into the path

iguyking:
Enable aws_completer as a binary to add to the path to support tooling that enables auto completion on the command line.


0cool-f (approved):

iguyking:
Is there something I need to fix here?  Is this a valid error?

Issue #1784, opened 2021-01-27
[Request] Update Haskell


glatavento:
GHC which is independent of haskell-stack is still useful in some situations (e. g. vscode haskell extension). Latest versions of GHC can be found [here](https://downloads.haskell.org/~ghc/), and hash codes can be found [here](https://github.com/commercialhaskell/stackage-content/blob/master/stack/stack-setup-2.yaml).

glatavento:
```
{
    "version": "8.10.3",
    "description": "An advanced, purely functional programming language.",
    "homepage": "https://www.haskell.org",
    "license": "BSD-3-Clause",
    "notes": [
        "For Haskell Stack, please install 'stack'.",
        "For 32bit Haskell, please install 'versions/haskell863'."
    ],
    "suggest": {
        "Stack": "stack"
    },
    "architecture": {
        "64bit": {
            "url": "https://downloads.haskell.org/~ghc/8.10.3/ghc-8.10.3-x86_64-unknown-mingw32.tar.xz",
            "hash": "927a6c699533a115cd49772ef2c753d9af2c13bf9f0b2d3bd13645cc6a144ee3"
        }
    },
    "pre_install": [
		"Rename-Item \"$dir\\ghc-$version\" ghc"
    ],
    "bin": [
        "ghc\\bin\\ghc.exe",
        "ghc\\bin\\ghci.exe",
        "ghc\\bin\\ghc-pkg.exe",
        "ghc\\bin\\haddock.exe",
        "ghc\\bin\\hp2ps.exe",
        "ghc\\bin\\hpc.exe",
        "ghc\\bin\\hsc2hs.exe",
        "ghc\\bin\\runghc.exe",
        "ghc\\bin\\runhaskell.exe"
    ],
    "env_add_path": "lib\\bin",
    "checkver": {
        "github": "https://github.com/commercialhaskell/ghc"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://downloads.haskell.org/~ghc/$version/ghc-$version-x86_64-unknown-mingw32.tar.xz"
            }
        }
    }
}
```

This seems to be working fine for me. `checkver` and `autoupdate` are not tested.

glatavento:
see https://github.com/ScoopInstaller/Main/pull/2790

PR #1792, opened 2021-02-01
Update haskell.json

glatavento:


glatavento:
close https://github.com/ScoopInstaller/Main/issues/1784

PR #1803, opened 2021-02-08
nheko: Add version 0.8.1

jaysonsantos:
*IMPORTANT* This is still incomplete because I wanted to ask some questions.
I didn't find any example of qt installer framework and I could not disable the operations that create the shortcuts, do you have any idea on how to do it? Maybe I could add a post install to delete the shortcuts and allow scoop to create its own shortcuts that point to the right paths.

### DIFF
--- a/bucket/nheko.json
+++ b/bucket/nheko.json
@@ -1,0 +1,41 @@
+{
+    "version": "0.8.1",
+    "description": "Desktop client for Matrix using Qt and C++17.",
+    "homepage": "https://nheko-reborn.github.io/",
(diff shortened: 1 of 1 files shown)


earnestma:
This patch should probably be moved into either **[extras](https://github.com/lukesampson/scoop-extras/)** or [non-portable](https://github.com/TheRandomLabs/scoop-nonportable) bucket.

[Criteria for main bucket](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket)

PR #1821, opened 2021-02-15
jfrog: Update to version 1.44.0

### LINKED ISSUE #1969: JFrog CLI Frozen at 1.43.2
The JFrog CLI is not autoupdating and is now stuck at 1.43.2. The checkver URL is valid but seems to stop at v 1.44.0. Latest is 1.45.2. PR #1821 seems to be a fix but is waiting on review.

nixxo:
address issue in #150

Close #1969

### DIFF
--- a/bucket/jfrog.json
+++ b/bucket/jfrog.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.43.2",
+    "version": "1.44.0",
     "description": "Command Line Interface for Artifactory and Bintray",
     "homepage": "https://jfrog.com/getcli/",
     "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


pratikpc (changes requested):
Could you also change checkver to GitHub?
```json
    "checkver": {
        "github": "https://github.com/jfrog/jfrog-cli"
    },
```
Using a GitHub checkver is usually more popular with most Scoop projects and usually less prone to breakage.

nixxo:
I think it's better to check the version in the same repository where the build is available.
On github is only available the source code and not the build package.

I don't know the release schedule of the jfrog project and I don't know if they release the source code and the build at the same time or maybe they release the source code before and the nthe build server creates the release... etc.

So I think it's better to check the version in the same place where you actually download the package because it gives you exactly the latest available package to download.

chawyehsu (approved):

PR #1831, opened 2021-02-18
Update outdated link

kenbaird13:
Line 11 points to an old link. Changed it to the current link.


pratikpc:
Scoop does not support zst yet sadly

For OpenSSH scoop users, using the git based packages all of which also ship ssh seems to be a nice alternative

PR #1832, opened 2021-02-19
Update gcc.json

orpharion:
Previous dependencies no longer in msys repo


pratikpc:
Does Scoop actually support zstd files?

I thought it does not

See also https://github.com/lukesampson/scoop/issues/3990

orpharion:
It did download and build correctly

Vektorweg:
> It did download and build correctly

It silently failed to unpack the zst. 

GCC itself doesn't depend on the listed dependencies, but programs to be compiled with GCC might expect that you have these.

chawyehsu:
duplicate of #1761

PR #1860, opened 2021-02-23
Add PG_LIB_DIR and PG_INCLUDE_DIR env vars

alextwothousand:
On most linux distros, you can just install postgresql-dev from whatever package manager.

This PR hopes to replicate that process but for Windows instead, as I believe Scoop can really pave the way for Windows development.

The new env variables provded:
* PG_LIB_DIR
* PG_INCLUDE_DIR

### DIFF
--- a/bucket/postgresql.json
+++ b/bucket/postgresql.json
@@ -28,7 +28,9 @@
     ],
     "env_add_path": "bin",
     "env_set": {
-        "PGDATA": "$dir\\data"
+        "PGDATA": "$dir\\data",
+        "PG_LIB_DIR": "$dir\\lib",
+        "PG_INCLUDE_DIR": "$dir\\include"
     },
     "persist": "data",
     "checkver": {


alextwothousand:
Hi? Any reviewers available?

rashil2000:
@alextwothousand is there any documentation available on what variables should be set?

rashil2000:
As far as I could find, these env vars are only needed when building Postgres from source, which isn't the case here.

alextwothousand:
Incorrect. My original purpose for it was to allow me to utilise the included SDK.

rashil2000:
That is why I asked for some documentation source regarding them. Can you show us?

alextwothousand:
> That is why I asked for some documentation source regarding them. Can you show us? 

Well, I hadn't received a response in 3 months, so I didn't bother.

There is nothing to show, if you use an IDE such as Visual Studio, it will allow you to input environment variables as paths to directories which contain static/shared libraries and header files.

rashil2000:
If they aren't well known and have no Windows related documentation, I think we should not add them for a singleton case like this.

Users can set it themselves, and Scoop will not touch them.

alextwothousand:
What do you mean, "aren't well known"? How is someone supposed to communicate with a database without accessing postgres inbuilt library?

https://www.postgresql.org/docs/7.2/libpqplusplus.html

PR #1873, opened 2021-02-27
Update openssl-mingw.json

helohe:
fixed extract error


rasa:
Thank you for your submission!

We also need to update
https://github.com/ScoopInstaller/Main/blob/9db2f54dfd2a43cab48d4675032c6db572d8a35c/bucket/openssl-mingw.json#L28-L32
as well, and replace `$version` with `$matchHead`, per
https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate#version-variables

PR #1875, opened 2021-02-28
xh: Update to version 0.8.0

ducaale:


r15ch13 (approved):

r15ch13:
To get the `xhs` feature we could simply copy `xh.exe` to `xhs.exe` and create a shim for that.

ducaale:
Would we do something like this to create a shim for the second executable?
```json
{
  "bin": [
    "xh.exe",
    ["xhs.exe", "xhs.exe"]
  ]
}
```

r15ch13:
The shim itself will not work, because it just calls `xh.exe` and that is not detected by `app.get_bin_name()`.
A simple `pre_install` script just copies the file. This keeps the package small and just works :smile:

```json
    "bin": [
        "xh.exe",
        "xhs.exe"
    ],
    "pre_install": "Copy-Item \"$dir\\xh.exe\" \"$dir\\xhs.exe\"",
```

image

ducaale:
Oh, that's very convenient. I will open a PR to add that 👍

PR #1878, opened 2021-03-02
Update urls for 64 bit packages

alaurie:
Updated outdated package urls and hashes for 64 bit packages.


pratikpc:
At the moment, I don't think Scoop supports zst files

As such, the unpackings are quite possibly silently failing.

Also see similar PR here at https://github.com/ScoopInstaller/Main/pull/1832

chawyehsu:
duplicate of #1761

Issue #1894, opened 2021-03-07
Upgrade yarn to version 1.2.10


altro3:
Yarn 1.22.10 was published 5 months ago, but it's still version 1.22.5 in the bucket 

https://www.npmjs.com/package/yarn?activeTab=versions

Please, upgrade yarn to latest version.

HUMORCE:
https://classic.yarnpkg.com/

1.22.5 is the stable version.

altro3:
@HUMORCE Hi!

Hm.. very strange. I think this is a bug on the yarn page because yarn 1.X is now frozen. 

But you can look at their official repository in github and npm - latest version 1.22.10. If you try to install yarn through the npm repository, you will have version 1.22.10 installed. 

LInks:

Yarn 1.22.10 Release https://github.com/yarnpkg/yarn/releases/tag/v1.22.10

Npm yarn
![изображение](https://user-images.githubusercontent.com/48474279/110299059-da54c300-8027-11eb-925a-8a2e6c515d55.png)

HUMORCE:
https://github.com/yarnpkg/yarn/releases/tag/v1.22.10

> Please check the changelog for details: https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md
For technical reasons, **no prebuilt artifacts will be generated for this release**. Please use npm install -g yarn to install it.

altro3:
yeah, this is the last release of yarn 1. Тo more releases are planned. But fix version 1.22.6 is also not included in version 1.22.5.

I understand that the changes are minor in recent releases, but the fact is the fact - the latest version is 1.22.10

PR #1900, opened 2021-03-08
yarn: Update to version 1.22.10

HUMORCE:
Fixes https://github.com/ScoopInstaller/Main/issues/1894


pratikpc:
According to https://scoop.netlify.app/concepts/#app-manifests, extract_to is
1. No longer used.
2. Deprecated property

PR #1933, opened 2021-03-15
fixed error by mirror

tldzyx:
use mirrors.huaweicloud.com replace repo.msys2.org temp, until upstream ready.

使用华为镜像暂代, 确保当前可用.


scrouthtv:
Thanks for your work.

However, I don't think that's really a fix. There are multiple PRs waiting on `scoop` for zstd support:
 - https://github.com/lukesampson/scoop/pull/4268
 - https://github.com/lukesampson/scoop/pull/4240
 - https://github.com/lukesampson/scoop/pull/4137

The last one being over 5 months old. I am in no way involved in scoop, but I think that the original author has disappeared and since then, this project has stalled.

tldzyx:
> Thanks for your work.
> 
> However, I don't think that's really a fix. There are multiple PRs waiting on `scoop` for zstd support:
> 
> * [lukesampson/scoop#4268](https://github.com/lukesampson/scoop/pull/4268)
> * [lukesampson/scoop#4240](https://github.com/lukesampson/scoop/pull/4240)
> * [lukesampson/scoop#4137](https://github.com/lukesampson/scoop/pull/4137)
> 
> The last one being over 5 months old. I am in no way involved in scoop, but I think that the original author has disappeared and since then, this project has stalled.

however, this make the latest version can be work, it better the file not exists, and new config need time.

scrouthtv:
No, my point is that there is no one who is able to merge this.

JeffCarpenter:
@scrouthtv how can this be? do we need to fork?

scrouthtv:
@JeffCarpenter I guess we need a maintainer to come forward and clarify the situation. If you (or anyone else) creates another fork on top of the 1k that already exist, the mess just gets bigger. If lukesampson won't (or can't) maintain the original repository anymore, the best way would be to create an active fork that should replace lukesampson's version **and also put a notice on lukesampson/scoop** referencing the active fork.

@MehmetUmit just pushed to the main repo, maybe they know more.

JeffCarpenter:
@scrouthtv I think the best outcome for such a fork would be to light a fire under the maintainers and maybe use it as a sort of temporary dev branch. Although it could feature configurable community package repos like other package managers do. 🤔

PR #1954, opened 2021-03-20
Updated AVR-GCC to version 10.2.0-1

LumitoLuma:
Changes from previous release: https://www.lumito.net/2021/03/20/released-avr-gcc-10-2-0-1/

Also, this release does not use Bintray. It downloads the files from my server.


LumitoLuma:
This does not work (idk why)

PR #1957, opened 2021-03-21
mozjpeg: Update to version 4.0.3

ddarfantasy:
Also switch to the official build.


ddarfantasy:
Well, that's meaningless now.

PR #1972, opened 2021-03-23
nano: fix shim

jerquiaga:
Looks like the upstream packager changed the names of the folder structure in their zip, this fixes it so it will install with Scoop.

### DIFF
--- a/bucket/nano.json
+++ b/bucket/nano.json
@@ -20,10 +20,10 @@
     },
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


0cool-f (approved):

jerquiaga:
@r15ch13 Any chance you can help with this PR? Looks like you committed the most recent nano manifest.

chawyehsu (approved):

PR #1979, opened 2021-03-24
Update to latest QEMU installer to fix firmware file lookup issues

phil-opp:
This should include a fix for https://bugs.launchpad.net/qemu/+bug/1915794

### DIFF
--- a/bucket/qemu.json
+++ b/bucket/qemu.json
@@ -5,8 +5,8 @@
     "license": "GPL-2.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://qemu.weilnetz.de/w64/2021/qemu-w64-setup-20210203.exe#/dl.7z",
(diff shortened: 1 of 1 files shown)


phil-opp:
Rebased to fix CI issues.

phil-opp:
Hmm, I don't think that my pull request has anything to do with the [failure](https://github.com/ScoopInstaller/Main/runs/2198080309?check_suite_focus=true) of the autoupdate check. The problem is that https://qemu.weilnetz.de/ has a "New QEMU installers" history entry, which only applies to the 64-bit version. Thus, the autoupdate looks for a non-exisitng 32-bit version. Is there an easy way to solve this?

phil-opp:
This was fixed by e44b7b841884fc9a23f8d2dcb63b74ad89ce373e.

PR #1985, opened 2021-03-25
Add go-task

telenieko:
Go task is a task runner (like Make or Ant) written in golang.

It's power comes from being fast and lightweight (one 2MB exe file) which even embeds a light shell runner.

Regarding [the criteria for inclussion in the main bucket](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket):

- go-task has 3.2k stars in github,199 forks and 54 contributors. I do not know if that qualifies for "widely used".
- the manifest points to the latest version and includes autoupdate
- it is the full version (open source project)
- install is fairly standard (unzip)
- it is not a GUI tool

A side of those official criteria, I'd add that it is a quite active project. It is my tool of choice for replacing Makefiles on every platform (it's golang, runs everywhere).

If it would be better to submit to the extras bucket just let me know

marc

### DIFF
--- a/bucket/go-task.json
+++ b/bucket/go-task.json
@@ -1,0 +1,33 @@
+{
+    "version": "3.3.0",
+    "description": "A task runner / simpler Make alternative written in Go",
+    "homepage": "https://taskfile.dev/",
+    "license": "MIT",
+    "bin": "task.exe",
+    "architecture": {
+        "32bit": {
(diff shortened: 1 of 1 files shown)


telenieko:
Hi, is there anything else I can to to have this merged?

telenieko:
Manifest updated for 3.7.0, please let me know if all is good.

PR #2025, opened 2021-04-05
Fix mtn v3.4 extract_dir

krokofant:
The folder structure for 3.4 if different so the installation breaks


krokofant:
Scrap this, I'll update to the latest 3.4.1 and fix the autoupdate in a new PR.

PR #2060, opened 2021-04-08
Using the portable zip files instead of innosetup

ShoofLLC:
currently innosetup throws an error when the version used is older than the version required, this fixes that for future releases too.


l2ichshop2 (approved):

ShoofLLC:
so is scoop not accepting pull requests?

PR #2064, opened 2021-04-09
fat32format: Add version 1.07

kriswilk:


pratikpc:
@kriswilk could you add a checkver and an auto-update?

kriswilk:
> @kriswilk could you add a checkver and an auto-update?

I've just added these, although I'm pretty new at this so not sure how to properly test it. I'm sure you do!

rasa:
It's GPL, but
https://github.com/rasa/scoops/blob/master/bucket/fat32format.json
is maintained.

kriswilk:
> It's GPL, but
> https://github.com/rasa/scoops/blob/master/bucket/fat32format.json
> is maintained.

Nice find, I wasn't aware of this fork. Seems like it has worthwhile improvements. Shall I revise this PR accordingly? Would be good to have whichever is the generally preferred version in the main bucket.

kriswilk:
Any direction on this? Although I've always used and been happy with the "official" application I'm happy to revise the PR if necessary. 

Looking forward to this being added to the main bucket either way...

PR #2092, opened 2021-04-15
Add conftest

jpreese:
Add `conftest` to scoop.

https://github.com/open-policy-agent/conftest


stevehipwell:
I've added a new PR https://github.com/ScoopInstaller/Main/pull/2126 with auto updating enabled that would supersede this.

PR #2121, opened 2021-04-28
pdfbox: Update to version 2.0.23

sualeh:


pratikpc:
Would be nice if we could update checkver too

PR #2131, opened 2021-04-30
Add pulseaudio

chuxubank:


pratikpc:
Possible to add a checkver?

pratikpc:
> Though PulseAudio was designed for Linux systems, it has also been ported by independent developers to Solaris, FreeBSD, NetBSD, MacOS X, Windows 2000 and Windows XP. We do not currently offer support for non-Linux systems, but patches for these systems are welcome and appreciated. If you're having issues on those systems, you must fix them yourself. Windows binaries are not provided by this site.

https://www.freedesktop.org/wiki/Software/PulseAudio/

This works on Windows??

pratikpc:
Wiki says latest version is 14.2 while you are adding 1.1

https://en.wikipedia.org/wiki/PulseAudio

chuxubank:
> This works on Windows??

Yep, detailed in this link: https://www.freedesktop.org/wiki/Software/PulseAudio/Ports/Windows/Support/
I'm using it to play audio from my WSL, so it turns out to be functional. The Windows binary also linked in this page.

> Wiki says latest version is 14.2 while you are adding 1.1

The Windows version doesn't update since about 9 years ago. (But it still works)
https://build.opensuse.org/package/show/home:mkbosmans:mingw32:pulseaudio/mingw32-pulseaudio

pratikpc:
> I'm using it to play audio from my WSL, so it turns out to be functional.

Yeah but WSL is kind of a different scenario though tbh

Have you tested the Windows Binary on Win10?
What were the results like?
I don't make the merges but details like this will make it easy for those that actually merge the PRs

chuxubank:
> > I'm using it to play audio from my WSL, so it turns out to be functional.
> 
> Yeah but WSL is kind of a different scenario though tbh
> 
> Have you tested the Windows Binary on Win10?
> What were the results like?
> I don't make the merges but details like this will make it easy for those that actually merge the PRs

Yes, I'm using Win10.
You can see the tutorial from x410(X Server for Windows 10) official site.
https://x410.dev/cookbook/wsl/enabling-sound-in-wsl-ubuntu-let-it-sing/

pratikpc:
Based on the status I can see, I don't feel like Checkver would be very useful because there is very little hope that it will receive updates

But yeah, this looks fairly useful although it would be nice if there would be a way to share the tutorial link with the users with the approval of the maintainers of course.

An example would be [vscode which uses notes](https://github.com/lukesampson/scoop-extras/blob/2a4981f55e48bb2bfba02e838ff963439bd2e245/bucket/vscode-portable.json#L9)

chuxubank (commented):
So can I just review myself?

Issue #2138, opened 2021-05-01
Upgrade pypy2 and pypy3 to version 7.3.4


denioma:
[PyPy](https://downloads.python.org/pypy/) was updated on 2021-04-08. It's now only released for 64bit architectures.

PR #2139, opened 2021-05-01
Update PyPy2 and PyPy3

denioma:
Changed manifests to account for 32bit to 64bit change in PyPy release, updated to the latest version.


denioma:
Related: issue #2138

PR #2144, opened 2021-05-04
python: persist user installed libs and scripts

### LINKED ISSUE #1680: Python should persist site-packages and scripts across minor versions
Currently site-packages are being stored inside of the app installation directory:
e.g. `~\scoop\apps\python\3.9.0\Lib\site-packages`

When you do minor upgrade of the package, e.g. from 3.9.0 to 3.9.1, all the installed packages and scripts remain with the previous version and you need to install everything again.

It would be great if all packages and scripts would be persisted outside of the app directory:
e.g.`~\scoop\persist\python\3.9\site-packages`

So it would be shared between 3.9.0 and 3.9.1 but not between 3.6.x and 3.9.x

dwettstein:
This PR add the folders `Scripts` and `Lib\\site-packages` to the list to persist.

Resolve #1680, #772

Overhauls PR #1483

### DIFF
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -79,6 +79,10 @@
             "idle3"
         ]
     ],
+    "persist": [
+        "Scripts",
+        "Lib\\site-packages"
+    ],
     "env_add_path": [
         "Scripts",
         "."
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -46,7 +46,7 @@
     "installer": {
         "script": [
             "Expand-DarkArchive \"$dir\\setup.exe\" \"$dir\\_tmp\"",
-            "@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {",
+            "@('path.msi', 'pip.msi') | ForEach-Object {",
             "    Remove-Item \"$dir\\_tmp\\AttachedContainer\\$_\"",
             "}",
             "(Get-ChildItem \"$dir\\_tmp\\AttachedContainer\\*.msi\").FullName | ForEach-Object { Expand-MsiArchive $_ \"$dir\" }",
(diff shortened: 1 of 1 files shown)


dwettstein:
Hey @r15ch13. Will that PR ever have a chance to be merged?

stevehipwell:
@dwettstein I'm keen to help you get this merged but I have a question. Isn't the `Scripts` folder managed as part of the package installation and how does making this persistent impact that?

dwettstein:
@stevehipwell, AFAIK, when adding these folders to `persist`, Scoop will save them to the `$SCOOP_HOME/persist/python` location instead of `$SCOOP_HOME/apps/python`. Thus they won't get removed or overwritten when updating Python to a newer version.

It should work similarly to [NodeJS](https://github.com/ScoopInstaller/Main/blob/master/bucket/nodejs.json).

stevehipwell:
@dwettstein I assume you've tested this locally, and by this I mean a simulated update?

rashil2000:
@stevehipwell I have tested this PR locally, with an update simulated too.

@dwettstein Could you remove these two entries from the `bin` field? Since we're already adding the root directory to user's PATH, they'll already have `python` and `pythonw` executables available. Creating shims for them would unnecessarily result in duplicated executables.
```
    "python.exe",
    "pythonw.exe",
```

dwettstein:
Thanks @rashil2000!

You're right, I will do that.

rashil2000 (dismissed):
LGTM!

dwettstein:
I have added a commit to install `py.exe` launcher as well. Input is from PR #2424 originally.

rashil2000 (approved):

stevehipwell (approved):
LGTM

PR #2173, opened 2021-05-14
Update openssh.json

tldzyx:
Fixed not exists files, update version with url and hash


pratikpc:
No zst support exists as of now

tldzyx:
> No zst support exists as of now

It's work in my PC with windows10. It's looks like support zst alreay?

pratikpc:
Silent extraction failure

Velgus:
Don't know what the best solution for this is for scoop's default buckets, but there is a [fork of 7-Zip that adds support for Zstandard](https://github.com/mcmilk/7-Zip-zstd/). 

If used, the fork works fine for the installation.

PR #2185, opened 2021-05-18
cmder: shortcut to start in home dir (`%USERPROFILE%`)

FriendlyNeighborhoodShane:
When launched from the start menu shortcut, cmder starts in
`%SCOOP%\apps\cmder\current`, which is not ideal behaviour for a
terminal program. This patch makes the shortcut always start
in the home directory with the argument '/start'.

When started from the shim or directly, cmder starts in your
current directory anyway, which is expected behaviour.

### DIFF
--- a/bucket/cmder-full.json
+++ b/bucket/cmder-full.json
@@ -14,7 +14,8 @@
     "shortcuts": [
         [
             "Cmder.exe",
-            "Cmder"
+            "Cmder",
+            "/start %USERPROFILE%"
         ]
     ],
     "persist": [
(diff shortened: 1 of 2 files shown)


FriendlyNeighborhoodShane:
Suggestion: maybe a dedicated field to control the "Start in" property of a shortcut should be added to the manifest? By default it's set to the directory of the executable `$SCOOP/apps/$app/current`.

I figure most GUI programs that care about a PWD will have an option like this anyway, but just in case they don't?

PR #2211, opened 2021-05-23
Add 6 apps

gaeulbyul:
> abs: ABS is a programming language that works best when you're scripting on your terminal.
> https://www.abs-lang.org/ ( https://github.com/abs-lang/abs )

> seonbi: Seonbi (선비) is an HTML preprocessor that makes typographic adjustments to an HTML so that the result uses accurate punctuations according to the modern Korean orthography.
> https://github.com/dahlia/seonbi

> fselect: Find files with SQL-like queries
> https://github.com/jhspetersson/fselect

> wormhole-william: End-to-end encrypted file transfer. A magic wormhole CLI and API in Go
> https://github.com/psanford/wormhole-william

> silicon: Create beautiful image of your source code.
> https://github.com/Aloxaf/silicon

> rage: A simple, secure and modern encryption tool (and Rust library) with small explicit keys, no config options, and UNIX-style composability.
> https://github.com/str4d/rage

### DIFF
--- a/bucket/abs.json
+++ b/bucket/abs.json
@@ -1,0 +1,30 @@
+{
+    "version": "2.5.0",
+    "description": "ABS is a programming language that works best when you're scripting on your terminal.",
+    "homepage": "https://www.abs-lang.org/",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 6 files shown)


rashil2000:
Hey, good effort but kindly open a different PR for each manifest. It's far easier to review stuff that way.

PR #2252, opened 2021-06-02
influxdb: Update to version 1.8.9, add influxdb2

Nefariis:
updating influxdb from 1.8.4 to 1.8.6


pratikpc:
Is this the same influxdb?
https://github.com/influxdata/influxdb/releases/tag/v2.0.7
They have reached 2.0.7

Nefariis:
They split off the builds, so 1.X development is continuing alongside 2.X development, 

I'm thinking there should probably be separate jsons for influx and influx2.

This would be the latest version for 1.X and I can add an influx2 json in about an hour.

I also don't know if I would put influx 1.X (this json) in the "previous versions" bucket because it is still under active development.

pratikpc:
How does Scoop handle this usually?

Is it influx2 and influx with 1 or influx with 2 and influx1.json separately?

Nefariis:
that is an excellent question that I don't have an answer for.

It could also be python formatting - influx1/influx2

Nefariis:
actually we probably should keep influx 1.x as plain "influx" and have influx 2.x be influx2. 

influx2 is a completely different system with breaking changes - and if we were to do influx1/influx2, it wouldn't update anyone that has "influx" already installed.

issaclin32 (approved):

bucket/influxdb2.json
```diff
@@ -0,0 +1,30 @@
...
+            "extract_dir": "influxdb2-2.0.8-windows-amd64"
+        }
+    },
+    "persist": "influxdb.conf",
+    "bin": [
+        "influx.exe",
+        "influxd.exe"
+    ],
+    "checkver": {
+        "url": "https://portal.influxdata.com/downloads",
```
pratikpc, line 19: Would it be a good idea here to use GitHub for checkver?

https://github.com/influxdata/influxdb/releases

pratikpc:
@Nefariis @issaclin32 I feel we also need to add Influx CLI
https://github.com/influxdata/influx-cli

Ash258:
Main bucket is only for the latest versions. Any other application does not have different versions using such ugly way in Main bucket and will not have.

1. Update influxdb manifes to version 2
2. Delete influxdb2 from this repository
3. Optionally open PR to Versions bucket with version 1
4. Fix persist of version 2. It cannot work with a folder instead of file

PR #2265, opened 2021-06-06
postgresql@13.3: update download URL

igitur:
Correct details:

"url": "https://get.enterprisedb.com/postgresql/postgresql-13.3-2-windows-x64-binaries.zip",
"hash": "b9027435bbfe1a10f25dac0689c09fe541f26a381c15aec37bc4636e09e537c3"

### DIFF
--- a/bucket/postgresql.json
+++ b/bucket/postgresql.json
@@ -10,8 +10,8 @@
     ],
     "architecture": {
         "64bit": {
-            "url": "https://get.enterprisedb.com/postgresql/postgresql-13.3-1-windows-x64-binaries.zip",
-            "hash": "cb1692d7a42cebb9d617bb51d9e8b940fad77e529e8714eec3e3baf26a73a227"
+            "url": "https://get.enterprisedb.com/postgresql/postgresql-13.3-2-windows-x64-binaries.zip",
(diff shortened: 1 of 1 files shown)


pratikpc:
Looks like checkver has an issue

Is there a way to update it?

igitur:
@pratikpc Where do you see the checkver issue?

pratikpc:
This is release 13.3-2 correct?

Now autoupdate is set to https://get.enterprisedb.com/postgresql/postgresql-$version-1-windows-x64-binaries.zip and the version detected is 13.3

Hence the selected URL is 13.3-1 due to the -1 present in the autoupdate URL.

If we could find a way to extract the 13.3-2 version instead of 13.3-1, we could download the correct file.

Is this possible?

igitur:
On the [download page](https://www.enterprisedb.com/download-postgresql-binaries), they publish it only as `13.3`, with not 3rd-level number, so checkver cannot detect anything more. That's why I believe a manual update is the only way to point to the new 13.3-2 files. I'm open to being corrected, though.

igitur:
@rasa Can you get this one through, please?

igitur:
This is superceded by https://github.com/ScoopInstaller/Main/commit/54fe638dd6c90d75fab0f74085c0d058648c76de

PR #2271, opened 2021-06-07
nvs: make it work in non-powershell shells

rashil2000:
The current method (before this PR) adds a shim which delegates nvs to the `nvs.ps1` script. 

NVS actually needs to modify the current shell's PATH on-the-fly, so calling the nvs shim from shells other than PowerShell (cmd.exe, Git Bash) does nothing. The solution is to add nvs.cmd, nvs (bash script) and nvs.ps1 all to the PATH so that they can be called directly. This is what the MSI installer of NVS does too.

### DIFF
--- a/bucket/nvs.json
+++ b/bucket/nvs.json
@@ -6,7 +6,7 @@
     "url": "https://github.com/jasongin/nvs/releases/download/v1.6.0/nvs-1.6.0.msi",
     "hash": "95b0fdff01142d5aa56217e32f79fe27c12ee9ac976a1c4e160b0fd6e8db320f",
     "extract_dir": "nvs",
-    "bin": "nvs.ps1",
+    "env_add_path": ".",
     "persist": "nodejs",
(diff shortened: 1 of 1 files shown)


rashil2000:
Hi! 

I would like to know what's the blocker for this simple one-line change; it's been open for almost a month now.

stevehipwell:
@rashil2000 why haven't you added the relevant files to `bin` as an array? This would be better than adding the whole folder to the PATH.

rashil2000:
Hey @stevehipwell, could you elaborate? 

Adding 3 separate entries for nvs.cmd, nvs.ps1, and nvs (bash script) would create conflicting shims. Note that for each type of shell, we need the corresponding script in PATH (shims won't work as they can't modify each shell session's environment on the fly).

stevehipwell:
@rashil2000 fair enough, that makes sense.

stevehipwell (approved):

Issue #2328, opened 2021-06-27
Why go@1.16 was rolled back to 1.15?


Bpazy:
Why `go@1.16` was rolled back to `1.15` in this commit: 5764396ba88010da6af9546c370272b24570e08d

I want to know why roll back, because I want to use this version (1.16).

HUMORCE:
checkver(regex) is not updated caused this issue.

Bpazy:
Thank you for your answer and PR

PR #2330, opened 2021-06-27
go: Update the regex for checkver

HUMORCE:
Fixes https://github.com/ScoopInstaller/Main/issues/2328


falco467 (approved):
simple and succint. Will fix the current problem.

HUMORCE:
the regex of checkver is updated.

PR #2424, opened 2021-07-09
Installs PEP-397, resolves #2181, resolves #377

### LINKED ISSUE #377: python: Installer needs to put Scripts dir into path
The python installer currently does not put the `Scripts` dir into the path. This means that installed scripts (or even `pip.exe` itself) aren't in the path.

To fix this, the following path needs to be added to the `%PATH%`:

`%HOMEPATH%\scoop\apps\python\current\Scripts`

### LINKED ISSUE #2181: Python Install - Missing py.exe
Hi Team,
Probably a duplicate of https://github.com/ScoopInstaller/Main/issues/377

I just installed python using scoop - python command works as expected but py.exe is nowhere to be found.
I did search through many links and issues and only found the one mentioned above relevant and close to my issue.
**Error message:**

`py.exe : The term 'py.exe' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.

At line:1 char:1
 py.exe -0p
     CategoryInfo          : ObjectNotFound: (py.exe:String) [], CommandNotFoundException
     FullyQualifiedErrorId : CommandNotFoundException`

Would be nice if we can include the launcher also as part of the install and the relevant environmental variable to go with it ?

JKorbelRA:
Not sure why the launcher is excluded.

Also not sure why install-pep-514.reg / uninstall-pep-514.reg aren't called automatically in post_install / uninstall for user's convenience.

### DIFF
--- a/bucket/python.json
+++ b/bucket/python.json
@@ -46,7 +46,7 @@
     "installer": {
         "script": [
             "Expand-DarkArchive \"$dir\\setup.exe\" \"$dir\\_tmp\"",
-            "@('launcher.msi', 'path.msi', 'pip.msi') | ForEach-Object {",
+            "@('path.msi', 'pip.msi') | ForEach-Object {",
             "    Remove-Item \"$dir\\_tmp\\AttachedContainer\\$_\"",
             "}",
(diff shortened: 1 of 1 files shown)


JKorbelRA:
resolves #2181, resolves #377

bilabar (approved):
I have tested it. It works!

PR #2427, opened 2021-07-09
xray: Update to version 1.4.3, Fix persist

MaoKwen:


issaclin32 (dismissed):

issaclin32:
I removed `autoupdate.hash` because **xray** does not provide digest (`.dgst`) in version 1.4.3.

issaclin32 (approved):

PR #2443, opened 2021-07-13
telegraf: Fix hash extract from download page

thesebas:
original regex matches some random hash from page (eg docker image hash)


thesebas:
Fixes issues like #2300 #2442 and so on...

thesebas:
@r15ch13 @Ash258 who can do review? #2555 another hash check failed after update

thesebas:
#2613 fails again

PR #2526, opened 2021-08-04
screenfetch: Update to v1.0

yangshunhuai:
Made a few modification to [JulianChow94's Windows-screenfetch](https://github.com/JulianChow94/Windows-screenFetch) and added it into the main bucket.

### DIFF
--- a/bucket/screenfetch.json
+++ b/bucket/screenfetch.json
@@ -1,0 +1,9 @@
+{
+    "version": "1.0",
+    "description": "A tool that can show you basic system info",
+    "homepage": "https://github.com/yangshunhuai/Windows-screenFetch",
+    "license": "Unknown",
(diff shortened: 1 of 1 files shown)


dwettstein:
Thanks for the PR. Why not contributing back to [JulianChow94's Windows-screenfetch](https://github.com/JulianChow94/Windows-screenFetch) repository? Additionally, the repository has no license.

Have a look at [Criteria-for-including-apps-in-the-main-bucket](https://github.com/lukesampson/scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) as well:
> For an app to be acceptable for the main bucket, it should be:
>
>    a reasonably well-known and widely used developer tool

I would suggest to use [Neofetch](https://github.com/dylanaraps/neofetch) instead.

PR #2553, opened 2021-08-12
Updated qemu bin mapping

VictorKoenders:
Updated the `"bin"` mapping to match the binaries in the qemu folder after installation


akirayamamoto:
Closes #2579

akirayamamoto (approved):
Tested and approved.

Issue #1247, opened 2020-07-05
Install legit without git as dependence.


gphg:
I have a python, but it's installed on MSYS2 and I don't want to have another or multiple Python installed on Windows. No `pip` for me. I don't want my user `%PATH%` populated either.

When I tried to install `legit`, it install `git` automatically. **But, I already has `mingit` installed**. I solve this by uninstall `git` after the installation. I had to `scoop reset mingit` to fix the shims too.

I suggest `git` should removed from `legit`'s dependency. You can add extra note later on; just like `mingit` with its `openssh`.

This is my assumption, I haven't tried it: will scoop installs `git` if I already had Git for Windows installed (not from Scoop, but by installer and system scope; eg, it's installed on `C:\Program Files`)? Please fix.

rasa:
@gphg Was the git executable in your path, when you installed legit? That might be a workaround for your issue.

gphg:
@rasa It was.

PR #2562, opened 2021-08-14
Updated "git" as depends on legit into suggest

gphg:
Resolved https://github.com/ScoopInstaller/Main/issues/1247

"Depends" installs the mentioned manifest. This could be trouble some for other that had get `git` installet out of scoop or other variants such as `mingit`.

### DIFF
--- a/bucket/legit.json
+++ b/bucket/legit.json
@@ -7,7 +7,10 @@
         "url": "https://github.com/frostming/legit/blob/master/LICENSE"
     },
     "suggest": {
-        "git"
+        "git": [
+            "git",
+            "git-with-openssh"
+        ]
     },
     "url": "https://github.com/frostming/legit/releases/download/1.2.0/legit.exe",
     "hash": "2c4aa8a777141cad9171a7aec9b91d6570443644f7ca040903773620074d361b",
--- a/bucket/legit.json
+++ b/bucket/legit.json
@@ -6,7 +6,9 @@
         "identifier": "Freeware",
         "url": "https://github.com/frostming/legit/blob/master/LICENSE"
     },
-    "depends": "git",
+    "suggest": {
+        "git"
+    },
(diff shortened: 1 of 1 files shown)


gphg:
Please forgive my typos. I done these fork, edit, commit, and PR from phone. 😞

bucket/legit.json
```diff
@@ -6,7 +6,9 @@
         "identifier": "Freeware",
         "url": "https://github.com/frostming/legit/blob/master/LICENSE"
     },
-    "depends": "git",
+    "suggest": {
+        "git"
```
pratikpc, line 10: Need to write suggest git like this
https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/git-town.json#L7

pratikpc, line 10: https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/git-town.json#L7
Need to write suggest like this for Git

pratikpc (changes requested):
Update Suggest to recommended approach
You can also look at https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/neofetch.json#L7
Here they recommend multiple packages and you can recommend git, mingit and git-with-openssh as well

bucket/legit.json
gphg: out look good.

gphg: by the way I'm not similar with this review on pr. what should I do?

dwettstein:
```suggestion
        "git": [
            "git",
            "git-with-openssh"
        ]
```

dwettstein (approved):
LGTM

PR #2563, opened 2021-08-14
radare2: Update to version 5.3.1

MarkKoz:
Download from GitHub instead of radare.mikelloc.com; Windows binaries have been release on GitHub since version 4.4.0. The previous site has an expired certificate anyway, which may be the reason why scoop has been unable to install radare2 using the current manifest.


MarkKoz:
Right... Before addressing that, I'm going to wait for the 5.4.2 release because it will fix a Windows distribution naming inconsistency introduced in 5.4.0.

PR #2579, opened 2021-08-21
qemu: Remove shims for removed binaries

tylerszabo:
QEMU 6.1 removed `moxie` <https://qemu-project.gitlab.io/qemu/about/removed-features.html> and the shims must be updated to avoid an error from missing target.


tylerszabo:
Tested with `scoop update -f qemu`

akirayamamoto:
Closed by #2553

PR #2595, opened 2021-08-25
Fix cygwin 64 sha512 hash

felipecrp:
Hi, I tried to install cygwin and found that 64bits hash is not matching. I updated the sha512.


akirayamamoto:
Closed by #2660

PR #2626, opened 2021-09-01
Added azcopy@10.11.0

ogerardin:


issaclin32:
You need to specify `license` and `description` for official buckets.
**license** can be found [here]: "https://github.com/Azure/azure-storage-azcopy/blob/main/LICENSE"
**description**: A short phrase/sentence to describe what the app does.

Also, please be kind. Add some brief information about **"what the application does"** in your PR, rather than throwing a link and let the maintainers to figure out.

bucket/azcopy.json
```diff
@@ -0,0 +1,17 @@
+{
+    "homepage": "https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10",
```
issaclin32, line 2: Put `version` in the first item, just like every other manifest (in this bucket) does.
```
    "version": "10.11.0",
    "description": "xxxxxxx",
    "homepage": "https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10",
    "license": "MIT",
```

bucket/azcopy.json
```diff
@@ -0,0 +1,17 @@
...
+        "64bit": {
+            "url": "https://azcopyvnext.azureedge.net/release20210616/azcopy_windows_amd64_10.11.0.zip",
+            "extract_dir": "azcopy_windows_amd64_10.11.0"
+        },
+        "32bit": {
+            "url": "https://azcopyvnext.azureedge.net/release20210616/azcopy_windows_386_10.11.0.zip",
+            "extract_dir": "azcopy_windows_386_10.11.0"
+        }
+    },
+    "bin": [
```
issaclin32, line 14: `bin` can be reduced to a single string if there is only 1 item.
`autoupdate` is not available, but it is possible to involve `checkver`. (and update manually when `checkver` tells us)
For details, check: https://github.com/lukesampson/scoop/wiki/App-Manifest-Autoupdate#add-autoupdate-to-a-manifest
Or, you can just check other packages in the bucket.
```
    "bin": "azcopy.exe",
    "checkver": {
        "github": "https://github.com/Azure/azure-storage-azcopy"
    },
```

PR #2644, opened 2021-09-05
Fix imagemagick download link and hash

sudiptochoudhury:
The download links seem to be not adhering to the correct version number. Found details from https://download.imagemagick.org/ImageMagick/download/binaries/


stevehipwell:
Fixed by https://github.com/ScoopInstaller/Main/pull/2676.

PR #2655, opened 2021-09-09
telegraf: Fix update and hash logic

stevehipwell:
The current Telegraf update logic is brittle and currently broken, this PR swaps the logic to use GitHub to detect releases and then a regex to extract the correct checksum. I have tested that this works correctly to automatically generate the last release from the previous one.

Fixes #2082.
Fixes #2191.
Fixes #2224.
Fixes #2286.
Fixes #2299.
Fixes #2441.
Fixes #2555.
Closes #2083.
Closes #2192.
Closes #2225.
Closes #2300.
Closes #2442.
Closes #2443.
Closes #2556.

PR #2660, opened 2021-09-10
cygwin: Fix and update hash logic

stevehipwell:
The current implementation of the auto update hash logic for cygwin is brittle, as can be attested by the number of open issues, so should be swapped out for a stable implementation.

Fixes 43 issues and closes 6 PRs.

### DIFF
--- a/bucket/cygwin.json
+++ b/bucket/cygwin.json
@@ -10,11 +10,11 @@
     "architecture": {
         "64bit": {
             "url": "https://cygwin.com/setup-x86_64.exe#/cygwin-setup.exe",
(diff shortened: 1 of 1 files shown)


stevehipwell:
@dwettstein could you take a look at this, it's similar in function to https://github.com/ScoopInstaller/Main/pull/2655 and should allow the clean-up of a number of issues/PRs.

stevehipwell:
@akirayamamoto could you take a look at this?

akirayamamoto (approved):
Installed and tested with `checkver` and `checkhashes`.

stevehipwell:
@akirayamamoto we should be able to clean-up the cygwin issues and PRs now.

dwettstein (commented):
Where did you got this hash? I can only find the file you removed, https://cygwin.com/sha512.sum.

stevehipwell:
@dwettstein if there isn't a reliable source for the hash with the release, which there isn't for cygwin as it's a single endpoint for a single hash that seems to lag behind releases, the updater will download and manually hash the files.

PR #2662, opened 2021-09-10
winfetch: Add v2.2.0

rashil2000:
https://github.com/kiedtl/winfetch

Highly informative and customizable system information tool, akin to neofetch

### DIFF
--- a/bucket/winfetch.json
+++ b/bucket/winfetch.json
@@ -1,0 +1,13 @@
+{
+    "version": "2.2.0",
+    "description": "A command-line system information utility for Windows",
+    "homepage": "https://github.com/kiedtl/winfetch",
+    "license": "MIT",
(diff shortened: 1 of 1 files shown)


Ash258:
https://github.com/lukesampson/scoop-extras/blob/master/bucket/winfetch.json

rashil2000:
@Ash258 would you mind explaining the reason for that?

Winfetch fits the Main bucket's criteria and is popular enough now. I was about to send a PR to remove it from extras.

rashil2000:
I'm pretty sure you're not gonna find the time to respond here with any reason. Which is fine. If that's the case, please don't find the time to outright dismiss a PR as 'invalid', that too without any explanation. 

I'll be opening another PR for the same, this time with a detailed reasoning for the same.

PR #2668, opened 2021-09-11
Update xray.json

MarksonHon:
Fix sha512 for Xray's releaseing a new version with same version number 1.4.3


stevehipwell:
Fixed by https://github.com/ScoopInstaller/Main/pull/2672.

PR #2667, opened 2021-09-11
winfetch: Add v2.2.0

rashil2000:
This manifest already exists in the extras bucket. It wasn't very robust and/or popular when this tool was created, but it is now. So I'm thinking of promoting it to the Main bucket. Once this PR gets merged, I'll make a PR to the extras bucket to have it removed from there to avoid any duplication.

Here are some additional reasons why winfetch would fit nicely in the Main bucket:
- No pre/post install scripts
- CLI tool, no GUI
- No dependencies
- No binary blobs/executables (the manifest downloads a single script from GitHub which anyone can read) - this eliminates risk of any malware
- Fairly popular (375 stars, 12 contributors, which, frankly is way more than [some](https://github.com/pviotti/sayit) [of](https://github.com/arsham/rainbow) [the](https://github.com/evansmurithi/cloak) [apps](https://github.com/dduan/tre) currently in Main bucket :))
- Is 3 years old (as of now)
- Fun fact: it is 200-250 times faster than neofetch (currently in Main bucket)

**Disclaimer**: I'm currently a maintainer of winfetch :)

A previous PR #2662 of mine was closed abruptly, which I understand as I didn't provide any justification. So I've tried to give as much details here.

PR #2670, opened 2021-09-11
solidity: Add version 0.8.7

rashil2000:
A manifest for this exists in the extras bucket. However, it is extremely out of date, the download URL gives 404, has no checkver and no autoupdate.

I've fixed those things and updated it to the latest version. I think this is another good fit for the Main bucket as:
- CLI only, no GUI
- Very popular (12000 stars, 415 contributors)
- No pre/post install script
- single binary

Once this gets merged, I'll send a PR to remove the manifest from extras.

### DIFF
--- a/bucket/solidity.json
+++ b/bucket/solidity.json
@@ -1,0 +1,23 @@
+{
+    "version": "0.8.7",
+    "description": "A statically typed, contract-oriented, high-level language for implementing smart contracts on the Ethereum platform.",
+    "homepage": "https://soliditylang.org",
+    "license": "GPL-3.0-only",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/ethereum/solidity/releases/download/v0.8.7/solc-windows.exe#/solc.exe",
+            "hash": "82db83111c6e2c892179486cb7050d85f1517bf851027607eb7f4e589e714bc5"
+        }
+    },
+    "bin": "solc.exe",
+    "checkver": {
+        "github": "https://github.com/ethereum/solidity"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/ethereum/solidity/releases/download/v$version/solc-windows.exe#/solc.exe"
+            }
+        }
+    }
+}


stevehipwell:
@rashil2000 have you tested this locally (I'm sure you have but I have to ask, maybe add if you've tested locally to your PRs)?

stevehipwell:
Like https://github.com/ScoopInstaller/Main/pull/2667 we'd probably need to know the implications of a switch from extras. Also is there a maintainer on extras who would be able merge related PRs?

rashil2000:
Yes, I have tested it (I actually use it for a course assignment, that's what propelled me to update the manifest). 

Extras bucket seems to be regularly maintained from what I can see (last PR was merged 3 days ago). I'll open the solidity-deprecation PR there myself.

stevehipwell:
@rashil2000 if something is moving, and it's agreed that it's OK to move apps, the extras deprecation should happen before it's added to main IMHO.

rashil2000:
Raised lukesampson/scoop-extras#6850

Ash258:
- Manifest was not tested properly
    - The binary provided has to be 64bit only, as could be easily identified by the size, manifest in extras, or few click/one command testing
- Why is it so hard to keep unified PR naming, which is maintained in all buckets across different maintainers??

From all the commits history it is clear that there are 3 main titles:

- `<manifest>: Add version <version>`
- `<manifest>: Update to version <version>`
- `<manifest>@<version>: Fix something version specific`

Commits like `Add renamer (#2466)` are completely useless for enduser as it does not indicate anything, only breaking the log and confusing endusers.
image

rashil2000:
Thank you for the detailed feedback. I will change the commit message from `solidity: add v0.8.7` to `solidity: Add version 0.8.7`.

However, from the binary size itself, how to discern whether it is 32bit or 64bit? (note that the manifest in extras is an outdated one (you won't even find those assets in the link), so I deliberately ignored the architecture there).

rashil2000:
@stevehipwell is there anything else that needs to be done here?

stevehipwell (approved):
LGTM

PR #2682, opened 2021-09-15
kubeseal: Add version 0.16.0

maksymgendin:
> **Problem**: "I can manage all my K8s config in git, except Secrets."
> 
> **Solution**: Encrypt your Secret into a SealedSecret, which is safe to store - even to a public repository. The SealedSecret can be decrypted only by the controller running in the target cluster and nobody else (not even the original author) is able to obtain the original Secret from the SealedSecret.
> 
> The `kubeseal` utility uses asymmetric crypto to encrypt secrets that only the Kubernetes controller can decrypt.

Source https://github.com/bitnami-labs/sealed-secrets

### DIFF
--- a/bucket/kubeseal.json
+++ b/bucket/kubeseal.json
@@ -1,6 +1,6 @@
 {
     "version": "0.16.0",
-    "description": "The kubeseal utility uses asymmetric crypto to encrypt secrets that only the Kubernetes controller can decrypt.",
+    "description": "With kubeseal you can encrypt your Kubernetes Secret into a SealedSecret, which is safe to store - even to a public repository. The SealedSecret can be decrypted only by the controller running in the target cluster and nobody else (not even the original author) is able to obtain the original Secret from the SealedSecret.",
     "homepage": "https://github.com/bitnami-labs/sealed-secrets",
     "license": "Apache-2.0",
     "architecture": {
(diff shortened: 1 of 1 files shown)


bucket/kubeseal.json
```diff
@@ -0,0 +1,23 @@
...
+    "homepage": "https://github.com/bitnami-labs/sealed-secrets",
+    "license": "Apache-2.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.16.0/kubeseal.exe",
+            "hash": "0fb3ff9552530524ee7eae04161cba089a68ec2b0b25d9f43b1c1024452017aa"
+        }
+    },
+    "bin": "kubeseal.exe",
+    "checkver": {
```
issaclin32, line 13:
```suggestion
    "checkver": "github",
```

bucket/kubeseal.json
```diff
@@ -0,0 +1,23 @@
...
+        "github": "https://github.com/bitnami-labs/sealed-secrets"
```
issaclin32, line 14:
```suggestion

```

bucket/kubeseal.json
```diff
@@ -0,0 +1,23 @@
...
+    },
```
issaclin32, line 15:
```suggestion

```

maksymgendin:
Thanks @issaclin32 🙏

issaclin32 (approved):

PR #2685, opened 2021-09-16
neovim-nightly: Add version 0.6.0-dev-276-g685cf3981

rashil2000:
Daily builds of Neovim

Inspired by https://github.com/ScoopInstaller/Main/blob/master/bucket/vim-nightly.json

### DIFF
--- a/bucket/neovim-nightly.json
+++ b/bucket/neovim-nightly.json
@@ -32,8 +32,7 @@
     ],
     "checkver": {
         "url": "https://api.github.com/repos/neovim/neovim/releases",
-        "regex": "NVIM v([\\w-.]+)+([\\w-.])",
-        "replace": "${1}-${2}"
+        "regex": "NVIM v([\\w-.+]+)"
     },
     "autoupdate": {
         "architecture": {
--- a/bucket/neovim-nightly.json
+++ b/bucket/neovim-nightly.json
@@ -1,0 +1,48 @@
+{
+    "version": "0.6.0-dev+276-g685cf3981",
+    "description": "Vim fork focused on extensibility and usability",
+    "homepage": "https://neovim.io/",
+    "license": {
+        "identifier": "Apache-2.0,Vim",
+        "url": "https://github.com/neovim/neovim/blob/master/LICENSE"
+    },
+    "suggest": {
+        "vcredist": "extras/vcredist2015"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip",
+            "hash": "77e039d1be4e39839e6cd47c8a9ed586a5b708e369b1380abcfc9632e1db138d"
+        },
+        "32bit": {
+            "url": "https://github.com/neovim/neovim/releases/download/nightly/nvim-win32.zip",
+            "hash": "788d2ac118507f0cd8956e86bd6efdcba0032e359330d7f314caf93a441d0f3c"
+        }
+    },
+    "extract_dir": "Neovim",
+    "bin": [
+        "bin\\nvim.exe",
+        "bin\\nvim-qt.exe"
+    ],
+    "shortcuts": [
+        [
+            "bin\\nvim-qt.exe",
+            "Neovim"
+        ]
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


rashil2000:
@stevehipwell looks like the AppVeyor CI failed because apparently the version string is too complex here. Should I change it to just 'nightly' instead (I looked at some other buckets in Main, some of them do have 'nightly' as their versions).

Also, I opened a bunch of other PRs, would love to get them reviewed :)

Thanks!

stevehipwell:
@rashil2000 I don't have any experience with this issue so I'd suggest looking for prior art already existing in this bucket.

RE you other PRs, I'm off on annual leave tomorrow so don't have a lot of time for OSS work; how about assigning one of the other new maintainers to the PRs?

rashil2000:
Thanks, I'll look into it!

Actually, it looks like I don't have the ability to assign reviewers. How should I go about it?

stevehipwell:
Probably just mention them in the comments or set them as assignees if you can do that?

rashil2000:
Same for assignees, can't do that either.

I'll try mentioning them.

rashil2000:
@issaclin32 could you give some insights into this? The version string here isn't valid it seems.

issaclin32:
*Neovim* does not release nightly version anymore. Not sure about the reason, but you `checkver` will not work.

[release page](https://github.com/neovim/neovim/releases)

[Past release page fetched by Google](https://webcache.googleusercontent.com/search?q=cache:dpX_XAm2nBMJ:https://github.com/neovim/neovim/releases+&cd=2&hl=zh-TW&ct=clnk&gl=tw)


For the **version string**, I suggest replacing `+` with `-` by using two match groups,
such as
```
"regex": "NVIM v([\\w-.]+)+([\\w-.])",
"replace": "${1}-${2}"
```

rashil2000:
Neovim does publish nightly builds, but on some days, the build fails (and so the artifact doesn't appear). See the related issue here - https://github.com/neovim/neovim/issues/15707#issuecomment-922254155

There is a third-party bucket here https://github.com/naderi/scoop-bucket/commits/master/bucket/neovim-nightly.json where I previously submitted a manifest for neovim-nightly. You can see from the commit history that a build gets generated almost daily, with rare exceptions. And the regex pattern is the same as the one in this PR.

issaclin32:
> You can see from the commit history that a build gets generated almost daily, with rare exceptions. And the regex pattern is the same as the one in this PR.

Thanks for the information.

The better way is to fix the problem **upstream**. (by modifying *Scoop*'s main program)
I will send a PR later to fix this.
However, this may take a long time since *Scoop*'s maintainers (who has access to the code) is not very active now.

I would recommend you to just replace `+` with `-` as a temporary solution. So that we can add this package sooner.

[Main bucket's AppVeyor.yml](https://github.com/lukesampson/scoop/blob/3d67b7d37c4a2714e8942724d4677dcbd3678f01/schema.json)
[Code related to the problem](https://github.com/lukesampson/scoop/blob/master/schema.json#L532)

rashil2000:
@issaclin32 it failed again...

issaclin32:
> @issaclin32 it failed again...

Your force push reverted the code to [2c34130](https://github.com/ScoopInstaller/Main/commit/2c34130071f1dc0db0d1e39b8de3e523ed470e13). Therefore I cannot check what is going wrong from the AppVeyor logs.

Please update again to [685df7c](https://github.com/ScoopInstaller/Main/commit/685df7c6162f7a365f8733a3a69762ce48b40255), so that we can figure out what is causing the problem.

rashil2000:
I force-pushed to keep the bucket's commit history clean. Since the history is now rewritten, I can't revert to it. I can make a different PR with the previous code though, or make another commit here. Should I do that?

issaclin32:
> I can make a different PR with the previous code though, or make another commit here.

Just make another commit here.

rashil2000:
@issaclin32 reverted.

issaclin32 (approved):

issaclin32:
I forgot that the symbol `+` needs to be escaped too. It is now working as expected.

rashil2000:
Thank you for the help @issaclin32 !

PR #2687, opened 2021-09-17
Add alibaba arthas

cdgeass:


bucket/arthas.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "3.5.4",
+    "description": "Arthas is a Java Diagnostic tool open sourced by Alibaba.",
+    "homepage": "https://github.com/alibaba/arthas",
+    "license": "Apache-2.0",
+    "url": "https://github.com/alibaba/arthas/releases/download/arthas-all-3.5.4/arthas-bin.zip",
+    "hash": "0763be7affda0b7528c91299e1131e38d1ed75a23812ccf5915c708be451d989",
+    "bin": "arthas-boot.jar",
+    "checkver": {
+        "url": "https://api.github.com/repos/alibaba/arthas/releases/latest",
```
pratikpc, line 10: Any reason we cannot use checkver github?
It does not support regex?

pratikpc: Based on what I can find, support for regex exists

Check metastore

https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/metastore.json#L17

pratikpc (changes requested):
Explore if checkver github instead of checkver url and jsonpath is possible

Refer to https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/metastore.json#L17

PR #2702, opened 2021-09-22
pwsh: persist global profiles

janicmikes:
PowerShell uses two installation wide profiles.

You can check them with:
`$PROFILE | Format-List * -Force`

And read about them here: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.1#the-profile-files

```
AllUsersAllHosts:  $dir\Profile.ps1
AllUsersCurrentHost:   $dir\Microsoft.PowerShell_profile.ps1
... two user profiles ...
```

Those two files should be persisted.

### DIFF
--- a/bucket/pwsh.json
+++ b/bucket/pwsh.json
@@ -13,12 +13,24 @@
             "hash": "b44702f129514e638798d62a6f3eacb62ef8a628052f71f1ceed179ede3d4564"
         }
     },
+    "pre_install": [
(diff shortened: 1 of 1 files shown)


janicmikes:
I update the commit to be somewhat similar to what "Test-Persistence" of Shovel does.
With the let's call it "Good Old" Test-File instead until we get a helper like this in Scoop.

issaclin32 (approved):

PR #1469, opened 2020-10-04
azure-cli: Set AzureCLIPath environment variable

alastairs:
As identified in, e.g., [the Azure Go SDK][1] and [other Azure SDKs][2], the Azure CLI program can be resolved using an environment variable called AzureCLIPath or AZURE_CLI_PATH if the program is not installed to the default locations. Setting this environment variable allows programs which build on the Azure SDKs to resolve the CLI program when it has been installed with Scoop.

  [1]: https://github.com/Azure/go-autorest/blob/afe17657e2844fc0def7ffbf572d7d869ffedb4d/autorest/azure/cli/token.go#L151
  [2]: https://github.com/search?q=org%3AAzure+AzureCLIPath&type=code

PR #2717, opened 2021-09-29
azure-cli: Fix AzureCLIPath environment variable

aadje:
The 2 AzureCLIPath environment variables introduced for azure-cli in https://github.com/ScoopInstaller/Main/pull/1469 are not set to the actual az cli path. 
az is installed in the \wbin sub folder, as seen in the bin property. 

This for example resulted in tools using the go-autorest library not working when the az cli is installed using scoop. 
https://github.com/Azure/go-autorest/blob/c7f947c0610de1bc279f76e6d453353f95cd1bfa/autorest/azure/cli/token.go#L128

### DIFF
--- a/bucket/azure-cli.json
+++ b/bucket/azure-cli.json
@@ -15,7 +15,7 @@
         "url": "https://azurecliprod.azureedge.net/msi/azure-cli-$version.msi"
     },
     "env_set": {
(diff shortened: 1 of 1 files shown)


rashil2000:
@stevehipwell is this pending testing by you?

PR #2741, opened 2021-10-07
gcc: Update to version 10.3.0-5

silverkorn:
## Warning! 🚨 
This update needs `zstd` decompression support.

Since most of the non-_.zst_ files are gone on MSYS2's official repository for version `9.3.0-2`, you can judge and evaluate if it's preferable to cancel this PR or merge it and being restricted to use alternative fork, which supports `zstd`.


silverkorn:
Ok, I totally see your point on this. Noted!

The main concern still, unless there's an integration of zstd decompression soon directly in the official scoop repository or anyone providing a an **official trusted** mirror using the supported compression format, **this application manifest (`gcc.json`) is now irrelevant and should be removed** since most of it's dependencies in the supported format are gone and lead to errors and incomplete/broken installation.

Many people are puzzled and concerned on the broken installation of GCC through scoop, as why the previous links to refer to all of them.

Thanks!

Issue #2742, opened 2021-10-08
kompose: Download failed


Divkix:
Error:
```
The request was aborted: The connection was closed unexpectedly.
URL https://github.com/kubernetes/kompose/releases/download/v1.24.0/kompose-windows-amd64#/kompose.exe is not valid
```

This maybe happening due to change in name of kompose binary.

The new name is: `kompose-windows-amd64.exe`

The Download URL should be changed with this:
`https://github.com/kubernetes/kompose/releases/download/v1.24.0/kompose-windows-amd64.exe`

I'll open a pull request regarding this

PR #2743, opened 2021-10-08
kompose: Fix download URL

Divkix:
This fixes the download URL for `kompose`

Issue: #2742


Divkix:
Hey @stevehipwell the installation is now working, thanks for pointing out the rename part, didn't think of that.

Issue #2690, opened 2021-09-18
Request for adding k3sup


paalkr:
[k3sup ](https://github.com/alexellis/k3sup) is a popular tool for deploying the [Rancher k3s](https://rancher.com/docs/k3s/latest/en/) Kubernetes distribution. Having k3sup available in scoop would make deployment of Kubernetes clusters even easier.

PR #2766, opened 2021-10-15
k3sup: Add version 0.11.0

wenmin92:
close https://github.com/ScoopInstaller/Main/issues/2690

### DIFF
--- a/bucket/k3sup.json
+++ b/bucket/k3sup.json
@@ -1,6 +1,6 @@
 {
     "version": "0.11.0",
-    "description": "A light-weight utility to get from zero to KUBECONFIG with k3s on any local or remote VM",
+    "description": "Utility to get from zero to KUBECONFIG with k3s on any local or remote VM",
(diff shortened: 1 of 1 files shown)


wenmin92:
I didn't use this tool either, simply because I happened to see the [request](https://github.com/ScoopInstaller/Main/issues/2690), so I simply submitted it. My original idea was to have someone use it and submit an issue, and then improve it accordingly. I'm sorry for this hasty action.

Maybe we should consult with people who have used it (@paalkr)

issaclin32:
> My original idea was to have someone use it and submit an issue, and then improve it accordingly. I'm sorry for this hasty action.

No worries. I do not think the manifest needs to be "perfect" either.
I just want to avoid some crucial problems that might affect users who uses and updates the package using *Scoop*.

issaclin32:
The config file is located at `%userprofile%\.k3sup`. Therefore *persist* is not needed.

PR #2773, opened 2021-10-18
Fix missing hyperfine.exe

jcwillox:
In the latest release of [hyperfine](https://github.com/sharkdp/hyperfine/releases), the binary was moved into a subdirectory `hyperfine-v$version-x86_64-pc-windows-msvc`, this PR adds `extract_dir` to fix that.


rashil2000:
Tested locally, works. 

Couple of nits:
- hyperfine's CD was recently updated to include 32-bit binaries as well, it'd be great if you could include them (see fd's manifest for reference)
- Kindly follow the standard commit message format: `hyperfine@1.12.0: Specify extract_dir entry` or `hyperfine@1.12.0: Include 32bit binary`

PR #2774, opened 2021-10-18
nexus-repository-oss: Add version 3.35.0

wenmin92:


wenmin92:
Thanks @Valinor @pratikpc , the manifest file has fixed.

PR #2779, opened 2021-10-18
7-zip: script to install explorer context menu

faabiosr:
Setup the 7zip context menu reg files.

### DIFF
--- a/bucket/7zip.json
+++ b/bucket/7zip.json
@@ -6,16 +6,44 @@
         "identifier": "Freeware,LGPL-2.0-only,BSD-3-Clause",
         "url": "https://www.7-zip.org/license.txt"
     },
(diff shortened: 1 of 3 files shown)


idbrii:
Some fallout I experienced from #846 was that the scripts people were using to update the installer version broke after changing the number of urls it downloaded. You may want to track down who updates the 7zip version and ping them here.

Nice work!

issaclin32:
I would suggest following the pattern of `VSCode` and `GrepWin`.

That is, renaming `install-7zip.reg` and `uninstall-7zip.reg` to **7zip-install-context.reg** and **7zip-uninstall-context.reg**.

https://github.com/lukesampson/scoop-extras/tree/master/scripts

PR #2784, opened 2021-10-20
latex: Update to version 21.8

### LINKED ISSUE #2783: latex: shortcuts creation error
While installing `latex` the following error occured:

```
Linking ~\scoop\apps\latex\current => ~\scoop\apps\latex\21.6
Creating shim for 'miktex'.
Creating shortcut for texmfs\install\miktex\bin\x64\miktex-console.exe (MiKTeX Console) failed: Couldn't find C:\Users\vadim\scoop\apps\latex\current\MiKTeX Console
Persisting texmfs\config
'latex' (21.6) was installed successfully!
```
I've found out that two strings in the ` shortcuts` sections are misplaced and should be swapped.

Instead
``` json
"shortcuts": [
    [
        "MiKTeX Console",
        "texmfs\\install\\miktex\\bin\\miktex-console.exe",
    ]
],
```
they should be
``` json
"shortcuts": [
    [
        "texmfs\\install\\miktex\\bin\\miktex-console.exe",
        "MiKTeX Console"
    ]
],
```

issaclin32:
closes #2783

- Fixes the shortcut error mentioned in #2783
- **MikTeX** does not provide 32-bit binaries anymore, causing `autoupdate` to fail. This fixes the problem

### DIFF
--- a/bucket/latex.json
+++ b/bucket/latex.json
@@ -1,6 +1,6 @@
 {
-    "version": "21.6",
-    "description": "MikTeX is an up-to-date implementation of TeX/LaTeX and related programs.",
+    "version": "21.8",
+    "description": "MikTeX, an up-to-date implementation of TeX/LaTeX and related programs.",
     "homepage": "https://miktex.org",
     "license": {
         "identifier": "LPPL1.3c,GPL-3.0-or-later,Public Domain,...",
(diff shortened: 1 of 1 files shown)


rashil2000:
Hey, no worries at all, just ping me when needed.

Since from 21.8 they don't provide a 32bit version, would it make sense to move the `bin`, `shortcuts` and `env_add_path` fields outside of the architecture block?

issaclin32:
> would it make sense to move the `bin`, `shortcuts` and `env_add_path` fields outside of the architecture block?

This will make the manifest look cleaner,
but it looks weird when there are `bin\\x64` inside the binary path.

Another way is to move all the items out of `architecture -> 64-bit` block, but this will make *Scoop* unable to stop the installation on a 32bit PC.

I personally think we can remove all 32-bit support, since there are few 32-bit PCs in use.
However, we should keep the pattern before further discussions and dicisions (from all the maintainers)

PR #2790, opened 2021-10-21
haskell: update to 9.0.1

pratikpc:
Based on @glatavento's [commit](https://github.com/ScoopInstaller/Main/pull/1792)

Differences common to both from existing on Main:-
1. Can autoupdate to 8.10.7
2. Update env_add_path
3. Remove WinGHCi which is no longer shipped
4. Switch to GitHub for checkver

Differences between current @glatavento's commit
1. Switch to GitHub for downloading the binaries themselves
2. No longer is the pre_install script required (handled with extract_dir)
3. Update to 9.0.1
4. Use GitHub API for checkver to always ensure 9.0.1 is received
5. Remove runhaskell which is absent in 9.0.1 from shims

I hope @glatavento can update their commit based on this or we can merge this.

### DIFF
--- a/bucket/haskell.json
+++ b/bucket/haskell.json
@@ -1,5 +1,5 @@
 {
-    "version": "8.10.7",
+    "version": "9.0.1",
     "description": "An advanced, purely functional programming language.",
     "homepage": "https://www.haskell.org",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


glatavento:
Hello @pratikpc! Thanks for this commit!
I'm really glad and surprised to see someone made a newer manifest for haskell, but actually I'm using ghcup now. I would delete my commit because it won't be actively maintained and yours is obviously a better choice.

PR #2847, opened 2021-10-29
Autoupdate manifest 2845

Valinor:
fix #2845

update and add autoupdate to some manifest :

apimtemplate: 0.7 (scoop version is 0.5)
dhall: 1.40.1 (scoop version is 1.39.0)
dnscrypt-proxy: 2.1.1 (scoop version is 2.1.0)
fio: 3.27 (scoop version is 3.20)
k0s: 1.22.2 (scoop version is 0.13.1)
megatools: 1.10.3 (scoop version is 1.10.2)
mockery: 2.9.4 (scoop version is 2.8.0)
pngquant: 2.16.0 (scoop version is 2.15.1)
rainbow: 1.1.0 (scoop version is 1.0.1)
uutils-coreutils: 0.0.8 (scoop version is 0.0.6)

PR #2869, opened 2021-11-01
Update fio to v3.27 

Valinor:
- Update v3.27
- Add autoupdate

### DIFF
--- a/bucket/fio.json
+++ b/bucket/fio.json
@@ -1,20 +1,32 @@
 {
-    "version": "3.20",
+    "version": "3.27",
     "description": "Flexible I/O Tester",
     "homepage": "https://bsdio.com/fio/",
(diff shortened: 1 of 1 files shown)


Valinor:
I need some help for this one. Don't understand the error : 

image

Ok, so I check https://github.com/Ash258/Scoop-GithubActions/blob/main/Entrypoint.ps1 : 

image

This is ok so next : 
image

Hum, never see Hash done in log... So : repeat the command manually : 

image

Ok for me....

See diff to be sure :  

image
 

Really, I don't get it...

rashil2000:
See my comment here https://github.com/ScoopInstaller/Main/pull/2847#issuecomment-955915603

Valinor:
No more important, see #2847

PR #2870, opened 2021-11-01
Manifests update 11 01 21

Valinor:
**Update manifest for** 
* clink.json
* concourse-fly.json
* gawk.json
* just.json
* jx.json
* k3d.json
* lessmsi.json
* macchina.json
* mdcat.json
* neovim-nightly.json
* oh-my-posh3.json
* pandoc.json
* rink.json
* rust-analyzer.json
* s3deploy.json
* tflint.json
* vim-nightly.json
* wasmtime.json
* xmake.json
* yq.json

Tested (install / uninstall )

**Not updated :** 
* terraform-provider-ibm.json

Dependancy to version bucket. Not tested

### DIFF
--- a/bucket/clink.json
+++ b/bucket/clink.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.2.40",
+    "version": "1.2.41",
     "description": "Powerful Bash-style command line editing for cmd.exe",
     "homepage": "https://chrisant996.github.io/clink/",
     "license": "GPL-3.0-only",
@@ -7,8 +7,8 @@
         "Run 'clink inject' to start clink on the current cmd",
         "Run 'clink autorun install' to auto start clink"
     ],
-    "url": "https://github.com/chrisant996/clink/releases/download/v1.2.40/clink.1.2.40.a4d3a0.zip",
-    "hash": "73b4b755f30b2881faf470f95a171609165652b984a761b3ac4e70e55b51e20f",
+    "url": "https://github.com/chrisant996/clink/releases/download/v1.2.41/clink.1.2.41.647c1b.zip",
+    "hash": "bab0ecb6a754d3b3411626d59f755e2cb47656c06e0b21f4558659591ba113a8",
     "bin": "clink.bat",
     "persist": "profile",
     "checkver": {
--- a/bucket/concourse-fly.json
+++ b/bucket/concourse-fly.json
@@ -1,12 +1,12 @@
 {
-    "version": "7.6.0",
+    "version": "7.4.2",
     "description": "Concourse CI/CD manager and controller",
     "homepage": "https://concourse-ci.org/fly.html",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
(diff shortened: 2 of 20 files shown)


rashil2000:
Would we have to this manually now?

Valinor:
> Would we have to this manually now?

I don't know, I have a automate process which work each day to check some stuffs (not update, this is a side effect). And because I see some latency with some packages, I propose this temporary push.

Exemple : https://github.com/JanDeDobbeleer/oh-my-posh/releases : 
- scoop version is 5.16.2
- current version is v5.16.4 (2 day ago)

But maybe it's not usefull, and we can discard this push. I will be happy with that

pratikpc (approved):

pratikpc:
Approved from my end

Does neovim-nightly even need a autoupdate block given the URL does not change anyways?

rashil2000:
> Approved from my end
> 
> Does neovim-nightly even need a autoupdate block given the URL does not change anyways?

You're right, it doesn't need it

pratikpc:
@Valinor gawk is the same TLS issue as well

```
INFO: Starting bucket/gawk.json checks

Manifest: /github/workspace/bucket/gawk.json

INFO: Hashes

The SSL connection could not be established, see inner exception.
At /SCOOP/apps/scoop/current/lib/install.ps1:130 char:9
+         throw $e
```

rashil2000:
I think the Sourceforge website uses a version of TLS that's not compatible with Scoop (in Actions)

stevehipwell:
Was this process always manual? If not there might be some thoughts required around this as the expected Scoop behaviour has changed. At a minimum I think something is required to split failing packages into a separate PR so the good updates can be merged ASAP.

rasa (approved):

rasa:
> > Does neovim-nightly even need a autoupdate block given the URL does not change anyways?
> 
> You're right, it doesn't need it

@pratikpc @stevehipwell So the hash will get updated even though there is no `autoupdate` section?

stevehipwell:
> @pratikpc @stevehipwell So the hash will get updated even though there is no autoupdate section?

I'm not sure, it may well do if there isn't a URL to template.

rashil2000:
@rasa It looks like Excavator hasn't run on the repo for two days, we might have a problem here.

r15ch13:
@rashil2000 Excavator is still running, but it constantly runs into Github's API call limiter.

rashil2000:
So can we use @Scoopbot account to split the load? What alternative do you suggest?

rashil2000:
Or maybe we can use a scheduled GitHub Action (like https://github.com/rasa/scoop-directory does) to check for updates every 3 hours or so, and commit them?

r15ch13:
> So can we use @Scoopbot account to split the load? What alternative do you suggest?

Nope, it's an IP limitation. The amount of checked manifests has to be split up. But that doesn't work with the current codebase.

rashil2000:
Yes, that why I was suggesting GitHub Actions. If there's a limit of say 100 (I'm not aware of the actual limit), then we can schedule the Actions to run 8 times per day (we have around 800 manifests total). Is such a thing possible?

rasa:
I added an Excavator workflow in #2871

PR #2878, opened 2021-11-03
fix(script): use scoop prefix to correctly find scoop's path

niheaven:


Valinor:
image

Ok, it's better to use true PowerShell command, but performance is bad. 15% minimum worst.

niheaven:
Ten milliseconds lost is acceptable, right?

pratikpc:
@Valinor to me it looks like that's down to scoop-prefix being slow

I took a look at the code.

It loads all 4 first, when it doesn't always need all 4 of them and doesn't outright use manifest and buckets

```
. "$psscriptroot\..\lib\core.ps1"
. "$psscriptroot\..\lib\help.ps1"
. "$psscriptroot\..\lib\manifest.ps1"
. "$psscriptroot\..\lib\buckets.ps1"
```

After a few changes in the scoop-prefix.ps1, I was able to get

(Measure-Command {..\bin\scoop.ps1 prefix scoop}).TotalMilliseconds
61.1987
(Measure-Command {resolve-path (split-path (split-path (..\bin\scoop.ps1 which scoop)))}).TotalMilliseconds
72.7342

Will create a PR later.

niheaven:
@pratikpc Just commit your fix to this one

pratikpc:
@niheaven @Valinor created a PR

https://github.com/ScoopInstaller/Scoop/pull/4493

Valinor (approved):

PR #2880, opened 2021-11-03
pngquant: update to v2.16.0

pratikpc:
add autoupdate

### DIFF
--- a/bucket/pngquant.json
+++ b/bucket/pngquant.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.15.1",
+    "version": "2.16.0",
     "description": "A command-line utility for lossy compression of PNG images",
     "homepage": "https://pngquant.org/",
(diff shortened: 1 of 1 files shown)


rashil2000:
The URL isn't changing, why is autoupdate required?

pratikpc:
Check with checkver.ps1 -Dir $env:SCOOP\buckets\main\bucket -App pngquant -Update

Without autoupdate present, it does not update the hash or version at the moment.

pratikpc:
We are facing the same issue here as well

> The SSL connection could not be established, see inner exception.

rashil2000:
This is something that needs to be fixed in Scoop core, as mentioned in my comment the other day

Valinor:
I already do this work with #2847 Autoupdate manifest 2845, with lot of more update.
Why this PR? Should I remove #2847 ?

Issue #2733, opened 2021-10-06
imagemagick: consider fetching the binaries from a mirror


brechtm:
I problem I'm constantly running into with ImageMagick is that old portable releases are being removed from their website when a new version is released. This makes it impossible to install a specific old release (to prevent unexpected regressions and optimize caching in my GitHub workflow).

I noticed that the [download mirrors](https://imagemagick.org/script/mirror.php) listed on the ImageMagick website do host the old portable releases. Would it be acceptable to use one of these mirrors instead of the official server?

brechtm:
Maybe I was confused when I created this issue (looking at the _releases_ directory instead?), but I can no longer see the older releases in the _binaries_ directory now. So I'm afraid the new fetch location won't make a difference 😢

Valinor:
Hum, there is the release directory, but the main problem (and the true problem for scoop) is that portable version don't have any subversion.
ImageMagick-7.1.0-X-Q16-x86-static.exe with X the subsubversion
but for portable we only have

ImageMagick-7.1.0-portable-Q8-x86.zip. So they upload a new portable with the same name each time they publish a new subversion. And I currently not found "old" portable version. Sorry

niheaven:
@Valinor The true problem is ImageMagick's hash and checksum are generated about few hours later than put into download folder. I'll change `checkver` to solve this.

PR #2885, opened 2021-11-04
ImageMagick - includes DLLs for programming APIs

tech189:
The manifest now downloads the full installer and unpacks it rather than the portable version. It also adds the environment variables and the installation folder to the path. This allows other programs such as Python Wand to find the programs and libraries for their own use. If further environment variables are missing, they can be added from this list: https://imagemagick.org/script/resources.php#environment. The bundled ffmpeg was unused in the portable version but as the manifest now adds the installation directory to the path, the outdated ffmpeg is renamed so that users can continue using the up-to-date ffmpeg downloaded by scoop.

### DIFF
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -21,7 +21,6 @@
         "compare.exe",
         "composite.exe",
         "conjure.exe",
-        "convert.exe",
         [
             "convert.exe",
             "imconvert"
@@ -42,7 +41,7 @@
     "post_install": [
         "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old"
     ],
-    "notes": "The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version",
+    "notes": "- The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version.\n- 'convert.exe' has been renamed to 'imconvert.exe' so that it does not conflict with the built-in command.\n",
     "checkver": {
         "url": "https://download.imagemagick.org/ImageMagick/download/binaries/?C=N;O=D",
         "regex": "ImageMagick-([\\d.-]+)-Q16-HDRI-x64-dll\\.exe\\.asc"
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -3,9 +3,6 @@
     "description": "Create, edit, compose, and convert 200+ of bitmap images formats.",
     "homepage": "https://imagemagick.org/",
     "license": "ImageMagick",
-    "suggest": {
-        "ffmpeg": "ffmpeg"
-    },
     "architecture": {
         "64bit": {
             "url": "https://www.imagemagick.org/download/binaries/ImageMagick-7.1.0-13-Q16-HDRI-x64-dll.exe",
@@ -26,7 +23,7 @@
         "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old",
         "Rename-Item $dir\\convert.exe $dir\\imconvert.exe"
     ],
-    "notes": "- The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version.\n- 'convert.exe' has been renamed to 'imconvert.exe' so that it does not conflict with the built-in command.\n",
+    "notes": "- The bundled ffmpeg has been renamed to 'ffmpeg.exe.old' to prevent conflict with a standalone ffmpeg installation - if you need it for ImageMagick tools, rename it back to 'ffmpeg.exe'.\n- 'convert.exe' has been renamed to 'imconvert.exe' so that it does not conflict with the built-in command.",
     "checkver": {
         "url": "https://download.imagemagick.org/ImageMagick/download/binaries/?C=N;O=D",
         "regex": "ImageMagick-([\\d.-]+)-Q16-HDRI-x64-dll\\.exe\\.asc"
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -3,7 +3,9 @@
     "description": "Create, edit, compose, and convert 200+ of bitmap images formats.",
     "homepage": "https://imagemagick.org/",
     "license": "ImageMagick",
-    "depends": "ffmpeg",
+    "suggest": {
+        "ffmpeg": "ffmpeg"
+    },
     "architecture": {
         "64bit": {
             "url": "https://www.imagemagick.org/download/binaries/ImageMagick-7.1.0-13-Q16-HDRI-x64-dll.exe",
@@ -40,7 +42,7 @@
     "post_install": [
         "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old"
     ],
-    "notes": "The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - scoop will install the latest ffmpeg instead.",
+    "notes": "The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version",
     "checkver": {
         "url": "https://download.imagemagick.org/ImageMagick/download/binaries/?C=N;O=D",
         "regex": "ImageMagick-([\\d.-]+)-Q16-HDRI-x64-dll\\.exe\\.asc"
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -17,29 +17,14 @@
         }
     },
     "innosetup": true,
-    "bin": [
-        "compare.exe",
-        "composite.exe",
-        "conjure.exe",
-        [
-            "convert.exe",
-            "imconvert"
-        ],
-        "dcraw.exe",
-        "identify.exe",
-        "IMDisplay.exe",
-        "magick.exe",
-        "mogrify.exe",
-        "montage.exe",
-        "stream.exe"
-    ],
     "env_set": {
         "MAGICK_HOME": "$dir",
         "MAGICK_CODER_MODULE_PATH": "$dir\\modules\\coders"
     },
     "env_add_path": ".",
     "post_install": [
-        "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old"
+        "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old",
+        "Rename-Item $dir\\convert.exe $dir\\imconvert.exe"
     ],
     "notes": "- The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version.\n- 'convert.exe' has been renamed to 'imconvert.exe' so that it does not conflict with the built-in command.\n",
     "checkver": {
--- a/bucket/imagemagick.json
+++ b/bucket/imagemagick.json
@@ -14,28 +14,16 @@
         }
     },
     "innosetup": true,
-    "bin": [
-        "compare.exe",
-        "composite.exe",
-        "conjure.exe",
-        "dcraw.exe",
-        "identify.exe",
-        "IMDisplay.exe",
-        "magick.exe",
-        "mogrify.exe",
-        "montage.exe",
-        "stream.exe"
-    ],
     "env_set": {
         "MAGICK_HOME": "$dir",
         "MAGICK_CODER_MODULE_PATH": "$dir\\modules\\coders"
     },
     "env_add_path": ".",
-    "post_install": [
-        "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old",
-        "Rename-Item $dir\\convert.exe $dir\\imconvert.exe"
+    "post_install": "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old",
+    "notes": [
+        "- The bundled ffmpeg has been renamed to 'ffmpeg.exe.old' to prevent conflict with a standalone ffmpeg installation - if you need it for ImageMagick tools, rename it back to 'ffmpeg.exe'.",
+        "- 'convert.exe' is deprecated in v7 (it also conflicts with the builtin Windows 'convert' utility). Use 'magick convert ...' instead."
     ],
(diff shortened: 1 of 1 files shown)


niheaven:
Please resolve conflicts

tech189:
I have resolved the conflicts and I have kept your changes to the checkver section of the manifest, so it should be okay now!

bucket/imagemagick.json
```diff
@@ -6,14 +6,15 @@
     "depends": "ffmpeg",
```
pratikpc, line 6: Should this be depends or suggest?

Personally I believe everything should move to suggests because users can install dependencies on their own as well

tech189: According to [this](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256215) it seems ffmpeg is not a proper dependency:

>Thierry Thomas freebsd_committer 2021-05-28 08:21:51 UTC  
>Actually ffmpeg is used for video delegation when it is found, but this is not a real dependency.

I've only used it for photo editing, so I can't be sure. But since ffmpeg is an extra 100MB maybe it's wise to put it as a 'suggest' with a note to install ffmpeg if the user needs it for video editing - I'll do that.

tech189: Done that, should be good to go now.

bucket/imagemagick.json
```diff
@@ -31,17 +34,26 @@
...
     ],
+    "env_set": {
+        "MAGICK_HOME": "$dir",
+        "MAGICK_CODER_MODULE_PATH": "$dir\\modules\\coders"
+    },
+    "env_add_path": ".",
+    "post_install": [
+        "Rename-Item $dir\\ffmpeg.exe $dir\\ffmpeg.exe.old"
+    ],
+    "notes": "The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version",
```
rashil2000, line 45: Now that you have removed the dependency, you can remove the post_install and note.

bucket/imagemagick.json
```diff
@@ -3,17 +3,20 @@
...
         },
         "32bit": {
-            "url": "https://download.imagemagick.org/ImageMagick/download/binaries/ImageMagick-7.1.0-portable-Q16-HDRI-x86.zip",
-            "hash": "366bab6ef6166412f2a456177026bb93202e265d0399314ffdc4820af277a065"
+            "url": "https://www.imagemagick.org/download/binaries/ImageMagick-7.1.0-13-Q16-HDRI-x86-dll.exe",
+            "hash": "477bf261a8ad6d3034e0b6f58ce7abdeb9a2c6186c9caef34e64fe9e24eafa42"
         }
     },
+    "innosetup": true,
     "bin": [
```
rashil2000, line 20: Remove the "convert.exe" from the bins to prevent conflict with Windows' internal convert utility. Add a note that ImageMagick's convert has been aliased as "imconvert"

rashil2000 (changes requested):

bucket/imagemagick.json
tech189: However, if I remove the post_install, ffmpeg will be added to the path (because of the `"env_add_path": "."`) Then if the user  decides later to install ffmpeg they will have two conflicting ffmpegs in their path. I think it should be kept for that reason.

rashil2000: The thing is, they have probably kept an older version of ffmpeg to prevent any incompatibilities with a newer version. Suggesting users to install the latest version via `scoop install ffmpeg` might cause issues with ImageMagick.

Also, I just noticed another thing - we are creating shims for binaries that are already in path (because of env_add_path) - which is very wasteful...

bucket/imagemagick.json
tech189: Done!

bucket/imagemagick.json
tech189: Yes it's possible it'll cause issues. So if I leave ffmpeg in the path and the user then installs ffmpeg from scoop, there will be a conflict. And they will have to manually rename/remove the imagemagick version every time there is a new version of imagemagick... Is there a better way?

I didn't notice that, nice catch. To resolve that I can remove all of the shims. Apart from convert.exe which I will have to rename to imconvert.exe in post_install.

bucket/imagemagick.json
```diff
@@ -17,29 +17,14 @@
...
-        "magick.exe",
-        "mogrify.exe",
-        "montage.exe",
-        "stream.exe"
-    ],
     "env_set": {
         "MAGICK_HOME": "$dir",
         "MAGICK_CODER_MODULE_PATH": "$dir\\modules\\coders"
     },
     "env_add_path": ".",
```
pratikpc, line 24: Is there an advantage of adding to path over bin?

Is it possible for us to add DLLs to bin?

Is it possible for us to copy the DLLs to a totally new folder and add that to bin?

rashil2000: I don't think e can create shims for DLLs. 

The number and names of DLLs is likely to be non-static, it might be problematic in the future if we hardcode them now.

pratikpc: The issue with adding to path is that we end up adding every new dependency present in them as well.

For example, if they add a future binary which causes a convert like issue that can be problematic.

Also renaming convert is likely to cause issue for scripts which might use convert.exe from MAGICK_HOME directory.

Are we sure we are not breaking workflows?

bucket/imagemagick.json
rashil2000: Let's just not suggest users to install the latest version. We can change the note to say - "The bundled ffmpeg has been renamed to 'ffmpeg.exe.old' to prevent conflict with a standalone ffmpeg installation - if you need it for ImageMagick tools, rename it back to 'ffmpeg.exe'."

bucket/imagemagick.json
rashil2000: The internal Windows `convert` utility is usually present higher up in PATH (because that PATH segment comes from system-wide environment instead of user-wide environment). So any scripts that do use `convert`, will use the internal `convert` (unless of course they specify the full path). I think this is a fair assumption we can make.

pratikpc: > internal convert (unless of course they specify the full path)

scripts which would use an environment variable like MAGICK_HOME are more than likely to use and specify the full path due to the creators being aware of the collision of names between the ImageMagick convert and Windows convert.

And if there is a low chance of collision as you mention, why rename it anyways?

Why not just add it to bin?

If existing Scoop based scripts using imconvert exists, we ensure they don't break and existing MAGICK_HOME based scripts too won't break.

rashil2000: I think a more experienced ImageMagick user like @tech189 might be able suggest better here...

bucket/imagemagick.json
```diff
@@ -3,45 +3,41 @@
...
+    "notes": "- The bundled ffmpeg is outdated and has been renamed to ffmpeg.exe.old - if you need it, run 'scoop install ffmpeg' for the latest version.\n- 'convert.exe' has been renamed to 'imconvert.exe' so that it does not conflict with the built-in command.\n",
     "checkver": {
         "url": "https://download.imagemagick.org/ImageMagick/download/binaries/?C=N;O=D",
         "regex": "ImageMagick-([\\d.-]+)-Q16-HDRI-x64-dll\\.exe\\.asc"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://download.imagemagick.org/ImageMagick/download/binaries/ImageMagick-$matchHead-portable-Q16-HDRI-x64.zip"
+                "url": "https://www.imagemagick.org/download/binaries/ImageMagick-$matchHead$matchTail-Q16-HDRI-x64-dll.exe"
```
pratikpc, line 37: Any reason we cannot use the download mirror?
https://github.com/ScoopInstaller/Main/issues/2733

bucket/imagemagick.json
tech189: Okay, done.

bucket/imagemagick.json
tech189: Apologies, I didn't mean to overwrite that mirror url! I'll change that back.

bucket/imagemagick.json
pratikpc: > How about we leave a note for those who upgrade to raise issues on Github?

I don't see this as the best of ideas.

rashil2000: I agree with @tech189 on the fact that most people will use ImageMagick through `magick <subcommand> ...` like `magick convert ...`, so maybe renaming it isn't that big of a deal (also the convert utility has been deprecated as tech189 said).

tech189: Ok just to confirm, I've kept the shims just in case and left a note to users to use 'magic convert ...' instead. We can remove the shims at a later stage if someone encounters issues with the duplicates.

> I don't see this as the best of ideas.

Sure, instead I'll monitor the issues and @ the owner of the rinohtype repo mentioned above to check their workflows are okay.

Otherwise, I think we're all done now?

rashil2000: Wait, the issue was only in the `convert.exe` shim, not others (as they were available both before this PR and will be available after the PR - they won't break any workflows)

tech189: That's right, I'm not touching any of the other shims. I'm just removing the convert.exe shim and renaming the binary to imconvert.exe to avoid conflicts.

rashil2000: Um no no, I think there's a misunderstanding. The problem of having duplicated binaries is still unsolved. 

Let me fix it.

rashil2000: Done

tech189: Ah okay, looks good to me. Just tested it. Sorry for the misunderstanding.

rashil2000: No worries 😊

rashil2000 (approved):

tech189:
Thanks for all your help everyone!

@brechtm sorry for the ping, but it would be great if you could just update and double check that this improvement to the imagemagick manifest doesn't break any of your workflows in rinohtype?

brechtm:
> @brechtm sorry for the ping, but it would be great if you could just update and double check that this improvement to the imagemagick manifest doesn't break any of your workflows in rinohtype?

No problem. In fact, thanks for the heads-up! The update breaks my GitHub Actions workflow. It seems that no shims for the ImageMagick tools are created. Compare section _Install tools required by the regression tests (Windows)_ of these runs:

* new, broken: https://github.com/brechtm/rinohtype/runs/4125719534
* old, working: https://github.com/brechtm/rinohtype/runs/4104983552

rashil2000:
The shims are not created because the ImageMagick binaries are now in PATH. Creating additional shims would just be redundant.

tech189:
The issue in your regression test is on this line: https://github.com/brechtm/rinohtype/blob/06e7490e316ba74470ccb3af548c101ea834126f/tests_regression/helpers/diffpdf.py#L190
```
if (MAGICK and not which(*MAGICK)) or not which('compare'):
```
When I run an adapted version on my computer:
```
from shutil import which
import sys
print(which("magick"))
print(which("compare"))

MAGICK = ['magick'] if sys.platform == 'win32' else []
if (MAGICK and not which(*MAGICK)) or not which('compare'):
    print("ImageMagick is required")
```
I get the following output:
```
C:\Users\$USER\scoop\apps\imagemagick\current\magick.EXE
C:\Users\$USER\scoop\apps\imagemagick\current\compare.EXE
```
So I think there might be an issue with your test suite's path variable. You could also solve it by using the new `MAGICK_HOME` variable which will direct you straight to where imagemagick is installed.

brechtm:
I suppose the changes to PATH do not transfer to the next steps in the workflow. I seem to remember having a similar problem with Chocolatey before, so I'm not sure whether this is an issue with [setup-scoop](https://github.com/MinoruSekine/setup-scoop/) or GitHub actions in general.

brechtm:
After adding `$env:USERPROFILE\scoop\apps\imagemagick\current` to the PATH manually in my GitHub workflow, `magick` is found but crashes:

```
convert: RegistryKeyLookupFailed `CoderModulesPath' @ error/module.c/GetMagickModulePath/665.
convert: no decode delegate for this image format `PGM' @ error/constitute.c/ReadImage/572.
convert: image sequence is required `-clone' @ error/convert.c/ConvertImageCommand/1013.
convert: RegistryKeyLookupFailed `CoderModulesPath' @ error/module.c/GetMagickModulePath/665.
convert: no decode delegate for this image format `PGM' @ error/constitute.c/ReadImage/572.
convert: image sequence is required `-clone' @ error/convert.c/ConvertImageCommand/1013.
```

Strangely, this doesn't happen on my local machine when using the same ImageMagick version installed using scoop.

brechtm:
One difference is that the GitHub runner is running Windows Server 2019 and my local machine is running Windows 10 Home.

> OS Name:                   Microsoft Windows Server 2019 Datacenter
> OS Version:                10.0.17763 N/A Build 17763

brechtm:
It appears that the installer [sets up registry keys](https://legacy.imagemagick.org/discourse-server/viewtopic.php?f=2&t=34990&p=161034&hilit=RegistryKeyLookupFailed#p161034) that are required at run-time. I'm assuming these are already present when ImageMagick was installed before using the installer, so the scoop-installed ImageMagick only fails to work on a fresh "machine" such as a GitHub Actions runner.

Setting `MAGICK_CODER_MODULE_PATH` fixes this issue. However, together with the removal of the shims, this makes it much less straightforward to install/use ImageMagick in a GitHub workflow.

rashil2000:
> It appears that the installer [sets up registry keys](https://legacy.imagemagick.org/discourse-server/viewtopic.php?f=2&t=34990&p=161034&hilit=RegistryKeyLookupFailed#p161034) that are required at run-time.
> 
> Setting `MAGICK_CODER_MODULE_PATH` fixes this issue. However, together with the removal of the shims, this makes it much less straightforward to install/use ImageMagick in a GitHub workflow.

I think this is why @tech189 explicitly modified the manifest to set the MAGIC_HOME and MAGICK_CODER_MODULE_PATH variables.

brechtm:
> I think this is why @tech189 explicitly modified the manifest to set the MAGIC_HOME and MAGICK_CODER_MODULE_PATH variables.

Yes. It's just that this requires extra work on GitHub actions, which is of course the scoop's fault.

rashil2000:
... how is it Scoop's fault?

brechtm:
Oops. I meant to say **not** scoop's fault, of course 😄

tech189:
Ah, annoying that github-actions works that way, I hadn't accounted for such a use case in my tests. Glad you found a workaround now! :D

One feature I think scoop should have is a command like `scoop shim` so that in this sort of situation you can manually create shims as needed. It's tracked in this issue https://github.com/ScoopInstaller/Scoop/issues/3303 but it hasn't been discussed in a while.

PR #2888, opened 2021-11-05
pelook: Add version 1.73

issaclin32:
[pelook](http://bytepointer.com/tools/index.htm#pelook) is an information tool for 32 and 64 bit Windows EXE, DLL, driver and OBJ files (PE/COFF images). This tool is suitable for reverse engineers or anyone needing to delve into the internals of Windows PE files.

### DIFF
--- a/bucket/pelook.json
+++ b/bucket/pelook.json
@@ -4,16 +4,16 @@
     "homepage": "http://bytepointer.com/tools/index.htm#pelook",
     "license": {
         "identifier": "Freeware",
-        "url": "http://bytepointer.com/tools/index.htm"
+        "url": "https://bytepointer.com/tools/index.htm"
     },
-    "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
+    "url": "https://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
     "hash": "12baccfb7db419c7533bcf495392e1baf9740c8b3bc558e496201c3c1a046cd7",
     "bin": "pelook.exe",
     "checkver": {
-        "url": "http://bytepointer.com/tools/pelook_changelist.htm",
+        "url": "https://bytepointer.com/tools/pelook_changelist.htm",
         "regex": "\\d{4}-\\d{2}-\\d{2}\\s+([\\d.]+)"
     },
     "autoupdate": {
-        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip"
+        "url": "https://bytepointer.com/download.php?name=pelook.zip#/dl.zip"
     }
 }
--- a/bucket/pelook.json
+++ b/bucket/pelook.json
@@ -14,7 +14,6 @@
         "regex": "\\d{4}-\\d{2}-\\d{2}\\s+([\\d.]+)"
     },
     "autoupdate": {
-        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
-        "extract_dir": "MagicaVoxel-$version-win64"
+        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip"
     }
 }
--- a/bucket/pelook.json
+++ b/bucket/pelook.json
@@ -1,0 +1,20 @@
+{
+    "version": "1.73",
+    "description": "Information tool for Windows EXE, DLL, driver and OBJ files (PE/COFF images)",
+    "homepage": "http://bytepointer.com/tools/index.htm#pelook",
+    "license": {
+        "identifier": "Freeware",
+        "url": "http://bytepointer.com/tools/index.htm"
+    },
+    "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
+    "hash": "12baccfb7db419c7533bcf495392e1baf9740c8b3bc558e496201c3c1a046cd7",
+    "bin": "pelook.exe",
+    "checkver": {
+        "url": "http://bytepointer.com/tools/pelook_changelist.htm",
+        "regex": "\\d{4}-\\d{2}-\\d{2}\\s+([\\d.]+)"
+    },
+    "autoupdate": {
+        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
+        "extract_dir": "MagicaVoxel-$version-win64"
+    }
+}


bucket/pelook.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "1.73",
+    "description": "Information tool for Windows EXE, DLL, driver and OBJ files (PE/COFF images)",
+    "homepage": "http://bytepointer.com/tools/index.htm#pelook",
+    "license": {
+        "identifier": "Freeware",
+        "url": "http://bytepointer.com/tools/index.htm"
+    },
+    "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
```
pratikpc, line 9: Just confirming

It is 32 bit right?

issaclin32: It is 32bit.
```
dumpbin /headers pelook.exe
Microsoft (R) COFF/PE Dumper Version 14.29.30133.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file pelook.exe

PE signature found

File Type: EXECUTABLE IMAGE

FILE HEADER VALUES
             14C machine (x86)
               4 number of sections
        5F5D5BDC time date stamp Sun Sep 13 07:38:04 2020
               0 file pointer to symbol table
               0 number of symbols
              E0 size of optional header
             103 characteristics
                   Relocations stripped
                   Executable
                   32 bit word machine
```

bucket/pelook.json
```diff
@@ -0,0 +1,20 @@
...
+    "hash": "12baccfb7db419c7533bcf495392e1baf9740c8b3bc558e496201c3c1a046cd7",
+    "bin": "pelook.exe",
+    "checkver": {
+        "url": "http://bytepointer.com/tools/pelook_changelist.htm",
+        "regex": "\\d{4}-\\d{2}-\\d{2}\\s+([\\d.]+)"
+    },
+    "autoupdate": {
+        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
+        "extract_dir": "MagicaVoxel-$version-win64"
```
pratikpc, line 18: Why is there an extract dir here and not outside?

bucket/pelook.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
+    "hash": "12baccfb7db419c7533bcf495392e1baf9740c8b3bc558e496201c3c1a046cd7",
+    "bin": "pelook.exe",
+    "checkver": {
+        "url": "http://bytepointer.com/tools/pelook_changelist.htm",
+        "regex": "\\d{4}-\\d{2}-\\d{2}\\s+([\\d.]+)"
+    },
+    "autoupdate": {
+        "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
```
pratikpc, line 17: Do we really need to rename the ZIP file?

bucket/pelook.json
issaclin32: Oops. Forgot to remove that,

pratikpc: Ohh. I thought that it needed an extract_dir but this wasn't the correct one!!

bucket/pelook.json
issaclin32: If we don't rename it, the extraction will fail due to wrong file name.

pratikpc: Any idea why?

Is it because of the ?name= in the URL?

pratikpc:
@issaclin32 is the website safe?

My antivirus is blocking access but VirusTotal says the domain is fine

Also do they support https?

bucket/pelook.json
```diff
@@ -0,0 +1,19 @@
+{
+    "version": "1.73",
+    "description": "Information tool for Windows EXE, DLL, driver and OBJ files (PE/COFF images)",
+    "homepage": "http://bytepointer.com/tools/index.htm#pelook",
+    "license": {
+        "identifier": "Freeware",
+        "url": "http://bytepointer.com/tools/index.htm"
+    },
```
Valinor, line 8: use https://bytepointer.com/tools/index.htm

bucket/pelook.json
```diff
@@ -0,0 +1,19 @@
...
+    "url": "http://bytepointer.com/download.php?name=pelook.zip#/dl.zip",
```
Valinor, line 9: use https https://bytepointer.com/download.php?name=pelook.zip#/dl.zip

bucket/pelook.json
```diff
@@ -0,0 +1,19 @@
+{
+    "version": "1.73",
+    "description": "Information tool for Windows EXE, DLL, driver and OBJ files (PE/COFF images)",
+    "homepage": "http://bytepointer.com/tools/index.htm#pelook",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://bytepointer.com/tools/index.htm"
```
pratikpc, line 7: Hey does this URL point to the license?

issaclin32: The license into is written on the top of the page.
```
FREEWARE Tools

*LEGAL DISCLAIMER: The content on this website, including tools are for educational purposes only. This website makes no warranties express or implied. You use any tools at your own risk!

Welcome to my FREEWARE tools page. The Open-Source GNU GPL tools listed here may be freely used and altered according to the enclosed license. The other tools may be used free of charge provided they are not altered without written permission if redistributed to the public. All tools may be freely distributed as long as they retain their original contents as available on this website.

These tools are a mix of mostly command line applications originally built for Windows XP although they should run on Windows 7 and up. Some newer tools include a Linux version. While the Windows tools have a dependency on the Visual C++ runtime (available as a separate download), I decided to make things simpler for the Linux versions by linking the executables statically, although the resutling executables are larger than I'd like.
```

pratikpc (approved):
Manifest looks fine but my antivirus is blocking access to the domain

issaclin32:
> Manifest looks fine but my antivirus is blocking access to the domain

Mine (*Avira*) works fine though.
Which antivirus software are you using? maybe we can sent a false positive and let the antivirus to check the problem.

Valinor (approved):

bucket/pelook.json
issaclin32: > Is it because of the ?name= in the URL?

It is. think this needs to be fixed in *Scoop*'s main program.

issaclin32:
I added the `discussion` label because there is a problem (related to this PR) need to be fixed upstream.

PR #2891, opened 2021-11-05
Update appveyor powershell names & scoop git url

### LINKED ISSUE #2824: [CI] Update to PowerShell 7
I've noticed that the CI is currently using PowerShell 5 and PowerShell Core 6 while PowerShell 7 is the current recommended version. I think we need to add PowerShell 7 into the CI process.

I'm also pretty sure PowerShell Core 6 is EOL, so once we have PowerShell 7 working we would probably want to remove it.

tech189:
Closes #2824

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,14 +14,14 @@ environment:
   innounp: '%scoop_helpers%\innounp\innounp.exe'
   matrix:
     - PowerShell: 5
-    - PowerShell: 6
+    - PowerShell: Current
 cache:
   - '%USERPROFILE%\Documents\WindowsPowerShell\Modules -> appveyor.yml'
   - C:\projects\helpers  -> appveyor.yml, test\bin\*.ps1
 matrix:
   fast_finish: true
 init:
-  - ps: if(!(Test-Path "$env:SCOOP")) { git clone -q --depth=1 "https://github.com/lukesampson/scoop" "$env:SCOOP" }
+  - ps: if(!(Test-Path "$env:SCOOP")) { git clone -q --depth=1 "https://github.com/ScoopInstaller/Scoop" "$env:SCOOP" }
 for:
   - matrix:
       only:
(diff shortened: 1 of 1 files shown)


stevehipwell (commented):
/lgtm

stevehipwell (approved):
/lgtm

rashil2000:
@stevehipwell is there a way to test this before merging into master (just to be sure it doesn't break anything)?

stevehipwell:
I don't think there is an easy way.

appveyor.yml
```diff
@@ -14,14 +14,14 @@ environment:
   innounp: '%scoop_helpers%\innounp\innounp.exe'
   matrix:
     - PowerShell: 5
-    - PowerShell: 6
+    - PowerShell: Current
```
pratikpc, line 17: Can you share an official link which mentions this is the correct behaviour to get latest Powershell?

rashil2000: The matrix just mentions the name, this part is not responsible for pulling in PowerShell

pratikpc: So would using PowerShell 5 for example lead to it also using PowerShell's newest version? That does sound a bit weird to me.

Not very well versed with AppVeyor.

stevehipwell: @pratikpc the names here are just matrix keys so the job gets run once for `powershell` and once for `pwsh`, changing the name stops the misconception that we're using PS 6 as AppVeyor keeps the version of `pwsh` current (so v7 ATM).

pratikpc: @stevehipwell ps indicates PowerShell 5 while pwsh indicates PowerShell's newest version right?

pratikpc (approved):

rashil2000:
Merging this now, fingers crossed 🤞


Thank you for the contribution @tech189 :)

PR #2893, opened 2021-11-05
Add scoop installer for cfsec

owenrumney:
- add cfsec.json for install cfsec, lifted the content of tfsec as a
  template.


pratikpc:
@rashil2000 this faces the same issue at torrenttools IMHO. It has only 33 stars and is a fairly new project.

rashil2000:
Hmm, but there's a difference - the org that made cfsec also has some pretty popular tools (tfsec, 3.5k stars, already in main bucket, and some other more popular ones). So I'm not sure how to go about this one.

owenrumney:
Just to offer some input - I'm one of the core developers of tfsec and we saw a gap for CloudFormation scanning. So yes, it's new and doesn't have a big user base yet, but it does come from the same stable as some big tooling

pratikpc (approved):
Manifest is fine.

PR #2895, opened 2021-11-05
hadolint: Update to version 2.8.0

darkvertex:


rashil2000:
Add the autoupdate section so that updates are handled by itself

darkvertex:
> Add the autoupdate section so that updates are handled by itself

I'm new to the autoupdate functionality, but I read the docs and added what I think is the correct data for the section.

Does it look good to you? Is there an easy way or command I can run to validate it's all valid and goood?

pratikpc:
@darkvertex see https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#using-checkverps1-to-query-and-autoupdate

issaclin32 (approved):

pratikpc (approved):

rashil2000 (approved):

PR #2905, opened 2021-11-08
wyam: deprecate manifest

### LINKED ISSUE #809: wyam@2.2.9: Shim wyam.exe error
The installation fails on 'Can't shim 'wyam.exe': File doesn't exist.'
Wyam version 2 was updated to be used as a dotnet global tool and includes a wyam.dll file.

https://github.com/Wyamio/Wyam/releases/tag/v2.0.0
https://wyam.io/docs/usage/obtaining

I'm not sure what the appropriate fix for this should be.  

image

rashil2000:
Closes #809

image


pratikpc:
@issaclin32 does excavator even run on deprecated folder?

pratikpc (approved):

rashil2000:
> _wyam_ removed their .exe at some point. That is why the exe-shim isn't working. I suggest adding:
> 
> ```
>     "pre_install": [
>         "Set-Content \"$dir\\wyam.ps1\" \"dotnet `\"$dir\\wyam.dll`\"\" | Out-Null"
>     ],
>     "bin": "wyam.ps1",
> ```

Then you will also need to add `depends: dotnet-sdk`.

The recommended way of installing Wyam (even if it's deprecated) through the dotnet tool.

The same reason we don't have any npm packages on Scoop (because they are better managed by the `npm` CLI). I think this should be deprecated instead.

issaclin32:
> The recommended way of installing Wyam (even if it's deprecated) through the dotnet tool.
> 
> The same reason we don't have any npm packages on Scoop (because they are better managed by the `npm` CLI). I think this should be deprecated instead.

I see your point.

issaclin32 (approved):

PR #2908, opened 2021-11-08
Persist rclone.conf in application directory for portability

### LINKED ISSUE #1634: Persist rclone.conf for rclone
According to the [document](https://rclone.org/docs/#config-config-file), "If there is a file rclone.conf in the same directory as the rclone executable it will be preferred". Suggest to add `persist` for this file for portability.
```
"persist": "rclone.conf"
```

CrendKing:
Closes #1634.

Also, I notice many manifests that utilizes the similar tech often omit the `-ItemType File` part, which results in creating an empty directory instead of file. So I decide to put it to guarantee a file, regardless of PowerShell's default.

### DIFF
--- a/bucket/rclone.json
+++ b/bucket/rclone.json
@@ -15,7 +15,7 @@
             "extract_dir": "rclone-v1.56.2-windows-386"
         }
     },
-    "pre_install": "New-Item -ItemType File \"$persist_dir\\rclone.conf\" -ErrorAction SilentlyContinue",
+    "pre_install": "if (!(Test-Path \"$persist_dir\\rclone.conf\")) { New-Item -ItemType File \"$persist_dir\\rclone.conf\" | Out-Null }",
     "uninstaller": {
         "script": "Copy-Item \"$dir\\rclone.conf\" \"$persist_dir\\\" -Force"
     },
--- a/bucket/rclone.json
+++ b/bucket/rclone.json
@@ -15,7 +15,12 @@
             "extract_dir": "rclone-v1.56.2-windows-386"
         }
     },
+    "pre_install": "New-Item -ItemType File \"$persist_dir\\rclone.conf\" -ErrorAction SilentlyContinue",
+    "uninstaller": {
+        "script": "Copy-Item \"$dir\\rclone.conf\" \"$persist_dir\\\" -Force"
+    },
     "bin": "rclone.exe",
+    "persist": "rclone.conf",
     "checkver": {
         "github": "https://github.com/rclone/rclone"
     },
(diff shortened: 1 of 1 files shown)


rashil2000:
You should check for the file in persist_dir, and create it in dir. See https://github.com/ScoopInstaller/Extras/blob/master/bucket/jcpicker.json for example.

CrendKing:
Why? If the file already exists, `New-Item` just error and exit, and the `-ErrorAction SilentlyContinue` hides the expected error message.

wiredmind:
What if the file doesn't exist and it cannot be created because antivirus program 8 it, or the volume is corrupt, or the iSCSI volume on which it resides disconnects? While this is of little consequence here, it hides and ignores errors which it doesn't handle.

CrendKing:
If something catastrophic happened (like the hardware fails), wouldn't the whole update process fail to begin with? For antivirus, why would any real AV block an empty file to be created? Persist dir hosted on a network volume could disconnect, true, but I think the user would notice that all his apps are misbehaving earlier than this.

All being said, if you guys insist on a existence check, I can change. No biggy.

rashil2000:
A check for the file would be better than a (silently) failing cmdlet. I feel it's better practice.

rashil2000 (approved):

CrendKing:
@rashil2000 I see that you removed the uninstaller section in https://github.com/ScoopInstaller/Main/pull/2908/commits/6465b6074e965e3728cbd9b1bfce0aa7455d792a. Wouldn't it break the copy process?

rashil2000:
The persisted file rclone.conf will stay in the persist directory if you do `scoop uninstall rclone`, so no need for explicit copying.

CrendKing:
When you run `rclone config`, rclone will break the hard link. Same happens to XnView MP. Meaning, the one in persist will always be 0 byte without the "copy back" mechanism.

rashil2000:
So `rclone config` deletes the older file, and then creates a new one (instead of modifying that old file)?

CrendKing:
I don't think it's delete-create, because if it's symbolic link the synchronization is always maintained. I don't know how do they write the config, but it's open source so we can check it out.

rashil2000:
That's definitely strange. If I understand correctly, it saves the older file with the `.old` extension and creates new one.

We can make an exception for this. Sorry for goofing this up. I will push back the `uninstaller` field.

CrendKing:
It renames the current config to ".old", moves the new config to `rconfig.conf`, then delete the ".old". The moment the .old is deleted the link is broken.

CrendKing:
@rashil2000 BTW, rclone updated to 1.57. Could you also update the version while you are on it? Thank you so much!

PR #2914, opened 2021-11-09
Update OpenOCD version to 0.11.0 (20210729)

luojia65:


bucket/openOCD.json
```diff
@@ -1,10 +1,10 @@
 {
-    "version": "20201228",
+    "version": "20210729",
     "description": "OpenOCD is an open-source tool that allows debugging various ARM devices with GDB using a wide variety of JTAG programmers.",
     "homepage": "https://gnutoolchains.com/arm-eabi/openocd/",
```
pratikpc, line 4: This is official website right?

https://openocd.org/

luojia65: Thank you! Will fix official website.

pratikpc:
Why not use releases from GitHub?

https://github.com/openocd-org/openocd/releases/download/v0.11.0/openocd-v0.11.0-i686-w64-mingw32.tar.gz

luojia65:
@pratikpc Thanks! I didn't noticed OpenOCD have an official github mirror here. Fixed and added some auto update settings.

bucket/openOCD.json
```diff
@@ -1,10 +1,20 @@
...
+    "hash": "b195e3c3e9fe3f4fe9dfd9780ff08336ef8ed3015a76cc883d9e21e57c456f33",
+    "bin": "bin\\openocd.exe",
+    "checkver": {
+        "github": "https://github.com/openocd-org/openocd",
+        "regex": "/releases/tag/([\\w.]+)"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/openocd-org/openocd/releases/download/$version/win-portacle.exe#/dl.7z"
```
pratikpc, line 16: Umm this won't work

Use https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#using-checkverps1-to-query-and-autoupdate with ForceUpdate param and then re-install

bucket/openOCD.json
```diff
@@ -1,10 +1,20 @@
...
-    "extract_dir": "OpenOCD-20201228-0.10.0",
-    "env_add_path": "bin"
+    "homepage": "https://openocd.org/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://github.com/openocd-org/openocd/releases/download/v0.11.0/openocd-v0.11.0-i686-w64-mingw32.tar.gz",
+    "hash": "b195e3c3e9fe3f4fe9dfd9780ff08336ef8ed3015a76cc883d9e21e57c456f33",
+    "bin": "bin\\openocd.exe",
+    "checkver": {
+        "github": "https://github.com/openocd-org/openocd",
+        "regex": "/releases/tag/([\\w.]+)"
```
pratikpc, line 11: Check if this is required.

bucket/openOCD.json
pratikpc: For offline testing.

There is no win-portacle.exe present.

This URL should match with version substitution the section above.

pratikpc: Also this is 32 bit right?
It should be in the global block

pratikpc (changes requested):

pratikpc:
@ScoopInstaller/maintainers OpenOCD present in both extras and main

One of them needs to be removed

https://github.com/ScoopInstaller/Extras/blob/master/bucket/openocd.json

issaclin32:
> @ScoopInstaller/maintainers OpenOCD present in both extras and main
> 
> One of them needs to be removed

I think we should remove the one in *extras*. The package is in *extras* because *main* bucket only accept "well known packages" before (maybe before mid-2019?). Now we put all CLI programs in *main*.

issaclin32:
I removed the duplicate package in *extras* (https://github.com/ScoopInstaller/Extras/commit/f82038f15256ddc4d33a75fceafd0d02136df1ac)

bucket/openOCD.json
luojia65: Removed, thanks

bucket/openOCD.json
luojia65: Hello! I've set autoupdate to github download url. Is this okay for this profile?

bucket/openOCD.json
```diff
@@ -1,10 +1,15 @@
...
-    "license": "Unknown",
-    "url": "https://sysprogs.com/getfile/1364/openocd-20201228.7z",
-    "hash": "2a45b7fee2974c36650c5bae95daa9b577c2241557332818f08e27a364de4311",
-    "extract_dir": "OpenOCD-20201228-0.10.0",
-    "env_add_path": "bin"
+    "homepage": "https://openocd.org/",
+    "license": "GPL-2.0-or-later",
+    "url": "https://github.com/openocd-org/openocd/releases/download/v0.11.0/openocd-v0.11.0-i686-w64-mingw32.tar.gz",
+    "hash": "b195e3c3e9fe3f4fe9dfd9780ff08336ef8ed3015a76cc883d9e21e57c456f33",
+    "bin": "bin\\openocd.exe",
```
pratikpc, line 8: Can you verify once if the file is 32 or 64 bit?

The URL mentioning w64 is a bit wierd.

I am not sure if it means mingw-w64 or Windows 64.

bucket/openOCD.json
pratikpc: @luojia65 that is not an issue.

Don't worry!!!

Create as many commits as you need to.

bucket/openOCD.json
luojia65: Seems like the distributed version is 32 bit executable (openocd.exe). I don't know why it's named as w64 :)

pratikpc (approved):

issaclin32 (approved):

PR #2915, opened 2021-11-09
questdb: Add version 6.0.9

niheaven:


rashil2000 (approved):

bucket/questdb.json
```diff
@@ -0,0 +1,32 @@
+{
+    "version": "6.0.9",
+    "description": "An open source SQL database designed to process time series data, faster.",
+    "homepage": "https://questdb.io",
+    "license": "Apache-2.0",
+    "suggest": {
+        "JDK": [
+            "java/oraclejdk",
```
pratikpc, line 8: Not around any longer

pratikpc: I guess this line could be removed

bucket/questdb.json
```diff
@@ -0,0 +1,32 @@
...
+    "persist": [
+        "conf",
+        "db",
+        "log"
+    ],
+    "checkver": {
+        "github": "https://github.com/questdb/questdb"
+    },
+    "autoupdate": {
+        "url": "https://github.com/questdb/questdb/releases/download/$version/questdb-$version-no-jre-bin.tar.gz",
```
pratikpc, line 29: This should be in 64 bit block BTW

pratikpc (changes requested):

bucket/questdb.json
niheaven: This can update all architecture, but yes, move it under `arch` is okay.

pratikpc: If it has 32 bit support then it should be outside though.

In architecture block, you had used 64 bit, so thought it was 64 bit.

pratikpc (approved):

PR #2921, opened 2021-11-09
openssh: Update to version 8.8p1-1

wstein:
new PR for openSSH update to version 8.3p1-1 as requested by pratikpc

Relates to #1494

### DIFF
--- a/bucket/openssh.json
+++ b/bucket/openssh.json
@@ -1,5 +1,5 @@
 {
-    "version": "8.2p1-1",
+    "version": "8.3p1-1",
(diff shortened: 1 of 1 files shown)


pratikpc:
@wstein could you update everything to the latest version with ZSTD support?

Check MSYS once and see how that would work and what versions would be correct dependencies.

For example, Wiki mentions OpenSSH 8.8 is latest.

niheaven:
Update to 8.8p1-1 and add some shim from `/usr/lib/ssh`

Issue #1708, opened 2020-12-25
[Request] Add luarocks


zzfnohell:
A package manager for Lua modules.

[Official site](https://luarocks.org/)
[Project](https://github.com/luarocks/luarocks)
[Wiki(Download)](https://github.com/luarocks/luarocks/wiki/Download)

rashil2000:
The (current) lua-for-windows.json manifest lists very old version of Lua. This will be a nice improvement.

niheaven:
I'm not familiar with Lua, and what will the config file be if we want to persist rocks?

```json
{
    "version": "3.7.0",
    "description": "The package manager for Lua modules",
    "homepage": "https://luarocks.org",
    "license": "MIT",
    "architecture": {
        "64bit": {
            "url": "https://luarocks.github.io/luarocks/releases/luarocks-3.7.0-windows-64.zip",
            "hash": "02755fc1e4cac076e17f393c1d8548f1c33467cbd79d3ca33e3c2db57887162e",
            "extract_dir": "luarocks-3.7.0-windows-64"
        },
        "32bit": {
            "url": "https://luarocks.github.io/luarocks/releases/luarocks-3.7.0-windows-32.zip",
            "hash": "ad1c5ac67e90152ab348acdf762020671ce924baf148eadd14d594472fbe141d",
            "extract_dir": "luarocks-3.7.0-windows-32"
        }
    },
    "pre_install": [
        "$file = 'config.lua'",
        "if (!(Test-Path \"$persist_dir\\$file\")) {",
        "   Write-Host 'File' $file 'does not exists. Creating.' -f Yellow",
        "   $CONT = @(\"rocks_trees = `\"$dir\\rocks`\"\")",
        "   Set-Content \"$dir\\$file\" ($CONT -join \"`r`n\") -Encoding Ascii",
        "}"
    ],
    "bin": [
        "luarocks.exe",
        "luarocks-admin.exe"
    ],
    "env_set": {
        "LUAROCKS_CONFIG": "$dir\\config.lua"
    },
    "persist": [
        "config.lua",
        "rocks"
    ],
    "checkver": {
        "url": "https://luarocks.github.io/luarocks/releases/",
        "regex": "luarocks-([\\d.]+)-windows"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://luarocks.github.io/luarocks/releases/luarocks-$version-windows-64.zip",
                "extract_dir": "luarocks-$version-windows-64"
[...]

PR #2929, opened 2021-11-11
luarocks: Add version 3.8.0

niheaven:
Close https://github.com/ScoopInstaller/Main/issues/1708

### DIFF
--- a/bucket/luarocks.json
+++ b/bucket/luarocks.json
@@ -3,7 +3,7 @@
     "description": "The package manager for Lua modules",
     "homepage": "https://luarocks.org",
     "license": "MIT",
-    "depends": "versions/lua53",
+    "depends": "lua",
     "architecture": {
         "64bit": {
             "url": "https://luarocks.github.io/luarocks/releases/luarocks-3.8.0-windows-64.zip",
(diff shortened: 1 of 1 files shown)


bucket/luarocks.json
```diff
@@ -0,0 +1,55 @@
...
+        "   Write-Host 'File' $file 'does not exists. Creating.' -f Yellow",
+        "   $CONT = @(\"rocks_trees = { `\"$rocks`\" }\")",
+        "   Set-Content \"$dir\\$file\" ($CONT -join \"`r`n\") -Encoding Ascii",
+        "}"
+    ],
+    "bin": [
+        "luarocks.exe",
+        "luarocks-admin.exe"
+    ],
+    "env_add_path": "rocks\\bin",
```
rashil2000, line 31: I downloaded the archive from https://luarocks.github.io/luarocks/releases/luarocks-3.8.0-windows-64.zip but it does not contain this folder. When is it created?

rashil2000 (approved):

bucket/luarocks.json
niheaven: When you add rocks and if it create bin.

I found this manifest needs more config and test... So convert to draft now

rashil2000:
In this PR can you also fix https://github.com/ScoopInstaller/Main/blob/master/bucket/lua-for-windows.json to change to a more maintained source here https://github.com/wsw0108/scoop-bucket/blob/main/bucket/lua-windows.json

niheaven:
Lua 5.3 is added in https://github.com/ScoopInstaller/Versions/pull/309

rashil2000:
Wait wait, no.

It doesn't make sense for a package in Main bucket (latest, stable versions) to have a dependency in Versions bucket (preview, unstable, older versions).

Besides, having the Lua package is not a hard requirement for Luarocks as people might have installed it through other means.

One more thing - the current package lua-for-windows is very out of date - their last release was in 2018 and it's not properly maintained. You should've added this to the Versions bucket instead, and put lua53 (which is the latest stable Luabinaries release) in the Main bucket.

niheaven:
Lua for Windows is just another Lua distribution that contains many butteries.

The newest version of Lua from Lua Binaries build is 5.4.2 and cannot be used with luarocks or the config file should be modified too much, so I use the past version 5.3 in `Versions`

rashil2000:
Okay, thanks for the info.

So now we can remove `"depends": "versions/lua53",`? 

Or maybe we can change it to `"suggest": "lua",`?

niheaven:
Change to 5.4, and should be merged after https://github.com/ScoopInstaller/Main/pull/2931

`lua` should be dependency, as the install script force luarocks using scooped lua

niheaven:
A simple `luarocks install` result:

image

rashil2000 (approved):

PR #2930, opened 2021-11-11
busybox-lean: Add version 4487-gd239d2d52

rashil2000:
A manifest for busybox already exists here. However, that manifest creates over 150 aliases! Many of which conflict with Windows internal tools - `sort`, `date`, `whoami`, `echo`, `find`, `mkdir`, `rmdir`, `tar`, `uptime`, `time` etc.

Many users will find this annoying, because these shims will never actually get called (in both cmd and powershell, internal commands have the highest priority).

I have created this 'lean' manifest which does not create these shims. This will also be helpful in small containers/images (like Actions). I have added a note for the existing busybox manifest too.

### DIFF
--- a/bucket/busybox-lean.json
+++ b/bucket/busybox-lean.json
@@ -1,0 +1,35 @@
+{
+    "version": "4487-gd239d2d52",
+    "description": "BusyBox is a single binary that contains many common Unix tools",
+    "homepage": "https://frippery.org/busybox",
+    "license": "GPL-2.0-only",
+    "notes": "To automatically create shims for all unix commands, install 'busybox' instead.",
+    "architecture": {
+        "64bit": {
+            "url": "https://frippery.org/files/busybox/busybox-w64-FRP-4487-gd239d2d52.exe#/busybox.exe",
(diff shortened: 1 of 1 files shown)


niheaven:
You'd better add this one to `Versions`, IMO.

rashil2000:
I don't think this will fit Versions. This is not a previous version, or old version or preview. The app is same.

It also fixes some problems with the existing busybox manifest (of non-reachable shims). I didn't change the existing manifest as I saw some examples in the Main bucket itself (like ffmpeg vs ffmpeg-shared, or b2sum vs b3sum, or cmder vs cmder-full). And of course to preserve backward compatibility I made a new manifest.

niheaven:
The examples here (cmder vs. cmder-full, etc.) are totally historical mistakes and should move one of them to `Versions`. It is an alternative version of the one in this bucket, so you'd better not put it here.

rashil2000:
I respectfully disagree. The distinction `lean` is what separates it, otherwise it's the entire same package (not alternative version).

niheaven:
Maybe alternative manifest? Okay, leave it here and wait for some discussion of buckets criteria

niheaven (approved):

rashil2000:
You mean some other name? I'm open to suggestions - what do you suggest?

PR #2937, opened 2021-11-11
starship: Update to v1.0.0

segevfiner:
Autoupdate starship to v1.0.0


niheaven:
Does autoupdate work?

segevfiner:
Do you mean did I try to `install` it? I did. By uninstalling the previous version and then `install` from the autoupdated manifest (`checkver.ps1 -u`).

PR #2941, opened 2021-11-12
clangd: Add version 13.0.0

rashil2000:
Language server for C/C++. Inspired by https://github.com/ScoopInstaller/Main/blob/master/bucket/rust-analyzer.json

### DIFF
--- a/bucket/clangd.json
+++ b/bucket/clangd.json
@@ -1,0 +1,25 @@
+{
+    "version": "13.0.0",
+    "description": "Language server for C/C++, by the LLVM project",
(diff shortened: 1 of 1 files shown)


pratikpc (approved):
LLVM also contains clangd.

LLVM users are probably equally likely to install clangd.

PR looks good but w should consider this

rashil2000:
You're right, but the LLVM package is more than 300MB and contains a lot of stuff.

This language server is compiler-agnostic, meaning it doesn't need the LLVM compiler to work - it works with GCC and MSVC as well.

PR #2954, opened 2021-11-15
chore: rename manifest - beef and zig

rashil2000:


issaclin32 (approved):

pratikpc:
This will break workflows though

And the previously chosen names were more clearer IMHO.

rashil2000:
Users updating scoop will get notified about the rename. Which other workflows would break?

pratikpc:
I doubt if most Scoop users actually read those.

If you are updating via a week there are going to be way too many updates for someone to notice for example.

Also beef and zig's website are named beeflang and ziglang respectively so the names were also appropriate.

I don't think such breaking changes should be made till we can actively support a redirect/rename feature for manifests within Scoop such that no one would actively have to reinstall.

rashil2000:
Go, Rust, Nim - all have websites that end with -lang. But their manifests don't. 

If we are to follow this, we need to standardize it.

pratikpc:
> If we are to follow this, we need to standardize it.

But not by breaking workflows. The best idea would have been to add support within Scoop to redirect/rename manifests on its own without breaking anything first and then create PRs which perform standardization like this.

rashil2000:
I agree. Moving forward that's definitely a better way.

However I believe this was a very tiny change - these 2 manifests didn't inherently have anything that would break stuff (no env vars modification, no PATH addition, no persistence etc.), so we can let this go for now.

issaclin32:
I think this change is *doable* because:

(1) This will only cause users who installed `beeflang` and `ziglang` not being able to update the package. This will not affect their daily workflow, unless they need to update the package above a certain version. In this case, they can easily figure out what is happening by `scoop search`.

(2) Both packages do not have *persist*.

(3) It is better to put a unified standard on the naming issue. Both `v` and `vlang` is acceptable, but we have to choose from one of them. Otherwise it is hard for us to make the decision when we encounter the same problem in the future.
Calling the language `rlang`, `vlang`, `golang` is clearer, and sometimes necessary, especially without context. But in the case of **package names**. Using `r`, `v`, `go` is just OK. (considering they also named their binaries as `v.exe`, etc.)

I like the idea about **redirect mechanism** proposed by @pratikpc. We can try to implement and discuss about it later.

loic-sharma:
As a Scoop and Zig newbie, +1 for this rename. I wanted to update Zig and it took me a minute to realize my installed manifest was called `ziglang`. I am glad the Scoop manifest now matches the executable name, and I'm happy to re-install zig once to never run into this problem again :)

PR #2971, opened 2021-11-18
openssh: Change to Microsoft build and update to 8.6.0.0p1

### LINKED ISSUE #1769: openssh@8.2p1-1: download failed
(py3x64) E:\Document\code-git>scoop install openssh
Installing 'openssh' (8.2p1-1) [64bit]
bash-4.4.023-2-x86_64.pkg.tar.xz (1.9 MB) [=============================================================================================================] 100%
Checking hash of bash-4.4.023-2-x86_64.pkg.tar.xz ... ok.
gcc-libs-9.3.0-1-x86_64.pkg.tar.xz (1.2 MB) [===========================================================================================================] 100%
Checking hash of gcc-libs-9.3.0-1-x86_64.pkg.tar.xz ... ok.
远程服务器返回错误: (404) 未找到。
URL http://repo.msys2.org/msys/x86_64/heimdal-7.7.0-1-x86_64.pkg.tar.xz is not valid

niheaven:
Ref: https://github.com/ScoopInstaller/Main/pull/2921#issuecomment-971634048

Fix #1769

Since I don't use scoop's openssh, and even after installing this one, system installed OpenSSH has higher priority, so please report if this may break someone's user environment.


rashil2000:
Actually, this Microsoft's build already exists in Main - https://github.com/ScoopInstaller/Main/blob/master/bucket/win32-openssh.json.

We can do this:
1. Delete the manifest which is modified in this PR (openssh.json)
2. Rename the manifest win32-openssh.json to openssh.json
3. Add a note in this renamed manifest - `Windows 10 or higher includes OpenSSH built-in, it is recommended to use the system OpenSSH there.`

rashil2000 (approved):

bucket/openssh.json
```diff
@@ -1,69 +1,49 @@
...
+        "sshd.exe"
+    ],
+    "checkver": {
+        "github": "https://github.com/PowerShell/Win32-OpenSSH",
+        "regex": "tag/V([\\d.]+p\\d)-Beta"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/PowerShell/Win32-OpenSSH/releases/download/V$version-Beta/OpenSSH-Win64.zip"
```
pratikpc, line 37: These are beta releases though.

Best of ideas to use this repo?

rashil2000: All there releases are beta :)

I guess the non-beta ones are those that ship with the OS updates.

pratikpc: That's what I mean

Not the best of ideas in my humble opinion

rashil2000: This manifest is not meant for Windows 10 anyway (as mentioned in the note) - since these builds are included with the system. This is for Windows 7, 8, 8.1 users.

pratikpc: > as mentioned in the note

It's possible for notes to be ignored

If the note is meant for something like this, then we should display the note while installation starts

With something like

Warning: Not recommended for using with Windows 10

And then ask the user to key in with a Proceed or not. (which ends up breaking scripts)

Other option is to disable installation of this based on Windows versions

We would need to create tooling which finds current Windows versions and checks if installation is allowed based on the Windows versions

rashil2000: Clarified the note to reflect this better.

pratikpc: Further Git for example ships with OpenSSH 8.8 while this is a beta version of an older OpenSSH release.

rashil2000: Currently there is no way in Scoop to check OS versions, and pre-maturely exiting from pre-install scripts would result in a broken installation.

rashil2000: We already have a separate manifest for git-with-openssh, users can install that if they want.

niheaven: This is for Windows 7 and up (https://github.com/PowerShell/Win32-OpenSSH/wiki/Install-Win32-OpenSSH#install-win32-openssh-test-release)

From 8.1.0.0p1, Microsoft remove `This is a pre-release (non-production ready)` from release note, so I think it is stable enough for production.

pratikpc: @niheaven personally I see bundling the beta version of an outdated OpenSSH version (8.6) while the current is 8.8 to be not the best of ideas.

niheaven: Microsoft use OpenSSH 8.6p1 even in Windows 11's latest Insider Preview.

Issue #2948, opened 2021-11-14
vim-nightly daily updating task is not working


palopezv:
vim-nightly updating task stopped working around October 31. Neovim-nightly works Ok. What gives?

issaclin32:
The *Excavator* bot was blocked by Github because of too many requests. (too many packages are being updated)
This should be automatically solved in a few hours.

https://scoop.r15.ch/main/mud-20211114-160001.log
`vim-nightly: The remote server returned an error: (429) too many requests.`

palopezv:
@niheaven Still doesn't work, see https://scoop.r15.ch/main/mud-20211117-140001.log And BTW, the neovim-nightly is now borked due to #2960

run2cmd:
This still seems to be not solved. Yesterday we got VIM release 8.2.3692 but I do not see bucket was updated.

PR #2972, opened 2021-11-18
chore: Migration of nightlies to `Versions` bucket

niheaven:
Ref: https://github.com/ScoopInstaller/Versions/pull/318

Close https://github.com/ScoopInstaller/Main/issues/2948


rashil2000:
- docker-nightly
- ffmpeg-nightly
- ffmpeg-nightly-shared
- ffmpeg-nightly-vulkan (no longer available)
- ffmpeg-nightly-shared-vulkan (no longer available)
- glslang-nightly
- neovim-nightly
- jar2app

Let's migrate these too in this PR.

niheaven:
`jar2app` is a nightly app, but it doesn't need to be moved. I'll move all other ones.

rashil2000:
> `jar2app`

It was mentioned in https://github.com/ScoopInstaller/Extras/issues/2821#issue-490708605, that's why.

There hasn't been any commit since 2018 in its source - should we deprecate it instead?

(Also, what is a Mac OS tool doing in Scoop repos? 😅)

niheaven:
This was added in https://github.com/ScoopInstaller/Scoop/pull/3277, and I really don't know why...

rashil2000 (approved):

pratikpc:
@niheaven in the interest of not breaking workflows, I believe we should not move the nightlies which update via Scoop

We should probably deprecate this though
> ffmpeg-nightly-vulkan (no longer available)
> ffmpeg-nightly-shared-vulkan (no longer available)

niheaven:
We could keep them until scoop has an alias feature and redirect these manifests to versions' ones. Then after one or two months, delete them as this PR.

JanPokorny:
Scoop really ought to have something like `"moved_to": "versions/docker-nightly"` which would tell the CLI to rewire the installed package to the new manifest. You can actually do this manually pretty easily.

rashil2000:
I think this PR can be merged. The download links for all ffmpeg builds are dead anyway, and Neovim's wiki recommends using Versions bucket since a long time now:

image

rashil2000:
Also glslang-nightly's manifest is wrong - the app is 64bit but there's no architecture block

niheaven:
I'll resolve the conflicts.

PR #2978, opened 2021-11-18
mingw-nuwen: Add version 18.0

### LINKED ISSUE #2977: [Request] nuwen gcc
### Criteria

- [x] a non-GUI tool

### Information

**Name:** Nuwen's MinGW Distro

**Description:** MinGW distro with Boost, GCC etc.

**Homepage:** https://nuwen.net/mingw.html

**Download link(s):** https://nuwen.net/files/mingw/mingw-18.0-without-git.exe

**Some indication of popularity/repute:** It's pretty well known when it comes to C++ development on Windows.

pratikpc:
Closes #2977 

Need some help when it comes to shortcuts.

I can add a batch file which will load the environment for the distro but am not sure if we should add it.

Views?

    "shortcuts": [
        [
            "Nuwen's MinGW",
            "open_distro_window.bat"
        ]
    ],


rashil2000:
I think the name nuwen-mingw.json is sufficient. This distro contains a large number of tools and is not just limited to GCC:

image

pratikpc:
> The SSL connection could not be established, see inner exception.

We really need to fix this!

pratikpc:
@rashil2000 its primarily used for GCC though so that being in the name is pretty important IMHO

A few of the major utilities for example have links to GCC project

rashil2000:
The terms `mingw` and `gcc` together are redundant - MinGW _is_ GCC (for Windows). 

Let's drop `mingw` then?

niheaven:
Don't think this is GCC, this is just a MinGW Distro. Why not rename it to `nuwen-mingw`?

rashil2000:
That's what I was saying too - it's more popularly known as Nuwen MinGW.

pratikpc:
@niheaven mentioning GCC is pretty important too because the primary reason most people install this is for using GCC on Windows.

pratikpc:
@niheaven I personally prefer nuwen-mingw-gcc as a name

chawyehsu:
> Don't think this is GCC, this is just a MinGW Distro. Why not rename it to nuwen-mingw? Or mingw-nuwen?

Sorry, but current `gcc.json` using winlibs distro isn't even a GCC package. So you made something really bad in previous commits I think as I said.

chawyehsu:
I assume that you are inspired from my bucket's one, but if you really want to add this then I would suggest renaming it to `nuwen-mingw-without-git.json`.

pratikpc:
> I assume that you are inspired from my bucket's one,

This is true. It's probably the best name possible for a package like this.

> renaming it to nuwen-mingw-without-git.json

Would we really add nuwen's mingw with Git?

I presume with Scoop it is expected most users would install Git via the many Git packages present in Scoop

chawyehsu:
It doesn't mean you need to add it in this PR, instead it *follows* the original naming of the distro. 

What if someone in the future want to add the full one?

rashil2000:
> > Don't think this is GCC, this is just a MinGW Distro. Why not rename it to nuwen-mingw? Or mingw-nuwen?
> 
> Sorry, but current `gcc.json` using winlibs distro isn't even a GCC package. So you made something really bad in previous commits I think as I said.

@chawyehsu Yes, WinLibs is just another distro, like Nuwen's. We should probably rename that manifest to winlibs-mingw.json if we want to keep it here.

As for the one "true" GCC (without any 3rd party distros), the only way to download that is from the source of MinGW itself, that is https://sourceforge.net/projects/mingw-w64. There's just one problem - they offer an installer, because there are multiple options for toolchains - 
image

How we'll capture this choice in the manifest is something I don't understand.

rashil2000:
> I would suggest renaming it to `nuwen-mingw-without-git.json`.

Also, I still think nuwen-mingw.json is a better name for this PR. None of the other distros include Git, so we can omit the the `without-git` part right?

pratikpc:
> What if someone in the future wants to add the full one?

Then that could be named with-git instead?

chawyehsu:
No, I beg to differ. I tend to follow its original naming.

- mingw-18.0.exe `nuwen-mingw`
- mingw-18.0-without-git.exe `nuwen-mingw-without-git`

bucket/nuwen-gcc.json
```diff
@@ -0,0 +1,24 @@
+{
+    "homepage": "https://nuwen.net/mingw.html",
+    "description": "Nuwen's MinGW Distro with GCC, Boost etc.",
```
rashil2000, line 3:
```suggestion
    "description": "Nuwen's MinGW Distro with coreutils, 7zip and extra headers",
```

Also `nuwen-gcc.json` -> `nuwen-mingw-without-git.json` as per @chawyehsu's suggestion

bucket/nuwen-gcc.json
```diff
@@ -0,0 +1,24 @@
...
+    "license": "GPL-2.0-only",
+    "version": "11.2.0-18.0",
```
rashil2000, line 5:
```suggestion
    "version": "18.0",
```

18.0 is the version of the distro - otherwise you'll have to write the version of everything (the MinGW version is actually 9.00)

bucket/nuwen-gcc.json
```diff
@@ -0,0 +1,24 @@
...
+    "architecture": {
+        "64bit": {
+            "url": "https://nuwen.net/files/mingw/mingw-18.0-without-git.exe#/dl.7z",
+            "hash": "532aa9fc377fa8cb3985fc2fadf61da43f1c50bcc464719b8274dc2b2590d688"
+        }
+    },
+    "extract_dir": "MinGW",
+    "checkver": {
+        "re": "(?sm)currently contains <b>GCC (?<version>[\\d.]+)</b>.*?components-(?<components>[\\d.]+).7z",
```
rashil2000, line 14: Modify this accordingly

rashil2000 (changes requested):

niheaven:
> > Don't think this is GCC, this is just a MinGW Distro. Why not rename it to nuwen-mingw? Or mingw-nuwen?
> 
> Sorry, but current `gcc.json` using winlibs distro isn't even a GCC package. So you made something really bad in previous commits I think as I said.

Maintaining a pure GCC in Scoop is a hard job, and what is a PURE GCC in Windows de facto? What we picked is a minimal distro that could fit people's need, and if someone wants some additional utilities, he must choose different distro.

And I agree with @chawyehsu that MSYS2's best but maintaining is annoying. I try to write a MSYS2 packages updating script but wonder where it should go, in `checkver` or `update`? It has no relationship with Scoop and has a different update mechanism (check update, find supporting packages, find package URLs and hashes, etc.).

niheaven (approved):
LGTM

Issue #2933, opened 2021-11-11
[Request] Help us to update all MSYS packages to their latest ZSTD versions


pratikpc:
We are working on ZSTD support currently and it might get released in a while and by then, we hope all packages will be ready for release.

We need your help to update them all to their latest ZSTD versions

niheaven:
Emm, zstd support is only in `develop`, so this should wait for a while.

pratikpc:
@niheaven yeah

I think we can get the PRs ready by then so we can help release ZSTD supported packages while we can work on and fix the PRs.

Might make the release easier

Thoughts?

niheaven:
And what's your experience with `develop`'s new commits? Any problem or error?

pratikpc:
ZSTD support works for me

rashil2000:
ZSTD works for me too (tried with the PR for openssh).

Is there a way to test the multiple-URLs in autoupdate functionality?

niheaven:
Use lua in main bucket，or some manifests that I attached in that PR.

rashil2000:
I'll past my reply from #2967 here:

> I agree that this might be a breaking change - for that we can redirect users to the Versions bucket as @niheaven said.
> 
> However, I strongly believe Scoop should not be used to download MSYS2 packages - from a maintenance perspective, it's impossible to keep track of individual versions of the 15-20 different download URLs there. Pacman (widely touted as the most respected package manager out there) in MSYS2 handles dependency trees beautifully; if people need MSYS2 tools they should use it.
> 
> It's a similar argument to why Scoop isn't (and shouldn't be) used to download/manage NPM packages, or PIP packages - these are far better managed by their respective package managers.
> 
> If applications provide standalone builds (like this WinLibs does), we can use it freely and the manifest will even get autoupdated.
> 
> One more problem - currently we have around 10 different manifests in Main which individually download the MSYS2 Runtime - this leads to unnecessary bloat. It might also cause inter-compatibility if someone uses tools from different runtime versions - for example openssh and diffutils.

rashil2000:
List of packages using MSYS2 repos:
- diffutils
- openssh
- gdb

List of packages using MSYS repos (these should ideally be deprecated or updated to use other sources - it's a security risk to keep 8-10 year old packages in Main):
- telnet (who still uses this in 2021???)
- tar
- findutils
- coreutils
- gzip
- patch

rashil2000:
[...]

PR #2996, opened 2021-11-22
Add mingw

rashil2000:
Relates to https://github.com/ScoopInstaller/Main/issues/2933#issuecomment-971566422


niheaven:
Then who will be the GCC?

rashil2000:
There's no standalone GCC, as you said in your comment in the other PR.

niheaven:
Yes, what I mean is, if user want GCC or only know GCC, should we provide a distro of GCC that fits user's need?

An alias or group selection may be better, but there isn't such feature. So I prefer left current one as GCC and change its name after the feature is added.

rashil2000:
Okay, I renamed mingw.json to gcc.json, and added a note 👍

niheaven:
I don't think such an outdated distro should be the default. But since I'm not target user of GCC, please wait other maintainers' feedback.

The one last thing I concerned is, please place build vendor behind app name, e.g., mingw-winlibs.

rashil2000:
> I don't think such an outdated distro should be the default.

It is the official source of mingw-w64, what other option do we have? Others are all third party (including MSYS2).

niheaven:
Don't confuse MinGW-w64 and GCC. MinGW-w64 is NOT GCC, it is just a C library that make GCC usable in Windows. Other distro use GCC from FSF and MinGW-w64 as underground C library.

rashil2000:
Yes, I'm not saying they are the same thing. But there's no GCC without MinGW-w64 (on Windows).

niheaven:
People is using GCC, not MinGW, so make GCC updated. `8.2.0` is far more outdated. (MinGW 9.0.0 is latest version, though)

rashil2000:
There's no official source for GCC that we can use in the manifest (the gcc.gnu.org website does not provide binaries).

The runtime library that's used to compile GCC is made by mingw-w64.org, and they provide binaries, so I used that in the manifest.

issaclin32:
I vote up for this solution. In regards to [Chawye's comment](https://github.com/ScoopInstaller/Main/pull/2967#issuecomment-971458707).

If the end-user **naively** installs *gcc* and compiles something with *gcc*, the MinGW-w64 is less likely to cause problems. However, we should **explicitly** point out that there are newer versions available in `notes`.

Such as:
```
You are installing the Mingw-w64 implementation of GCC (version 8).
This implementation has a better compatibility, but may not support newer C/C++ standards.
For GCC version 11, install 'mingw-winlibs.json'
```

issaclin32 (approved):

tech189:
@rashil2000 - mingw-winlibs checkver is broken, regex101 says it detects catastrophic backtracking but I couldn't work out what you were trying to match, could you fix it please?

rashil2000:
The URLs will need to be updated. checkver can be:
```
GCC ([\\d.]+) \\+ MinGW-w64 ([\\d.]+) \\(MSVCRT\\) - release ([\\d.]+)
```

issaclin32:
> @rashil2000 - mingw-winlibs checkver is broken, regex101 says it detects catastrophic backtracking but I couldn't work out what you were trying to match, could you fix it please?

Fixed in https://github.com/ScoopInstaller/Main/commit/af540077b0dad8734983e977e5847c20d0bf156e

PR #3002, opened 2021-11-24
offlineinsiderenroll: Add version 2.6.2

sitiom:


goostleek:
Hi @sitiom. Thx for the contribution.

One comment from my side though. When we run the installed `offlineinsiderenroll` as a regular (non Administrator) user  then it says that:
```
=====================================================
This script needs to be executed as an administrator.
=====================================================

Press any key to continue . . . 

```
What I suggest is to provide a convenient way to run the script as an Administrator from the CLI as well. We can do that by extending the manifest with the [`sudo`](https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/sudo.json) dependency and [a `note` property](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties) how to run the installed script properly, I mean:
```
...
notes: [
    "Run the script as an Administrator using:",
    "sudo offlineinsiderenroll"
]
...
```

WDYT?

rashil2000:
We should use [gsudo](https://github.com/ScoopInstaller/Main/blob/master/bucket/gsudo.json) instead of sudo as a dependency - gsudo is much more robust and regularly maintained.

sitiom:
> What I suggest is to provide a convenient way to run the script as an Administrator from the CLI as well. We can do that by extending the manifest with the [`sudo`](https://github.com/ScoopInstaller/Main/blob/3d8c2e1c5eb8adc8fc9627165d0e796cad846876/bucket/sudo.json) dependency and [a `notes` property]

It's better to put `gsudo` in the `suggest` property than in the `depends` since you don't really need it to run the script as admin (Just run the console as admin first). Added a note as well now.

PR #3004, opened 2021-11-24
update pandoc manifest

cderv:
This is juts an update for the Pandoc manifest so that 2.16.2 is used. 

Current 2.16 is already two versions behind. 

Thanks

### DIFF
--- a/bucket/pandoc.json
+++ b/bucket/pandoc.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.16",
+    "version": "2.16.2",
     "description": "Universal markup converter",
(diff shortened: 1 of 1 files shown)


goostleek (changes requested):
Thx @cderv for the contribution.

I see you manually updated the version to the latest version released. Scoop has built-in mechanics to automatically update its manifests if they are written in a proper way. Please see the [Autoupdate feature documentation](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#special-cases) and adjust the manifest accordingly so we don't have to update the manifest manually anymore.

PR #3008, opened 2021-11-25
get-iplayer: Add version 3.27.1

issaclin32:
Supersedes https://github.com/ScoopInstaller/Extras/pull/3812

This adds [get_iplayer](https://github.com/get-iplayer/get_iplayer/), a tool for downloading TV and radio programmes from BBC iPlayer and BBC Sounds.

NOTES:
* The package has `shortcut` because the installer of **get_iplayer** creates shortcuts too.
* The "profile directory" [can only be specified at runtime](https://github.com/get-iplayer/get_iplayer/wiki/prefs). Therefore I use alternative batch files (`scoop_get_iplayer.cmd`, etc.) to do that.
* I tested the app by `get_iplayer --pid=m0011rxw --type=radio`
* See [previous PR](https://github.com/ScoopInstaller/Extras/pull/3812) for more discussions/details.

### DIFF
--- a/bucket/get-iplayer.json
+++ b/bucket/get-iplayer.json
@@ -3,10 +3,6 @@
     "description": "Utility for downloading TV and radio programmes from BBC iPlayer and BBC Sounds",
     "homepage": "https://github.com/get-iplayer/get_iplayer",
     "license": "GPL-3.0-only",
-    "notes": [
-        "Default Download directories (for TV and radio) are set to",
-        "'$dir\\tv' and '$dir\\radio'"
-    ],
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


tech189:
Looks good to me! Just one question: because every command uses the profile that the manifest sets up, the `--output` argument no longer works. But that's okay because you can use `--prefs-add` to change your preferred output directory. However, when an update is installed, will scoop change the prefs again back to the default folders in the manifest? If so, maybe you could run a check not to change the prefs if they're set already?

issaclin32:
> If so, maybe you could run a check not to change the prefs if they're set already?

Done.

rashil2000 (approved):

PR #3010, opened 2021-11-26
hugo: Update to version 0.89.4

eallion:
Closes #XXXX

Relates to #XXXX


rashil2000:
Update the hashes too

PR #3013, opened 2021-11-27
k9s: update to version 0.25.7

Slach:
Look like auto updater not works now?


rashil2000:
Auto updater is currently under maintenance

PR #3014, opened 2021-11-28
ffmpeg: Remove .ps1 shims after installation

### LINKED ISSUE #2611: ffmpeg.ps1 causes issues
Calling `ffmpeg.exe` shim doesn't cause any issue, but calling `ffmpeg.ps1` (default behaviour when calling from PowerShell 7) causes issues.

Running the following in PowerShell: `ffmpeg -i i.avi -bsf:v mpeg4_unpack_bframes -codec copy o.avi` causes this error:
```
[NULL @ 00000235ef2ebd00] Unable to find a suitable output format for 'mpeg4_unpack_bframes'
mpeg4_unpack_bframes: Invalid argument
```

I need to explicitly call `ffmpeg.exe` from PowerShell to avoid issues.

Check this link for more detailed discussion: https://trac.ffmpeg.org/ticket/9398

issaclin32:
close #2611 
close https://github.com/ScoopInstaller/Scoop/issues/4473

This is a workaround for the issue mentioned in https://github.com/ScoopInstaller/Main/issues/2611.
(ffmpeg has problem handling arguments containing `:` if `ffmpeg.ps1` is involved)

Also see upstream issue (closed by the maintainer of *ffmpeg*) : https://trac.ffmpeg.org/ticket/9398

### DIFF
--- a/bucket/ffmpeg-nightly-shared-vulkan.json
+++ b/bucket/ffmpeg-nightly-shared-vulkan.json
@@ -10,6 +10,10 @@
             "extract_dir": "ffmpeg-N-101857-g0617e578a3-win64-gpl-shared-vulkan"
         }
     },
+    "post_install": [
+        "# Workaround for https://github.com/ScoopInstaller/Main/issues/2611",
+        "Remove-Item \"$scoopdir\\shims\\ffmpeg.ps1\", \"$scoopdir\\shims\\ffplay.ps1\", \"$scoopdir\\shims\\ffprobe.ps1\""
+    ],
     "bin": [
         "bin\\ffmpeg.exe",
         "bin\\ffplay.exe",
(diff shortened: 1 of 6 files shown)


rashil2000:
I have no idea why the .ps1 shims are created in first place - I've only encountered problems with it.

But it seems for the time being we have to fix it this way.

rashil2000 (approved):

issaclin32:
Actually .ps1 shims [exist even before the shim-exes](https://github.com/ScoopInstaller/Scoop/commit/5c52166fcd813aaee0865e50f516b09d696bf9c7)

I am not sure why .ps1 shims still remain for exes. Probably because it is easier to debug for the *shim* mechanism? I cannot tell because there are not too many info provided in the commits.

rashil2000:
My guess is that scoop was initially probably only usable from PowerShell itself, so the .ps1 shims were created. Later they created .cmd shims so that it can be used from other shells. And then after that started using a compiled .exe shim so that programs with hardcoded extensions could work. But the .ps1 shims remained.

We can probably get rid of them, now that we know .exe will work from any shell.

PR #3016, opened 2021-11-28
win-acme: update to version 2.1.20.1185 & fix autoupdate field

delassiops:
Update manifest version and fix auto-update field ( revert to matchHead version variable)

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/win-acme.json
+++ b/bucket/win-acme.json
@@ -1,16 +1,16 @@
 {
-    "version": "2.1.19.1142",
+    "version": "2.1.20.1185",
     "description": "A Simple ACME Client",
     "homepage": "https://www.win-acme.com",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/win-acme/win-acme/releases/download/v2.1.19.1142/win-acme.v2.1.19.1142.x64.trimmed.zip",
-            "hash": "9b25c9ffa8ed4ae6034661d5e53e6b19a465fc94cc6d948c5a329edaf6037d43"
+            "url": "https://github.com/win-acme/win-acme/releases/download/v2.1.20.1/win-acme.v2.1.20.1185.x64.trimmed.zip",
+            "hash": "c2da8ad05f8e9cdc433619340a4fa48b4c5c49234bcf06330c4e5065b428f787"
         },
         "32bit": {
-            "url": "https://github.com/win-acme/win-acme/releases/download/v2.1.19.1142/win-acme.v2.1.19.1142.x86.trimmed.zip",
-            "hash": "0bb2079636427bfe365db783e8b642fa1577dc243c68248e1244ed9888d60149"
+            "url": "https://github.com/win-acme/win-acme/releases/download/v2.1.20.1/win-acme.v2.1.20.1185.x86.trimmed.zip",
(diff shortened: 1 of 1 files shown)


rashil2000:
@delassiops can you fix this? https://github.com/ScoopInstaller/Main/runs/4346458400?check_suite_focus=true

delassiops:
verified: checkver.ps1 -App win-acme is ok

rashil2000:
I was talking about autoupdate - it is failing.

delassiops:
sorry i will do, i only try checkver.ps1

delassiops:
Hi, 

when looking file: [https://github.com/win-acme/win-acme.github.io/blob/master/_config.yml](url) 

the version variable  **releasename**: 2.1.20.1 and **releasebuild**: 2.1.20.1185  are defined manually and used in the download url,

this make impossible to automate script, or add the possibility to retrieve multiple version from checkver.ps1

thanks

rashil2000:
Can you open an issue in their repo and ask the author how the build number is obtained?

delassiops:
> Can you open an issue in their repo and ask the author how the build number is obtained?

after checking repo, almost it's manually,

also the problem is in the download url: the download folder & file name doesn't match

https://github.com/win-acme/win-acme/releases/download/v**2.1.20.1**/win-acme.v**2.1.20.1185**.x64.trimmed.zip 

it's can't replaced by only one variable, maybe two will solve the problem

thanks,

delassiops:
i have opened an issue to fix from origine  : https://github.com/win-acme/win-acme/issues/2003

rashil2000:
Great, thanks!

WouterTinus:
The part of the download url that you're trying to guess is based on the tag given to the commit that is released, which is set manually. The full version build number and file names on the other hand are generated by build automation. I will try to remember to be more consistent now that I know that some people depend on it, but I cannot guarantee that I won't forget or someone else might publish a release in the future that doesn't know about this. 

> or add the possibility to retrieve multiple version from checkver.ps1
> it's can't replaced by only one variable, maybe two will solve the problem

This seems to be the only real solution. If you can get the tag and the build number from Github, you can reliably produce the download link. 

Alternatively you could maybe just get the *full* download link from either Github or even better https://www.win-acme.com/. That way even if we'd change to a whole different platform it could keep working.

rashil2000:
We can get as many variables as we want from checkver, we just need to use a custom regex to capture those named variables.

delassiops:
Hi, this code make possible to retrieve tag and build number from www.win-acme.com that it published under repo:

https://github.com/win-acme/win-acme.github.io

we can check file: _config.yml, and get value of releasetag: 2.1.20.1, releasebuild: 2.1.20.1185

`   "checkver": {
        "url": "https://raw.githubusercontent.com/win-acme/win-acme.github.io/master/_config.yml",
		"regex": "releasename: ([\\d.]+)",
                "regex": "releasetag: ([\\d.]+)"
    }`

thanks,

rashil2000:
You can't have 2 fields with the same name. 

Refer to this manifest, where it captures the `date` and `sha` variables, and uses them as `matchDate` and `matchSha` - https://github.com/ScoopInstaller/Extras/blob/master/bucket/synfig-studio.json

delassiops:
autoupdate is fixed, how to make new edit

delassiops:
i found an issue on file generated by "Writing updated win-acme manifest"

PR #3018, opened 2021-11-28
Add diffsitter 0.6.7

Slach:
`diffsitter` creates semantically meaningful diffs that ignore formatting differences like spacing. It does so by computing a diff on the AST (abstract syntax tree) of a file rather than computing the diff on the text contents of the file.


rashil2000 (approved):

stevehipwell (approved):
It's a shame the project doesn't provide a hash for Windows but this looks good to me.

PR #3024, opened 2021-11-29
Update postgresql to version 14.1

CaselIT:
Update postgresql to version 14.1.

I've also changed the checkver regex since the current one no longer matches the website.


CaselIT:
thanks!

PR #3026, opened 2021-11-30
env_add_path adjust

HUMORCE:
see also: https://github.com/ScoopInstaller/Main/pull/1901


niheaven:
This may not be user friendly, so I prefer fixed in scoop's env_add_path function.

HUMORCE:
the symbol '\' at the end of path should also be removed

PR #3039, opened 2021-12-02
yt-dlp: update to version 2021.12.01

jdambron:
yt-dlp: update to version 2021.12.01

### DIFF
--- a/bucket/yt-dlp.json
+++ b/bucket/yt-dlp.json
@@ -12,7 +12,7 @@
             "hash": "sha512:1C9C0D2A0D98BDA5F92D2CFDAB0B2C3FF81715B4C390E256FE27999ADF54AB44E7AAF53B12511E4B5BEC9B358291593F531A6257D5D10762B4CD225E506F306A"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


bucket/yt-dlp.json
```diff
@@ -24,7 +24,7 @@
                 "url": "https://github.com/yt-dlp/yt-dlp/releases/download/$version/yt-dlp.exe"
             },
             "32bit": {
-                "url": "https://github.com/yt-dlp/yt-dlp/releases/download/$version/yt-dlp_x86.exe#/yt-dlp.exe"
```
rashil2000, line 27: Why did you change this?

rashil2000 (changes requested):

bucket/yt-dlp.json
jdambron: @rashil2000 It seems they changed the URL in the last version. I replaced with the correct one.

rashil2000: No, you need to keep the `#/yt-dlp.exe`, it's not a part of the URL, it is for Scoop to rename the file.

jdambron:
@rashil2000  Thank you 👍

PR #3046, opened 2021-12-04
gopher360: add manifest

FriendlyNeighborhoodShane:
It would've been nice to be able to persist config.ini, but that file is created upon first launch, and as far as I can understand, there is no way to persist files that don't exist before installation.

### DIFF
--- a/bucket/gopher360.json
+++ b/bucket/gopher360.json
@@ -1,0 +1,20 @@
+{
+    "version": "0.989",
+    "description": "App to turn a controller into a mouse and keyboard",
+    "homepage": "https://github.com/Tylemagne/Gopher360",
+    "license": "GPL-3.0-only",
+    "url": "https://github.com/Tylemagne/Gopher360/releases/download/v0.989/Gopher.exe#/gopher360.exe",
+    "hash": "13b45c1487d4920df5acc5e69ab4d827e69f086bdf0e666244038856e5ac8d4d",
+    "shortcuts": [
+        [
+            "gopher360.exe",
+            "Gopher360"
+        ]
+    ],
+    "checkver": {
+        "github": "https://github.com/Tylemagne/Gopher360"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Tylemagne/Gopher360/releases/download/v$version/Gopher.exe#/gopher360.exe"
+    }
+}


rashil2000:
It is possible, see https://github.com/ScoopInstaller/Main/blob/87a6c5f0bb98ff3524e2b89957c07f96d5b4bfb1/bucket/mpxplay.json#L8

FriendlyNeighborhoodShane:
Ahhhh, thanks for the tip, I'll amend it in a moment.

(aside: that seems very roundabout. Why don't we just have a `persist_file`??)

FriendlyNeighborhoodShane:
Hmmm... yeah, won't do. The program autogenerates the config upon first launch only if the file doesn't exist. An empty file just makes it do nothing.

The docs say it regenerates the config if the existing file is messed up. Wonder if I can make a 'default' mangled config to do this.

rashil2000:
Why not put something meaningful there, as a default?

FriendlyNeighborhoodShane:
But that creates an extra maintainance burden of having to keep a default. (I know, for a program that hasn't been updated in 4 years??! But still, I'd have preferred to not have to do this "duplication" of data.) Regardless, that might be the only viable option to take here.

Though the blame lands squarely on the program here for being so dumb. It should just use sensible defaults itself upon reading an empty config.

rashil2000:
The other option is to not have a persist at all - given that the repo hasn't had a release for long, that is viable too.

FriendlyNeighborhoodShane:
I think I'm satisfied with this, for now. Might PR/fork upstream to fix this (+ other issues) in the future.

rashil2000 (approved):

HUMORCE (changes requested):
- missing `bin`, cli program does not need `shortcuts`.
- sort propetrys by repo requirements.

FriendlyNeighborhoodShane:
> * sort propetrys by repo requirements.

Addressed.

> * missing `bin`, cli program does not need `shortcuts`.

This isn't exactly a CLI program.

It launches its own terminal if you launch the executable or its shortcut. You _can_ launch it inside your own terminal, but it sends escape codes that will mangle everything other than CMD (tried on git-bash+cmder). It takes no arguments, so there's no point to launching a terminal and having to type out its name. Moreover, launching it from the terminal makes it choose the current working directory (often home) for its `config.ini`. Launching it from the shortcut makes it use its own scoop directory, which is what we want. (Arguably, some of these are problems that should be fixed upstream.)

So yeah, it behaves more like a GUI app that happens to use a terminal as its interface.

HUMORCE:
i think this app should be move to [extras bucket](https://github.com/ScoopInstaller/Extras).

@rashil2000

rashil2000:
Okay, since it needs to be launched standalone, we can have it in Extras.

FriendlyNeighborhoodShane:
Continued in https://github.com/ScoopInstaller/Extras/pull/7475

PR #3062, opened 2021-12-08
docker: Update to version 20.10.11

### LINKED ISSUE #3058: Discussion: Installing Docker (CLI & engine)
# Docker on Windows

In the last years, the tooling around Docker on Windows has evolved a lot. It appears that all the current packages in Scoop for installing Docker ([`docker`](https://github.com/ScoopInstaller/Main/blob/master/bucket/docker.json) and [`docker-nightly`](https://github.com/ScoopInstaller/Versions/blob/master/bucket/docker-nightly.json)) are obsolete. The [`docker-toolbox`](https://github.com/docker-archive/toolbox) project, as well as [`docker-machine`](https://github.com/docker/machine) and [`lcow`](https://github.com/linuxkit/lcow) are deprecated and archived. I wasn't able to make either work, even though it might be a problem on my side.

The current "go-to" official way of installing Docker on Windows is [Docker Desktop](https://www.docker.com/products/docker-desktop) -- a huge monolithic GUI app with a proprietary licence forbidding its free use in companies with >250 employees. Definitely not the "Scoop way" of doing things.

## Components of Docker

Now, when I talk about Docker on Windows, there are actually three components I have in mind -- Docker CLI (`docker.exe` that can connect to any backend, local or remote), Docker Engine for Windows Containers, and Docker Engine for Linux Containers. You might think that a developer tool like Docker would extensively document installation of all the parts, quite the opposite is true -- the official documentation focuses on "just use Docker Desktop".

### Docker CLI

There is a project https://github.com/StefanScherer/docker-cli-builder, which builds `docker.exe` CLI from the latest sources. Should be easy to add into scoop.

### Docker Engine for Windows Containers

[The official documentation actually shows a way of installing this engine](https://docs.docker.com/engine/install/binaries/#install-server-and-client-binaries-on-windows) -- and it appears it consists only of unzipping it in a folder and registering and exe as a service -- also should be easy to add into scoop.

### Docker Engine for Linux Containers

The most interesting and the hardest part. Docker on Windows used to use VirtualBox in the past, but it gradually switched to Hyper-V and later WSL2 as it became available. (You might be aware that WSL2 uses Hyper-V internally. However, the WSL2 Docker backend has better performance and furthermore it runs even on Home editions of Windows where Hyper-V is not available.) WSL2 support in Docker Desktop is proprietary, even though it is known how it works -- it creates its own WSL2 distro where it runs the Linux version of Docker Engine.

There are two main possibilities (I can think of) of avoiding Docker Desktop here:

1. Install Docker Egine inside WSL and use some painful networking shenanigans to use it from the Docker CLI on Windows, [as described by Nicolas Luis on DEV.TO](https://dev.to/_nicolas_louis_/how-to-run-docker-on-windows-without-docker-desktop-hik).
2. Install [`minikube`](https://github.com/ScoopInstaller/Main/blob/master/bucket/minikube.json) which already exists as a Scoop manifest. Why Kubernetes, you might ask? As Kubernetes uses Docker Engine internally, Minikube packs one too. Minikube also supports [pretty much anything _except WSL2_](https://minikube.sigs.k8s.io/docs/drivers/) as a backend and allows exposing its internal Docker daemon using `minikube docker-env`. But as you might already guess, mounting FS and networking is complicated using this setup -- more in [this article from ITNEXT](https://itnext.io/goodbye-docker-desktop-hello-minikube-3649f2a1c469).

## What I propose

_(And by "propose" I mean "let's discuss", I don't expect anyone to do this for me.)_

- Remove current obsolete `docker` and `docker-nightly`
- Add `docker-cli`: https://github.com/StefanScherer/docker-cli-builder
- Add `docker-engine-windows`: https://download.docker.com/win/static/stable/x86_64/ (instructions: https://docs.docker.com/engine/install/binaries/#install-server-and-client-binaries-on-windows)
- Create a convenience CLI installer that goes through the process in the [DEV.TO article](https://dev.to/_nicolas_louis_/how-to-run-docker-on-windows-without-docker-desktop-hik) and basically mimics Docker Desktop -- creates a new WSL2 distro, configures it for Docker, and configures the local Docker CLI to work with it. (Make it available as `wsl-docker-installer` or something like that.)
- Finally, create a meta-manifest `docker` that installs all of the above in a seamless way. Basically the "Docker Desktop experience", but The Scoop Way™ (Yeah I know automatically creating stuff in WSL2 is not very scoopy but what can you do...)

I am interested in hearing what the community thinks of this approach. It's a shame that Scoop doesn't pack a working Docker installer, since Docker is one of the most important developer tools nowadays -- and there are ways to fix this.

rashil2000:
Closes #3058

The deprecation of `lcow` and `docker-nightly` is being handled in #2972.

/cc @JanPokorny

### DIFF
--- a/bucket/docker.json
+++ b/bucket/docker.json
@@ -27,7 +27,7 @@
     "notes": [
         "The 'dockerd' binary here only supports running Windows containers.",
         "However it is possible to connect to existing Linux containers using the 'docker' binary",
-        "To register Docker as service, run `dockerd --register-service`",
+        "To register Docker as a service, run `dockerd --register-service`",
         "Similarly, to unregister, run `dockerd --unregister-service`"
     ]
 }
--- a/bucket/docker.json
+++ b/bucket/docker.json
@@ -1,6 +1,6 @@
 {
     "version": "20.10.11",
-    "description": "Docker is an open platform for developing, shipping, and running applications. Docker enables you to separate your applications from your infrastructure so you can deliver software quickly.",
+    "description": "Docker CLI & Docker Engine for Windows containers. Docker is an open platform for developing, shipping, and running applications.",
     "homepage": "https://docs.docker.com/engine",
     "license": "Apache-2.0",
     "architecture": {
@@ -26,6 +26,7 @@
     },
     "notes": [
         "The 'dockerd' binary here only supports running Windows containers.",
+        "However it is possible to connect to existing Linux containers using the 'docker' binary",
         "To register Docker as service, run `dockerd --register-service`",
         "Similarly, to unregister, run `dockerd --unregister-service`"
     ]
--- a/bucket/docker.json
+++ b/bucket/docker.json
@@ -1,40 +1,32 @@
 {
-    "version": "19.03.1",
+    "version": "20.10.11",
     "description": "Docker is an open platform for developing, shipping, and running applications. Docker enables you to separate your applications from your infrastructure so you can deliver software quickly.",
-    "homepage": "https://docker.com",
+    "homepage": "https://docs.docker.com/engine",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/docker/toolbox/releases/download/v19.03.1/DockerToolbox-19.03.1.exe",
-            "hash": "bf39c167f1c2bfc18651c5ebcbc2deb044119dc527f49f3b0c1012c4fe8e9768"
+            "url": "https://download.docker.com/win/static/stable/x86_64/docker-20.10.11.zip",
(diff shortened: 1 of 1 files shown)


JanPokorny:
I'd tweak the description a bit. It needs to say what exactly does this install (since the Docker ecosystem is a clusterfun as shown earlier).

My proposal for description: "Docker CLI & Docker Engine for Windows containers."

My proposal to add into notes: "The CLI can connect to a Linux container backend, if you have one. It is not included in this package."

This should clear any confusion for people who do `scoop search docker` and expect to find Docker Desktop.

JanPokorny:
Oh, and a minor nitpick 😄 "register as **a** service"

issaclin32 (approved):

PR #3074, opened 2021-12-11
libxml2: Add version 2.9.3

issaclin32:
[Libxml2](http://xmlsoft.org/) is the XML C parser and toolkit developed for the Gnome project. (but usable outside of the Gnome platform)

### DIFF
--- a/bucket/libxml2.json
+++ b/bucket/libxml2.json
@@ -14,8 +14,8 @@
         }
     },
     "bin": [
-        "xmlcatalog.exe",
-        "xmllog.exe"
+        "bin\\xmlcatalog.exe",
+        "bin\\xmllog.exe"
     ],
     "checkver": {
         "url": "https://www.zlatkovic.com/pub/libxml/64bit/",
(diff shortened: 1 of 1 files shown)


rashil2000 (approved):

rashil2000:
Any particular reason for `libxml2.json` instead of `libxml.json`?

issaclin32:
> Any particular reason for `libxml2.json` instead of `libxml.json`?

Because it is called **libxml2** everywhere (including the logo), and we install it by `apt-get install libxml2` or `brew install libxml2` under UNIX systems. I think it is easier for users to find it with the name **libxml2**.



There is actually a package called **libxml**, but it is only kept for backward compatibility.

http://www.xmlsoft.org/FAQ.html
>I see libxml and libxml2 releases, which one should I install ?
If you are not constrained by backward compatibility issues with existing applications, install libxml2 only
If you are not doing development, you can safely install both. Usually the packages libxml and libxml2 are compatible (this is not the case for development packages).
If you are a developer and your system provides separate packaging for shared libraries and the development components, it is possible to install libxml and libxml2, and also libxml-devel and libxml2-devel too for libxml2 >= 2.3.0
If you are developing a new application, please develop against libxml2(-devel)

rashil2000:
Okay, LGTM!

PR #3080, opened 2021-12-13
chore: Update some very old apps

rashil2000:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/grep.json
+++ b/bucket/grep.json
@@ -16,6 +16,7 @@
     ],
     "bin": "bin\\grep.exe",
     "checkver": {
+        "url": "https://nuwen.net/mingw.html",
         "regex": "(?sm)>grep (?<version>[\\d.]+)</a>.*?components-(?<components>[\\d.]+).7z"
     },
     "autoupdate": {
--- a/bucket/grep.json
+++ b/bucket/grep.json
@@ -1,19 +1,29 @@
 {
-    "version": "2.5.4",
+    "version": "3.7",
     "description": "Print lines matching a pattern.",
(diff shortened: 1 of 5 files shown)


issaclin32 (approved):

niheaven:
Why download such a big file (~90MB) while I only need `grep` or `sed`?

rashil2000:
What's the alternative then?

Otherwise the apps will remain on very old versions.

niheaven:
Take them from Chocolatey:

https://community.chocolatey.org/packages/sed Download: https://packages.chocolatey.org/sed.4.8.nupkg

https://community.chocolatey.org/packages/grep Download: https://packages.chocolatey.org/grep.3.7.nupkg

rashil2000:
Okay, will make PR

niheaven:
You could get `make` from Chocolatey, too.

rashil2000:
Make source is standalone, small package, I don't think it's required

niheaven:
If the source is SF, sha1 could be extracted from webpage, and it sounds good.

Chocolatey's file has native sha256, extracted by `$sha256.*?$basename` (this should works).

niheaven:
The PR https://github.com/ScoopInstaller/Scoop/pull/3249 tries to use Chocolatey as distribution server and checkver on it. Maybe it is worth being merged.

rashil2000:
Oh, that looks interesting, didn't know about it 🚀

rasa:
We could have a process unzip https://nuwen.net/files/mingw/components-18.0.7z and push the components we want into https://github.com/ScoopInstaller/Binary .

rashil2000:
But how will they autoupdate?

PR #3082, opened 2021-12-13
sed, grep: Use choco source

rashil2000:
https://github.com/ScoopInstaller/Main/pull/3080#issuecomment-992718380


niheaven (approved):

rasa:
Just an FYI, we asked the Chocolatey folks if we could snarf their files, and they didn't reply. We may want to ask again, to be courteous.

rashil2000:
Was that done privately or openly? Do you have a link to the conversation?

PR #3083, opened 2021-12-13
ncspot: Update to version 0.9.3

itytophile:
Update from `0.9.0` to `0.9.3`. Add `x86_64` to urls.

### DIFF
--- a/bucket/ncspot.json
+++ b/bucket/ncspot.json
@@ -1,12 +1,12 @@
 {
-    "version": "0.9.0",
+    "version": "0.9.3",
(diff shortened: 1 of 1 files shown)


rashil2000:
Can you fix the hash extraction?

itytophile:
Hi, which command do you use to test if the manifest is correct? I replaced `"url": "$urlNoExt.sha256"` by `"url": "$baseurl/ncspot-v$version-windows-x86_64.sha256"` but I don't really know if this works. Thanks.

rashil2000:
You can make the commit here, and then comment `/verify` like I did

PR #3086, opened 2021-12-14
gcc: Add version 11.2.0

rashil2000:
Use standalone gcc package

### DIFF
--- a/bucket/gcc.json
+++ b/bucket/gcc.json
@@ -1,0 +1,41 @@
+{
+    "version": "11.2.0",
+    "description": "GNU Compiler Collection",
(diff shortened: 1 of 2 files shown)


niheaven:
Why not just copy `mingw-nuwen` manifest? It is no-git version.

rashil2000:
> Why not just copy mingw-nuwen manifest? It is no-git version.

Yes, but it contains many other stuff - GDB, make, 7zip, coreutils

This is just pure GCC, nothing else.

For people who want all things, they can install mingw-nuwen manifest

niheaven (approved):
LGTM, if there isn't a more proper GCC dist, this should be the best.

PR #3091, opened 2021-12-14
Update pwsh to 7.2.1

couleurm:


rashil2000:
/verify

Btw Scoop runs its autoupdater every 4 hours, so this wasn't required

PR #3096, opened 2021-12-15
es: Add version 1.1.0.20

niheaven:
From https://github.com/ScoopInstaller/Extras/pull/2710 (@niboan)

Close https://github.com/ScoopInstaller/Extras/pull/2710


rashil2000:
The manifest is fine, but I think the name `es` is too vague. Why not call it everything-cli?

niheaven:
`bin` name is es, and I think user will search es for this utility, not everything-cli.

rashil2000:
The command `scoop search es` is too non-specific and will show too many results.

`es` could mean anything - ES File Manager, ESLint tool, etc.

`everything-cli` is clear and unambiguous.

PR #3106, opened 2021-12-19
remove oraclejdk references

### LINKED ISSUE #2927: Remove all references to oraclejdk from main
There are a lot of packages where suggests references java/oraclejdk but java/oraclejdk has been removed due to issues.

It would be nice if someone could create a PR with all references removed.

amreus:
Closes #2927

### DIFF
--- a/bucket/ammonite.json
+++ b/bucket/ammonite.json
@@ -5,7 +5,6 @@
     "license": "MIT",
     "suggest": {
         "JDK": [
-            "java/oraclejdk",
             "java/openjdk"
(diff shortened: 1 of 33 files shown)


rashil2000 (changes requested):
If the array contains only one item, write it as a string

amreus:
> If the array contains only one item, write it as a string

I can change it.  Is that a convention? Is it documented somewhere?

rashil2000:
It is a convention. Not documented, yet.

PR #3128, opened 2021-12-27
schemacrawler: Add version 16.16.5

sualeh:


tech189 (changes requested):
Hi,
Could you remove the pdfbox commit in the PR please?

It installs okay for me but the checks are failing because there is no final line with whitespace.

Could you also re-order the sections as per the [contributing guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md#for-scoop-buckets)?

Another issue is the `post_install`, I don't think it's necessary to print out the version and information, usually this section is for performing tasks (like modifying downloaded files) after the main installation has completed.

sualeh:
Thanks, @tech189 - let me redo this PR from scratch.

PR #3132, opened 2021-12-28
add termshark 2.3.0, TUI ispired by Wireshark

Slach:
fix https://github.com/ScoopInstaller/Extras/pull/5420#issuecomment-1001937746

PR #3133, opened 2021-12-28
Add termshark 2.3.0

Slach:
fixes from https://github.com/ScoopInstaller/Main/pull/3132#issuecomment-1002111786
and from https://github.com/ScoopInstaller/Extras/pull/5420 applied


bucket/termshark.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "2.3.0",
+    "description": "A terminal UI for tshark, inspired by Wireshark",
+    "homepage": "https://github.com/gcla/termshark",
+    "license": "MIT",
+    "depends": "wireshark",
+    "notes": [
+        "Run 'scoop install wireshark' (from a elevated shell) to install 'tshark.exe' automatically"
+    ],
+
```
rashil2000, line 10:
```suggestion
    "depends": "extras/wireshark",
```

bucket/termshark.json
```diff
@@ -0,0 +1,37 @@
...
+            "hash": "6c59c1d4f856f5611242588826274a93422a9ca9052b0a0c15a13177a604fcf9",
+            "extract_dir": "termshark_2.3.0_windows_x64"
+        }
+    },
+    "shortcuts": [
+        [
+            "termshark.exe",
+            "termshark"
+        ]
+    ],
```
rashil2000, line 23:
```suggestion
    "bin": "termshark.exe",
```

rashil2000 (changes requested):

bucket/termshark.json
Slach: @rashil2000 `tshark` is optional requirements, and need only for packet capture
also we have multiple `wireshark` package variants `wireshark-np` for example

rashil2000: > @rashil2000 `tshark` is optional requirements, and need only for packet capture 

How can it be optional if it depends on it? It doesn't work without tshark. See: 

image


> also we have multiple `wireshark` package variants `wireshark-np` for example

The extras bucket only has 'wireshark'

Slach:
@rashil2000 all changes done

PR #3154, opened 2022-01-05
coursier: Persist cache and bin

niheaven:
Coursier always create dummy `null` directory in %USERPROFILE%, I don't know why...

### DIFF
--- a/bucket/coursier.json
+++ b/bucket/coursier.json
@@ -8,13 +8,13 @@
     "bin": "coursier.jar",
     "env_set": {
         "COURSIER_CACHE": "$dir\\cache",
-		"COURSIER_BIN_DIR": "$dir\\bin"
+        "COURSIER_BIN_DIR": "$dir\\bin"
     },
-	"env_add_path": "bin",
-	"persist": [
-		"cache",
-		"bin"
-	],
+    "env_add_path": "bin",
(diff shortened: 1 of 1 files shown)


bucket/coursier.json
```diff
@@ -7,8 +7,14 @@
     "hash": "4c554d97e7b17493735aed86db98753d253fd9ca095d9ff82a117153d61b78c3",
     "bin": "coursier.jar",
     "env_set": {
-        "COURSIER_HOME": "${env:LOCALAPPDATA}\\Coursier"
+        "COURSIER_CACHE": "$dir\\cache",
+        "COURSIER_BIN_DIR": "$dir\\bin"
     },
+    "env_add_path": "bin",
```
rashil2000, line 13: `bin` field is already present, so is this required? We should keep only one of them.

rashil2000 (changes requested):

bucket/coursier.json
niheaven: This is for installed app's bin, e.g., coursier install scala3 will put scala3.bat into bin folder.

rashil2000: Understood

rashil2000 (approved):

PR #3170, opened 2022-01-10
curlie: Add version 1.6.7

niheaven:
The power of curl, the ease of use of httpie.
curlie


rashil2000 (approved):
- An instruction in Curlie's Readme  to install via Scoop would be nice (once this is merged)

> Curlie is a frontend to curl

- So does it depend on curl?

niheaven:
> * So does it depend on curl?

It's a go app, so it should not depend on curl.

PR #3178, opened 2022-01-13
update rtools, use permanent url

jeroen:
Hello I am the upstream maintainer of rtools: https://cran.r-project.org/bin/windows/Rtools/rtools40.html

This updates to the lates version, and also switches to the new permanent urls, as noted on https://cran.r-project.org/bin/windows/Rtools/rtools40.html.


niheaven:
Nice to meet you here @jeroen, I'm a deep user of R and RTools 😄 

Now `rtools` 4.0 has different version number that we scoop hasn't support yet, so the autoupdate feature of this one may not work. And if I install some additional packages that R packages need using pacman, is there some way to keep them?

jeroen:
@rashil2000 thank you for the feedback! I have pushed another attempt, does this make sense?

niheaven:
I use scoop to install rtools, that means if the installer update, the scoop will reinstall a new copy of rtools and abandon the old one, so post installed components will be lost in the new one, right?

So is there some convenient way to keep packages while updating the whole packages?

niheaven:
For the manifest and rtools file name, now R's stable version is 4.1.2 and next gen is 4.2.0, will you change rtools' name in the future? (i.e. rtools**40**)

jeroen:
@niheaven ah I did not know that. In that case perhaps it makes sense we keep the major version to 4.0 so that you are not forced to update. This is really just a new build of the installer just like the previous update (https://github.com/ScoopInstaller/Main/commit/bfdce8c2e3b28), no need to reinstall existing installations

niheaven:
I prefer using commit https://github.com/ScoopInstaller/Main/commit/d5d0f635ec2d998863e67a209375cb86a801decb with checkver and au as following:

```json
    "checkver": {
        "github": "https://github.com/r-windows/rtools-installer",
        "regex": "/tag/(\\d+)-(\\d+)-(\\d+)",
        "replace": "4.0.${1}${2}${3}"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/r-windows/rtools-installer/releases/download/$match1-$match2-$match3/rtools40-x86_64.exe"
            }
        }
    }
```

For my question, if there is not a good way to keep packages, then redownload them in demand is acceptable.

jeroen:
@niheaven thank you I have updated the PR to use your first suggestion!

niheaven (approved):
LGTM, when `checkver` in `architecture` is supported, 32bit could be added again IMO

niheaven:
A small problem left here, `current_version` is depreciated, and they should be changed to `Select-CurrentVersion`, and there're 4 in the manifest.

rashil2000:
> A small problem left here, `current_version` is depreciated, and they should be changed to `Select-CurrentVersion`, and there're 4 in the manifest.

Will it possible to use the .7z download instead of .exe? Maybe we can get rid of the (un)installer scripts

jeroen:
The installer is really preferred, it sets a bunch of system environment variables and registry settings that will enable R to find the installation. The 7z is mostly for certain CI use, but not ideal for end-users.

jeroen:
OK I have substituted all occurrences of `current_version ` with `Select-CurrentVersion`.

niheaven:
> Will it possible to use the .7z download instead of .exe? Maybe we can get rid of the (un)installer scripts

In fact, the installer and uninstaller script here is let R or MRO find rtools... It's about rtools' usage, not itself.

The difference between installer and zip file here is downloading size 😭

niheaven (approved):

PR #3200, opened 2022-01-17
linkerd: Add version 22.1.3

ksaffarian:
- Add linkerd version 22.1.3


niheaven:
- Don't add app name in description
- Use `#/` to change downloaded bin's name, so you could remove the two `Rename-Item` lines
- Use `https://linkerd.io` as homepage
- And, is there some file that could be persisted?

rashil2000:
The stable version of this application already exists - https://github.com/ScoopInstaller/Extras/blob/master/bucket/linkerd.json

The `edge` builds are weekly - so rename the manifest to linkerd-edge.json and create the PR in Versions bucket.

PR #3202, opened 2022-01-18
Update symfony-cli.json

BahmanMD:


rashil2000:
Application name in description is not required. We keep the descriptions as short and succinct as possible, and name is already apparent while installing/searching.

PR #3213, opened 2022-01-19
nodejs: add bin

jsbr:
Just add bin entry in the manifest to add node, npm and npx into the user path


rashil2000:
Not required. The root directory is added to path, so no need for shims.

jsbr:
Hmmm, In this case it doesn't work for me. I will investigate. Thank you

rashil2000:
Restart your shell, just  in case

PR #3231, opened 2022-01-22
emscripten: Add version 3.1.1

### LINKED ISSUE #3219: [Request] Add emscripten sdk
### Criteria

- [x] a non-GUI tool
- [ ] a fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts) **(I'm not confident with this one :/ )**

### Information

**Name:** emscripten

**Description:** Emscripten is an An LLVM to WebAssembly **compiler**. It compiles C/C++ to a machine code format (WebAssembly) that can be run on [most web browsers](https://webassembly.org/), and is quickly becoming a standard.

**Homepage:** https://emscripten.org/

**Download link(s):** https://emscripten.org/docs/getting_started/downloads.html/

**Some indication of popularity/repute:** Decently popular with 21.7k [stars](https://github.com/emscripten-core/emscripten) (_more than Scoop itself's_). Also an exciting emerging technology for web dev, well [WebAssembly is atleast](https://developer.mozilla.org/en-US/docs/WebAssembly/C_to_wasm).

issaclin32:
close #3219

[EMScripten](https://emscripten.org/) is an An LLVM to WebAssembly compiler. It compiles C/C++ to a machine code format (WebAssembly) that can be run on most web browsers, and is quickly becoming a standard.

**NOTES**:

* **license**: `Emscripten is available under 2 licenses, the MIT license and the
University of Illinois/NCSA Open Source License.` (https://emscripten.org/docs/introducing_emscripten/emscripten_license.html)

* `emcmdprompt.bat` opens `emsdk_env.bat` in a new window. I assume this is designed to start EMScripten environment via the shortcut (such as the case for **Anaconda**)

* I tested the app by the following steps:
1. install the package.
1. run `emsdk install latest` and `emsdk activate latest`.
1. create a file `example.c` with the following content:
```
#include <stdio.h>

int main() {
  printf("hello, world!\n");
  return 0;
}
```
4. open "EMScripten command prompt" (via the shortcut)
5. run `emcc -o example.js example.c`
    -> This generates `example.js` and `example.wasm`, in which `example.js` acts as a loader, and `example.wasm` contains actual compiled binary.
6. run `node example.js`
   -> will see `hello, world!` in echo.

### DIFF
--- a/bucket/emscripten.json
+++ b/bucket/emscripten.json
@@ -1,0 +1,48 @@
+{
+    "version": "3.1.1",
+    "description": "LLVM to WebAssembly compiler. Compiles C/C++ to WebAssembly that can be run on most web browsers.",
+    "homepage": "https://emscripten.org/",
+    "license": "MIT|NCSA",
+    "notes": [
+        "Run \"emsdk install latest\" and \"emsdk activate latest\" to set up emsdk.",
+        "For more details, check \"https://emscripten.org/docs/getting_started/downloads.html\""
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/emscripten-core/emsdk/archive/refs/tags/3.1.1.zip",
+            "hash": "604cadeecbd84eda429971d9b9a19e9fd8c2ecc3c2a1c2dc50a859dd5fcfb083"
+        }
+    },
+    "extract_dir": "emsdk-3.1.1",
+    "pre_install": "if(!(Test-Path \"$persist_dir\\.emscripten\")) {New-Item \"$dir\\.emscripten\" -ItemType File | Out-Null}",
+    "bin": [
+        "emsdk.bat",
+        "emsdk_env.bat"
+    ],
+    "shortcuts": [
+        [
+            "emcmdprompt.bat",
+            "EMScripten Command Prompt"
+        ]
+    ],
+    "persist": [
+        "java",
+        "node",
+        "python",
+        "upstream",
+        "zips",
+        ".emscripten"
+    ],
+    "checkver": {
+        "url": "https://github.com/emscripten-core/emsdk/tags",
+        "regex": "tag/([\\d.]+)"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/emscripten-core/emsdk/archive/refs/tags/$version.zip"
+            }
+        },
+        "extract_dir": "emsdk-$version"
+    }
+}


rashil2000:
Some questions:
- Any docs mentioning that the SDK works only on 64bit systems? We are downloading the repo archive, which is just a bunch of scripts essentially.
- IMO we should add `.` to PATH, instead of creating shims. Reason being, there are various shell-specific scripts in the root directory - and a `.bat` won't be able to modify shell sessions of PowerShell, Fish, Bash etc.

issaclin32:
> * Any docs mentioning that the SDK works only on 64bit systems? We are downloading the repo archive, which is just a bunch of scripts essentially.

The components (C/C++ compiler, WebAssembly compiler, bundled libs) will be downloaded after running `emsdk install latest`. The WebAssembly compiler was built in **64-bit**, so it is not usable on 32-bit systems.

> * IMO we should add `.` to PATH, instead of creating shims. Reason being, there are various shell-specific scripts in the root directory - and a `.bat` won't be able to modify shell sessions of PowerShell, Fish, Bash etc.

True. That is a concern.
However this could not be solved by just adding `.` to PATH, 
and `$dir\node\14.18.2_64bit\bin` does not exist before running `emsdk install latest`. 
Also, setting `JAVA_HOME` could pollute the base env.
Any good ideas?
```
> emsdk_env
Adding directories to PATH:
PATH += C:\Users\Issac\scoop\apps\emscripten\3.1.1
PATH += C:\Users\Issac\scoop\apps\emscripten\3.1.1\upstream\emscripten
PATH += C:\Users\Issac\scoop\apps\emscripten\3.1.1\node\14.18.2_64bit\bin

Setting environment variables:
PATH = C:\Users\Issac\scoop\apps\emscripten\3.1.1;C:\Users\Issac\scoop\apps\emscripten\3.1.1\upstream\emscripten;(ORIGINAL_PATH_VARS...)
EMSDK = C:/Users/Issac/scoop/apps/emscripten/3.1.1
EM_CONFIG = C:\Users\Issac\scoop\apps\emscripten\3.1.1\.emscripten
EMSDK_NODE = C:\Users\Issac\scoop\apps\emscripten\3.1.1\node\14.18.2_64bit\bin\node.exe
EMSDK_PYTHON = C:\Users\Issac\scoop\apps\emscripten\3.1.1\python\3.9.2-1_64bit\python.exe
JAVA_HOME = C:\Users\Issac\scoop\apps\emscripten\3.1.1\java\8.152_64bit
```

rashil2000 (approved):

PR #3244, opened 2022-01-23
lxc: Add version 4.22

### LINKED ISSUE #1668: [Request] Add LXD client
I want LXD windows client which can be found in https://ci.appveyor.com/project/lxc/lxd/branch/master/artifacts

Erisa:
@rashil2000 here is a manifest derived from the sed one using chocolatey as a source.

Some thoughts I had though:
- ~~I'm not sure how the hash regex in autoupdate is supposed to work, there is a hash on the html of the page but it seems to be the hash for `lxc.exe` and some other files rather than the nupkg archive itself. Since this also seemed to be the case for sed as well I left it intact.~~
  I was looking in the wrong place 😅 all's good there.
- `lxc.exe` stores its configuration data and files in `~\.config/lxc` when necessary - is there an example of how I can easily persist a directory like that while keeping any potential existing user data from other installs?

Closes #1668

### DIFF
--- a/bucket/lxc.json
+++ b/bucket/lxc.json
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/lxc/lxd",
     "license": "Apache-2.0",
     "url": "https://packages.chocolatey.org/lxc.4.22.nupkg",
-    "hash": "354f2d01d4874aaf58d7f7f44621c1b900881565671a37cb1540951fa3bc31ef",
+    "hash": "bd4bb1df187c883f7429d084ad78cef21fc3a4ec3d94904f8f63c977382b17e8",
     "extract_dir": "tools",
     "architecture": {
         "64bit": {
--- a/bucket/lxc.json
+++ b/bucket/lxc.json
@@ -1,29 +1,28 @@
 {
     "version": "4.22",
     "description": "LXD client",
-    "homepage": "https://github.com/lxc/lxd",
+    "homepage": "https://linuxcontainers.org/lxd",
     "license": "Apache-2.0",
-    "url": "https://packages.chocolatey.org/lxc.4.22.nupkg",
-    "hash": "bd4bb1df187c883f7429d084ad78cef21fc3a4ec3d94904f8f63c977382b17e8",
-    "extract_dir": "tools",
     "architecture": {
         "64bit": {
-            "bin": "lxc.exe"
+            "url": "https://packages.chocolatey.org/lxc.4.22.nupkg",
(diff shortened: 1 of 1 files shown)


rashil2000:
Use this:

```json
{
    "version": "4.22",
    "description": "LXD client",
    "homepage": "https://linuxcontainers.org/lxd",
    "license": "Apache-2.0",
    "architecture": {
        "64bit": {
            "url": "https://packages.chocolatey.org/lxc.4.22.nupkg",
            "hash": "bd4bb1df187c883f7429d084ad78cef21fc3a4ec3d94904f8f63c977382b17e8"
        }
    },
    "extract_dir": "tools",
    "bin": "lxc.exe",
    "checkver": {
        "url": "https://community.chocolatey.org/packages/lxc",
        "regex": "LXD client ([\\d.]+)</title>"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://packages.chocolatey.org/lxc.$version.nupkg",
                "hash": {
                    "url": "https://community.chocolatey.org/packages/lxc",
                    "regex": "$sha256.*?$basename"
                }
            }
        }
    }
}
```

rashil2000:
> lxc.exe stores its configuration data and files in ~\.config/lxc when necessary

If this location is not configurable (by a flag or any environment variable), then persist is not required

Erisa:
> > lxc.exe stores its configuration data and files in ~.config/lxc when necessary
> 
> If this location is not configurable (by a flag or any environment variable), then persist is not required

Seems you can override it by setting `LXD_CONF` environment variable.
https://github.com/lxc/lxd/blob/5f6f4c64f5ac3db69cce2a60d9224c2542eb9f01/lxc/main_aliases.go#L162-L175

rashil2000:
> Seems you can override it by setting `LXD_CONF` environment variable. https://github.com/lxc/lxd/blob/5f6f4c64f5ac3db69cce2a60d9224c2542eb9f01/lxc/main_aliases.go#L162-L175

Hmm. Do you recommend changing it (for the purpose of persist)? I'm not a user of LXC, so I don't know.

Personally, if it were me I would not recommend setting it.

Erisa:
I think (In my opinoin of course) that most users of the tool wouldn't expect it to be set. It's good for persistence, but not so much for consistency. Some of the things it stores there are generated certificates and keys which are supposed to be specific to your machine and might get confusing.

rashil2000:
Yeah, I have pretty much the same thought.

Anyway, thanks for the help :)

PR #3253, opened 2022-01-25
ninja-kitware: Add version 1.10.2.g51db2.kitware.jobserver-1

### LINKED ISSUE #3243: [Request] Ninja, Kitware fork/build
## Package Request
Kitware (CMake devs among other libs) makes available a fork/build of Ninja that integrates the long languishing job server PR/feature.

This feature is practically a necessity when integrating Ninja with other build systems, but sadly hasn't been merged for years, leading to this fork.

This fork is also what is installed by `pip install ninja` (Which is maintained by Kitware).

P.S. The binary in the PyPI repo might be newer than the one in GitHub. The maintainer might need to be contacted about this.

* https://github.com/Kitware/ninja
* https://pypi.org/project/ninja/
* https://github.com/ninja-build/ninja/issues/1139

### Criteria

- [x] a non-GUI tool

### Information

**Name:** ninja-kitware (Kitware fork/build)

**Description:** A fork/build of ninja by Kitware with the job server PR/feature merged in.

**Homepage:**
https://github.com/Kitware/ninja https://pypi.org/project/ninja/

**Download link(s):**
ninja-1.10.0.gfb670.kitware.jobserver-1_i686-pc-windows-msvc.zip
ninja-1.10.2.3-py2.py3-none-win_amd64.whl 

**Some indication of popularity/repute:** https://github.com/Kitware/ninja

segevfiner:
Closes #3243 

Note: ninja (Existing and this fork) provide 64-bit builds only but the manifests don't use `architecture`.

### DIFF
--- a/bucket/ninja-kitware.json
+++ b/bucket/ninja-kitware.json
@@ -1,0 +1,18 @@
+{
+    "version": "1.10.2.g51db2.kitware.jobserver-1",
+    "description": "A small build system with a focus on speed (Kitware fork).",
+    "homepage": "https://github.com/Kitware/ninja",
+    "license": "Apache-2.0",
+    "url": "https://github.com/Kitware/ninja/releases/download/v1.10.2.g51db2.kitware.jobserver-1/ninja-1.10.2.g51db2.kitware.jobserver-1_x86_64-pc-windows-msvc.zip",
+    "hash": "6695ef68138f6c34713e48bb1a44ea7627e868fbc658424cc81e7c08f934da92",
+    "extract_dir": "ninja-1.10.2.g51db2.kitware.jobserver-1_x86_64-pc-windows-msvc",
+    "bin": "ninja.exe",
+    "checkver": {
+        "github": "https://github.com/Kitware/ninja",
+        "regex": "/releases/tag/(?:v|V)?([\\w.-]+)"
+    },
+    "autoupdate": {
+        "url": "https://github.com/Kitware/ninja/releases/download/v$version/ninja-$version_x86_64-pc-windows-msvc.zip",
+        "extract_dir": "ninja-$version_x86_64-pc-windows-msvc"
+    }
+}
--- a/bucket/ninja-kitware.json
+++ b/bucket/ninja-kitware.json
@@ -1,18 +1,26 @@
 {
     "version": "1.10.2.g51db2.kitware.jobserver-1",
-    "description": "A small build system with a focus on speed (Kitware fork).",
+    "description": "A small build system with a focus on speed (Kitware fork with jobserver support).",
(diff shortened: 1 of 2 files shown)


rashil2000:
- 1.10.2.g51db2 should be enough for the version string
- You need to use architecture 64bit block
- Please update the existing ninja manifest with 64bit architecture (as part of this PR)

segevfiner:
@rashil2000 Note that `ninja --version` does include `.kitware.jobserver-1`. The idea is that tool can look for "jobserver" to detect this fork.

I'm not sure if they will ever release a version with `-2` but with the same prefix part.

rashil2000:
> I'm not sure if they will ever release a version with `-2` but with the same prefix part.

Can you ask them? Otherwise the `.kitware.jobserver-1` part is just noise w.r.t. Scoop - we already know it's by Kitware (the name has it), and users of ninja are aware why this fork is special (jobserver support).

(In fact you can modify the description to say `A small build system with a focus on speed (Kitware fork with jobserver support).`)

segevfiner:
> > I'm not sure if they will ever release a version with `-2` but with the same prefix part.
> 
> Can you ask them? Otherwise the `.kitware.jobserver-1` part is just noise w.r.t. Scoop - we already know it's by Kitware (the name has it), and users of ninja are aware why this fork is special (jobserver support).
> 
> (In fact you can modify the description to say `A small build system with a focus on speed (Kitware fork with jobserver support).`)

@jcfr Will you ever release a version suffixed with something other than `-1` but with the same prefix `1.10.2.g51db2` part?

jcfr:
The string `1.10.2.g51db2` is indeed sufficient to identify the associated source. `g` stands for `git` and `51db2` corresponds to the tip of the `kitware-staged-features` branch. See https://github.com/Kitware/ninja/commit/51db22c9ece4cb08f6c460b3b0257ce1a6fb5d8e

>  they will ever release a version with -2 but with the same prefix part.

If we revisit the approach associated the job server support (unlikely) or contribute/integrate some other fixes (very possible), having a version identified like the following should be considered:

```
x.y.z.g<SHA{N}>.kitware.jobserver-2.awesomefix-1
x.y.z.g<SHA{N}>.kitware.jobserver-2
```

where `<SHA{N}>` corresponds to the first `N` characters of the tip of the corresponding staging branch when creating the release and `x.y.z` the ninja release used as a base.

For example, in the past had the following while the support for dynamic dependencies was being integrated upstream:

```
ninja-1.9.0.gad558.kitware.dyndep-1.jobserver-1
```

rashil2000:
Thanks for the detailed breakup!

@segevfiner we can keep the version string as is. Please make the remaining changes.

PR #3268, opened 2022-01-29
rizin: Add version 0.3.4 (cutter will be moved to extras)

segevfiner:
Closes N/A (I can open one if you want me to).

P.S. Cutter is a GUI tool (For the underlying rizin CLI tool). Are you sure it should be in `main` and not in `extras`?


rashil2000:
If the CLI is available standalone, we can add that to Main. Otherwise, yes, it should be moved to Extras.

segevfiner:
> If the CLI is available standalone, we can add that to Main. Otherwise, yes, it should be moved to Extras.

It is. https://github.com/rizinorg/rizin/releases/tag/v0.3.4

rashil2000:
Can you add it to Main here?

rashil2000:
Move cutter.json from here to Extras (make a PR there).

Also, difference between the rizin installer and zip download?

segevfiner:
They only have a 64-bit zip for some reason 🤷‍♂️ Which is why I opted to extract the installer.

rashil2000:
Ah okay, no worries.

rashil2000:
Put 64bit URLs before 32bit URLs

PR #3269, opened 2022-01-29
rustup: Explicitly set toolchain host

tao-eternal:
compared to rustup-msvc, we should ensure the host to be `gnu` with this.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/rustup-msvc.json
+++ b/bucket/rustup-msvc.json
@@ -18,8 +18,8 @@
         "script": [
             "[Environment]::SetEnvironmentVariable('CARGO_HOME', \"$persist_dir\\.cargo\", 'Process')",
             "[Environment]::SetEnvironmentVariable('RUSTUP_HOME', \"$persist_dir\\.rustup\", 'Process')",
-            "$arch=$(If ($architecture -eq '64bit') {'x86_64'} Else {'i686'})",
-            "Invoke-ExternalCommand -Path \"$dir\\rustup-init.exe\" -Args \"-y --no-modify-path --default-host $arch-pc-windows-msvc\" | Out-Null"
+            "$host_arch = if ($architecture -eq '64bit') {'x86_64'} else {'i686'}",
+            "Invoke-ExternalCommand -Path \"$dir\\rustup-init.exe\" -Args \"-y --no-modify-path --default-host $host_arch-pc-windows-msvc\" | Out-Null"
         ]
     },
     "env_add_path": ".cargo\\bin",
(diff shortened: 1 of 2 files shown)


rasa (approved):

rashil2000:
Can you tell why this is necessary here (and not in rustup-msvc)?

tao-eternal:
> Can you tell why this is necessary here (and not in rustup-msvc)?

because there are two packages which one is `rustup` and another is `rustup-msvc`.
I want to use the `rustup` to install the rust tool-chain, and on windows it has two choice which corresponding to the `gnu` and `msvc`.
The `rustup-msvc` is default to install the rust tool-chain based on `msvc`, so we should ensure the `rustup` is default to the 'gnu'.
Also, you can take a shot at this file, if we have downloaded the `$ARCH-pc-windows-gnu/rustup-init.exe`, why we still use the `msvc`.

tao-eternal:
Actually this two `rustup-init.exe` for windows has no difference.
It's one installer, and the result just based on which you choose.
So this option `--default-host x86_64-pc-windows-gnu` is needed for those who want the rust based gnu and choose the `rustup` but not `rustup-msvc`.
I am also one choose the `rustup` but get the rust based msvc defaultly.

tao-eternal:
Sorry, I should considier thoose use the 32-Bit windows.
I'll fix it.

rashil2000:
Do not move stuff into pre_install.

Also make this change (to be explicit) in rustup-msvc.json, for msvc toolchain.

tao-eternal:
Thanks for your advice.
I have changed them.
And is there anything I can do continue.

bucket/rustup-msvc.json
```diff
@@ -18,7 +18,8 @@
         "script": [
             "[Environment]::SetEnvironmentVariable('CARGO_HOME', \"$persist_dir\\.cargo\", 'Process')",
             "[Environment]::SetEnvironmentVariable('RUSTUP_HOME', \"$persist_dir\\.rustup\", 'Process')",
-            "Invoke-ExternalCommand -Path \"$dir\\rustup-init.exe\" -Args '-y --no-modify-path' | Out-Null"
+            "$arch=$(If ($architecture -eq '64bit') {'x86_64'} Else {'i686'})",
```
rashil2000, line 21:
```suggestion
            "$arch = if ($architecture -eq '64bit') {'x86_64'} Else {'i686'}",
```

bucket/rustup-msvc.json
```diff
@@ -18,7 +18,8 @@
...
+            "Invoke-ExternalCommand -Path \"$dir\\rustup-init.exe\" -Args \"-y --no-modify-path --default-host $arch-pc-windows-msvc\" | Out-Null"
```
rashil2000, line 22:
```suggestion
            "$host_arch = if ($architecture -eq '64bit') {'x86_64'} else {'i686'}",
            "Invoke-ExternalCommand -Path \"$dir\\rustup-init.exe\" -Args \"-y --no-modify-path --default-host $host_arch-pc-windows-msvc\" | Out-Null"
```

$arch is used internally, so use $host_arch. Also, parenthesis is not required.

Same below.

tao-eternal:
As you say.

PR #3337, opened 2022-02-26
Add WinSW

SonaliBendre:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/winsw.json
+++ b/bucket/winsw.json
@@ -1,6 +1,6 @@
 {
     "version": "2.11.0",
-    "description": "A wrapper executable that can be used to host any executable as an Windows service",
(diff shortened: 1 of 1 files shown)


SonaliBendre:
https://github.com/winsw/winsw/releases/download/v3.0.0-alpha.10/WinSW-net461.exe is more smaller file, can we use that? will it work in x86 system? I don't have x86 machine to test. Also they provide zip file

rashil2000:
Don't add alpha version. Use the stable release.

Use the x86 and x64 exe files which they provide, they are self contained.

rashil2000:
Use sha256 (no prefix required).

PR #3338, opened 2022-02-26
openssl: Set OPENSSL_MODULES environment variable

### LINKED ISSUE #3335: openssl: Set the OPENSSL_MODULES environment variable
## Bug Report

**Package Name:** `openssl`

### Current Behaviour

Using the `-legacy` parameter with the [openssl pkcs12](https://www.openssl.org/docs/manmaster/man1/openssl-pkcs12.html) command fails due to OpenSSL being unable to locate the provider:

```
pkcs12: unable to load provider legacy
Hint: use -provider-path option or OPENSSL_MODULES environment variable.
78450000:error:12800067:DSO support routines:win32_load:could not load the shared library:crypto\dso\dso_win32.c:108:filename(C:\Program Files\OpenSSL\lib\ossl-modules\legacy.dll)
78450000:error:12800067:DSO support routines:DSO_load:could not load the shared library:crypto\dso\dso_lib.c:152:
78450000:error:078C0105:common libcrypto routines:provider_init:init fail:crypto\provider_core.c:907:name=legacy
```

### Expected Behaviour

The `-legacy` parameter should work without having to provide the full path to the legacy provider with the `-provider-path` parameter.

### Additional context/output

Scoop now installs the v3.x releases of OpenSSL using the builds provided by [Shining Light](https://slproweb.com/products/Win32OpenSSL.html). Several major breaking changes were introduced in OpenSSL v3.x, one of which is a change to the default encryption ciphers used on creating PKCS#12 bundles (aka. PFX files). While earlier releases of OpenSSL would use the `3DES-CBC` or `RC2-CBC` ciphers, OpenSSL v3.x defaults to the much stronger `AES-256-CBC` cipher with PBKDF2 for key derivation.

That's a good thing given the previous ciphers are no longer considered secure, but unfortunately many supported Windows releases don't support importing PFX files encrypted using the AES cipher. Support for importing PFX files encrypted with AES was introduced very recently, with the Windows Server 2019 and Windows 10 v1803 releases respectively. Earlier releases will fail to import PFX files encrypted with AES with an extremely misleading error indicating a wrong password was entered.

OpenSSL 3.x can create PKCS#12 bundles encrypted using the older, legacy ciphers if the `-legacy` parameter is provided in the `openssl pkcs12` command. This instructs OpenSSL to load the legacy provider, which is included as a separate library. As of a few days ago it is now included in the Shining Light installers, but `openssl` will not be able to find it without some additional hinting.

Note that the legacy provider has been added to the v3.0.1 installers, which have been republished. If you have already installed `openssl` with `scoop` and are already running the v3.0.1 release, you'll need to reinstall it to ensure you have the legacy provider:

```
scoop uninstall openssl
scoop cleanup -k openssl
scoop install openssl
```

### Possible Solution

Add the `OPENSSL_MODULES` environment variable to the package manifest set to `$dir\\bin`. This will hint OpenSSL so it can locate the legacy provider without having to provide a full path using `-provider-path` on each invocation.

Note this does not lower security defaults, as the legacy cryptography provider will only be used if explicitly requested.

### System details

**Windows version:** 10

**OS architecture:** 64-bit

**PowerShell version:** 7.2.1

**Additional software:** N/A

#### Scoop Configuration
```json
{
  "lastupdate": "2022-02-26T11:37:38.6928700+11:00",
  "SCOOP_REPO": "https://github.com/lukesampson/scoop",
  "SCOOP_BRANCH": "master",
  "7ZIPEXTRACT_USE_EXTERNAL": true,
  "aria2-enabled": false
}
```

ralish:
Required for OpenSSL to locate various modules (aka. providers).

Closes #3335

### DIFF
--- a/bucket/openssl.json
+++ b/bucket/openssl.json
@@ -14,7 +14,6 @@
         }
     },
     "innosetup": true,
-    "bin": "bin\\openssl.exe",
     "env_add_path": "bin",
     "env_set": {
         "OPENSSL_CONF": "$dir\\bin\\cnf\\openssl.cnf",
--- a/bucket/openssl.json
+++ b/bucket/openssl.json
@@ -19,7 +19,8 @@
     "env_set": {
         "OPENSSL_CONF": "$dir\\bin\\cnf\\openssl.cnf",
         "OPENSSL_LIB_DIR": "$dir\\lib",
-        "OPENSSL_INCLUDE_DIR": "$dir\\include"
+        "OPENSSL_INCLUDE_DIR": "$dir\\include",
+        "OPENSSL_MODULES": "$dir\\bin"
     },
     "checkver": "Win32 OpenSSL v([^\\s]+)",
     "autoupdate": {


rashil2000:
Can you also remove the `bin` field (it is redundant).

ralish:
Sorry, I'm a bit confused, without the `bin` it will be pointing to the parent folder, which does not contain the relevant file?
image

rashil2000:
I was talking about the `"bin": "bin\\openssl.exe",` part. I've made the change.

ralish:
Ah, I see. That part was already present in the manifest so I left it untouched. Thanks for merging!

PR #3339, opened 2022-02-27
rename latex to miktex

### LINKED ISSUE #1427: [Proposal] Rename package 'latex' to 'miktex-basic'
Hi,

I wanted update my CV written in LaTeX. I downloaded **latex** package via **scoop** and tried build the project using **Visual Studio Code**.
During project building a few packages has been downloaded.
I checked what happen - why that packages has been downloaded.

I checked manifest file and be surprised when saw **miktex-basic** as package which has been installed.

In my opinion this package name is not intuitive for users and should be renamed.

My CV is advanced document and I need a full **LaTeX** package to build it.

I will try add new manifest which will install a full package but now I don't have enought time for it and I must do it manually.

zStruCat:
Closes #1427

Though discussion in this issue agree that renaming is worthy due to reinstallation, but there are precedents of renaming such as #3315, #3390.


rashil2000:
Why miktex-basic instead of just miktex?

zStruCat:
> Why miktex-basic instead of just miktex?

Possibly because miktex is a package manager for latex and doesn't automatically download all related latex package.

rashil2000:
Existence of miktex-basic would imply there is some other full package too. I think miktex would be sufficient for us.

niheaven:
Though I don't install latex from scoop, this renaming could break UX. You should add a separate `miktex-basic` and leave `latex` alone. This is just like `gcc`, so where the `latex` package is?

The original issue is `discussion`, not a conclusion and a recommendation.

rashil2000:
I actually use latex from Scoop, and I endorse this change.

IMO this is a bit different from GCC case where one package has less tools (`gcc`) and other has many tools (`mingw`). 

Miktex is just a distro of latex, so it would be better to use `miktex` unless we have a standalone/official `latex` package.

niheaven:
From https://www.latex-project.org/get/, MiKTeX is the recommended way to install LaTeX, and there is no standalone/official latex package (we just sure that TeXLive is the official build of TUG). So if someone want to use `latex`, MiKTeX is the best way and for a long time the only way to use under Windows.

This could not simply be called a `distro` which may have alternates. I'm not against call this manifest `miktex`, but want to keep a same manifest as `latex`.

rashil2000:
> From https://www.latex-project.org/get/, MiKTeX is the recommended way to install LaTeX, and there is no standalone/official latex package (we just sure that TeXLive is the official build of TUG). So if someone want to use `latex`, MiKTeX is the best way and for a long time the only way to use under Windows.

image

Hmm they actually list 3 different distributions, so I'm not sure which one out of MikTex, proText and TexLive to call official?

niheaven:
proTeXt is based on MiKTeX, TeX Live is mainly for Unix-related (incl. macOS), this is what `for a long time the only way` means. (There IS AND ONLY IS MiKTeX for several years while TeX Live hadn't supported Windows)

rashil2000:
See #3341

PR #3340, opened 2022-02-27
zoxide: Remove `env_set` and `persist`

sitiom:
Each user should have their own database, and if zoxide is installed globally, it fails to write to the data directory when not running as admin, so the default value set by zoxide (`$Env:APPDATA`) is already ok.


rashil2000:
I think we can remove the note about PowerShell, as zoxide can be used from any shell

sitiom:
Ah right, but the clink config isn't mentioned in the README yet.

rashil2000:
Maybe you can remind the author to update the Readme? See https://github.com/ajeetdsouza/zoxide/issues/218

PR #3341, opened 2022-02-28
latex: Add back - #3339

rashil2000:
I'm okay with reverting this change but I'd still want to get an opinion of more experienced LaTeX users.

/cc @JanPokorny @niheaven @zStruCat @issaclin32


JanPokorny:
Naming the package `miktex` makes sense. Calling it `latex` is basically like calling GCC package `c-compiler` or Firefox package `web-browser`. Furthermore, Linux package managers use the distribution name (like `texlive-full`), I have never seen one use just `latex`.

niheaven:
Copying it instead of renaming it, I mean `miktex` is a proper name, but `latex` should be kept for some convenience. Until scoop supports alias app name, the two should be the same.

JanPokorny:
I think that's just confusing. Users would ask "what's the difference
between the two, which one should I install"? Also makes it harder to
maintain, sooner or later someone merges a PR that touches one and not the
other.

There are no other "convenience" aliases in Scoop and users don't even
expect them to exist since no other package manager has them.

What users do when installing latex is probably "scoop install miktex"
directly (I did, I had to Google that the package is called "latex") or
"scoop search latex" which would find miktex. So I don't really see any
added convenience, just inconvenience.

niheaven:
For those who have installed miktex with `scoop install latex`, give them a warning that what they actually installed is `miktex` and what they do with this `latex` app is being done to `miktex` app (scoop update latex, scoop uninstall latex, etc.). But this is what we called `alias` feature (TBD). For now, I think keeping `latex` here has no other inconvenience since it is here for several years.

JanPokorny:
@rashil2000 I'd also add "DEPRECATED" to notes and description (since search engines like ScoopSearch don't show notes) to the `latex` package (and eventually remove it once Scoop finally has the ability to move manifests)

niheaven (approved):

PR #3345, opened 2022-03-02
vut: Add version 0.1.3

forbjok:

Vut is a utility for propagating version numbers using methods such as regex replacement in files and templates.
Uses the Semantic Versioning 2.0.0 specification. (http://semver.org/)

No persist, as there is no global configuration, only per-repository configuration placed in the repositories it's used in.


rashil2000:
Hi!

Thanks for the PR. However, this application does not meet the criteria for official buckets. 

I would recommend creating an issue for a package request for this instead, so that we can track it.

PR #3354, opened 2022-03-06
goodbyedpi: Add version 0.2.1

issaclin32:
closes https://github.com/ScoopInstaller/Extras/issues/8094

supersedes https://github.com/ScoopInstaller/Extras/pull/8096

[GoodbyeDPI](https://github.com/ValdikSS/GoodbyeDPI) is a tool to bypass Deep Packet Inspection systems found in many Internet Service Providers which block access to certain websites.

### DIFF
--- a/bucket/goodbyedpi.json
+++ b/bucket/goodbyedpi.json
@@ -8,12 +8,6 @@
     "hash": "acb0568cfc184cdd52c235c9ccc51a38fd78eb57564dcf7dce5188b9ad93f391",
     "extract_dir": "goodbyedpi-0.2.1",
     "bin": "goodbyedpi.exe",
-    "shortcuts": [
-        [
-            "goodbyedpi.exe",
-            "GoodbyeDPI"
-        ]
-    ],
     "architecture": {
         "64bit": {
             "pre_install": [
--- a/bucket/goodbyedpi.json
+++ b/bucket/goodbyedpi.json
@@ -3,7 +3,6 @@
     "description": "Bypass Deep Packet Inspection systems found in many Internet Service Providers which block access to certain websites.",
     "homepage": "https://github.com/ValdikSS/GoodbyeDPI",
     "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


bucket/goodbyedpi.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "0.2.1",
+    "description": "Bypass Deep Packet Inspection systems found in many Internet Service Providers which block access to certain websites.",
+    "homepage": "https://github.com/ValdikSS/GoodbyeDPI",
+    "license": "Apache-2.0",
+    "notes": "Follow the instructions on 'https://github.com/ValdikSS/GoodbyeDPI#readme' to set up GoodbyeDPI",
```
rashil2000, line 6:
```suggestion
```

We can omit this because this link is already present in homepage property

issaclin32: OK, I'll take your advice.

bucket/goodbyedpi.json
```diff
@@ -0,0 +1,29 @@
...
+        "64bit": {
+            "pre_install": [
+                "Move-Item \"$dir\\x86_64\\*\" \"$dir\\\"",
+                "Remove-Item \"$dir\\x86\", \"$dir\\x86_64\" -Force -Recurse"
+            ]
+        },
+        "32bit": {
+            "pre_install": [
+                "Move-Item \"$dir\\x86\\*\" \"$dir\\\"",
+                "Remove-Item \"$dir\\x86\", \"$dir\\x86_64\" -Force -Recurse"
```
rashil2000, line 20: It's not necessary, but since this line is common, should we move it outside architecture block (post_install)?

issaclin32: I don't think we should do this. It reduces `pre_install` by a little bit, but introduces `post_install`. This actually increases the complexity IMO.

rashil2000: Okay, sounds good

rashil2000 (approved):

PR #3369, opened 2022-03-09
ffmpeg: Persist font folder

KaranKad:
For ffmpeg to work with fonts correctly font folder is needed.
See https://web.archive.org/web/20200511024225/https://ffmpeg.zeranoe.com/forum/viewtopic.php?f=7&t=2554

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/ffmpeg-shared.json
+++ b/bucket/ffmpeg-shared.json
@@ -10,7 +10,6 @@
             "extract_dir": "ffmpeg-5.0-full_build-shared"
         }
     },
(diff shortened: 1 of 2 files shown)


bucket/ffmpeg-shared.json
```diff
@@ -10,10 +10,12 @@
             "extract_dir": "ffmpeg-5.0-full_build-shared"
         }
     },
+    "pre_install": "if (!(Test-Path \"$persist_dir\\fonts\")) { New-Item \"$dir\\fonts\" -ItemType Directory | Out-Null }",
```
issaclin32, line 13:
```suggestion
```
*pre_install* is actually not needed, because *Scoop* will treat items in persist as a folder by default, if it does not exist in the archive.

PR #3383, opened 2022-03-14
enomic: Fix checkver URL

rasa:
Per https://github.com/ScoopInstaller/Main/runs/5537110586?check_suite_focus=true#step:3:257

Closes #XXXX

Relates to #XXXX


tech189:
I just noticed this file is called `enomic.json` with an 'm', should be an 'n'.

tech189:
Is checkver failing because there is an update available?

rasa:
/verify

I have no idea why the checkver is failing. It works fine when I run checkver.ps1 locally.

rasa:
> Is checkver failing because there is an update available?
> 
> https://github.com/ScoopInstaller/Main/runs/5543184605?check_suite_focus=true#step:3:248

@tech189 It appears you were right. Thanks for the help.

PR #3386, opened 2022-03-14
cake: remove autoupdate because no longer updated

tech189:
- Relates to Excavator error: https://github.com/ScoopInstaller/Main/runs/5534119474?check_suite_focus=true#step:3:202
- Adds note explaining what the user should do to upgrade


rashil2000:
There are some new releases, do we have a separate manifest for them? 

https://github.com/cake-build/cake/releases

tech189:
The new releases don't contain a separate cake.exe anymore, they're just C# modules/packages as far as I can tell.

PR #3388, opened 2022-03-15
coq: update to version 2022.01.0

tech189:
- Relates to Excavator update available - https://github.com/ScoopInstaller/Main/runs/5534119474?check_suite_focus=true#step:3:228
- No autoupdate because no stable file names at the moment, raised it as a problem here: https://github.com/coq/platform/issues/164#issuecomment-1067416021


rashil2000:
Remove checkver too (otherwise it will needlessly consume an excavator cycle)

PR #3391, opened 2022-03-15
Update 1password-cli to version 2.0.0

astelmachonak-nydig:
This PR updates 1password-cli to v2.0.0. Note: v2 is a major release and changes command line API completely: https://developer.1password.com/docs/cli/upgrade/#step-2-update-your-scripts

### DIFF
--- a/bucket/1password-cli.json
+++ b/bucket/1password-cli.json
@@ -1,33 +1,33 @@
 {
-    "version": "1.12.4",
+    "version": "2.0.0",
     "description": "The 1Password command-line tool makes your 1Password account accessible entirely from the command line.",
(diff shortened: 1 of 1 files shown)


astelmachonak-nydig:
We might want to wait with merging this - the CLI commands is very different and pushing this is simple update will break lots of people.

astelmachonak-nydig:
Also, since download URLs changed, I can't install older version after this one is installed using `scoop install 1password-cli@1.12.4` as well. Not sure what is the best way to handle this...

rashil2000:
The best way in this case would be to use the old manifest manually

PR #3398, opened 2022-03-17
argo: update to version 3.3.0

tech189:
- Relates to Excavator error: https://github.com/ScoopInstaller/Main/runs/5585775074?check_suite_focus=true#step:3:196
- Simplify by removing pre_install because they no longer release the exe without an extension


rashil2000:
I think renaming is still a cleaner solution. We can rename argo-windows-amd64.exe to argo.exe

PR #3405, opened 2022-03-18
azure-kubelogin: Add version 0.11.0

stevehipwell:
This PR adds Azure Kubelogin which is required to authenticate with an AKS cluster.

### DIFF
--- a/bucket/azure-kubelogin.json
+++ b/bucket/azure-kubelogin.json
@@ -1,0 +1,26 @@
+{
+    "version": "0.0.11",
+    "description": "A client-go credential (exec) plugin implementing azure authentication.",
+    "homepage": "https://github.com/Azure/kubelogin/",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Azure/kubelogin/releases/download/v0.0.11/kubelogin-win-amd64.zip",
+            "hash": "8d509b64638bf2acba247b5333b8dc1f16a578ecb1917b562ec1ee938e49d85f"
+        }
+    },
+    "bin": "bin\\windows_amd64\\kubelogin.exe",
+    "checkver": {
+        "github": "https://github.com/Azure/kubelogin"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/Azure/kubelogin/releases/download/v$version/kubelogin-win-amd64.zip"
+            }
+        },
+        "hash": {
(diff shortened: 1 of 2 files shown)


bucket/azure-kubelogin.json
```diff
@@ -0,0 +1,26 @@
+{
+    "version": "0.0.11",
+    "description": "A client-go credential (exec) plugin implementing azure authentication.",
```
rashil2000, line 3:
```suggestion
    "description": "A client-go credential (exec) plugin implementing Azure authentication.",
```

bucket/azure-kubelogin.json
```diff
@@ -0,0 +1,26 @@
...
+    "homepage": "https://github.com/Azure/kubelogin/",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Azure/kubelogin/releases/download/v0.0.11/kubelogin-win-amd64.zip",
+            "hash": "8d509b64638bf2acba247b5333b8dc1f16a578ecb1917b562ec1ee938e49d85f"
+        }
+    },
+    "bin": "bin\\windows_amd64\\kubelogin.exe",
```
rashil2000, line 12: This might conflict with https://github.com/ScoopInstaller/Extras/blob/cc0020bb0368d7380ebbef24c4cc8e214bfce0a4/bucket/kubelogin.json#L12

bucket/azure-kubelogin.json
```diff
@@ -0,0 +1,26 @@
...
+    "checkver": {
+        "github": "https://github.com/Azure/kubelogin"
+    },
```
rashil2000, line 15:
```suggestion
    "checkver": "github",
```

bucket/azure-kubelogin.json
stevehipwell: I've changed this, but it's now not the same as the official docs on the repo README.

bucket/azure-kubelogin.json
stevehipwell: The package name wont conflict, but the binary will. This isn't a Scoop problem, but a problem with vendors not checking for prior art. I strongly expect that these two tools are mutually exclusive. The binary needs to be called `kubelogin` so there is no way round it.

bucket/azure-kubelogin.json
stevehipwell: I've changed this, but I'm not convinced that the suggested pattern is better than the explicit one provided.

PR #3416, opened 2022-03-21
JEnv: Add version 2.0.0

nightscape:
Adds https://github.com/FelixSelter/JEnv-for-Windows for managing Java versions.

The Contributing Guide says that unless the architecture is 32-bit only, one should explicitly specify it.
JEnv is a set of Powershell scripts, so it's independent of architecture.
Should I specify the same autoupdate url twice then?

### DIFF
--- a/bucket/jenv.json
+++ b/bucket/jenv.json
@@ -1,0 +1,16 @@
+{
+    "version": "2.0.0",
+    "description": "Change your current Java version with one line",
+    "homepage": "https://github.com/FelixSelter/JEnv-for-Windows",
+    "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


nightscape:
Hey @rashil2000, which of the criteria do you see as not fulfilled?
Should I create this PR against another bucket?
I'm using this, and it makes Java management on Windows much more flexible compared to https://github.com/ScoopInstaller/Scoop/wiki/Java#switching-javas

rashil2000:
The popularity/well-known criterion.

I think the Java bucket might be a good place for this

nightscape:
Ok 👌
Thanks for the quick response!

PR #3452, opened 2022-03-28
xtensa-esp32-elf: Update to -patch3, fix checkver

tech189:
- Relates to Excavator checkver not working: https://github.com/ScoopInstaller/Main/runs/5709628840?check_suite_focus=true#step%3A3%3A337=

- Added autoupdate hash extraction - is there a simpler way to not repeat the url?


rashil2000:
>  is there a simpler way to not repeat the url?

I don't think so, since the hash regex is different we will have to duplicate URL

PR #3457, opened 2022-03-29
cacert: added env "SSL_CERT_FILE"

KaranKad:
added `"SSL_CERT_FILE"` as enviroment because [axel](https://github.com/ScoopInstaller/Main/blob/master/bucket/axel.json) downloader requires it. Without that it fails to download anything unless you use `-k` switch which skips SSL certificate check of the URL and is insecure.

Will add `cacert` as dependency to axel.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/cacert.json
+++ b/bucket/cacert.json
@@ -6,6 +6,9 @@
     "license": "MPL-2.0",
     "url": "https://curl.haxx.se/ca/cacert-2022-03-29.pem#/cacert.pem",
     "hash": "1979e7fe618c51ed1c9df43bba92f977a0d3fe7497ffa2a5e80dfc559a1e5a29",
+    "env_set": {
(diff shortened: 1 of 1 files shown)


rashil2000:
This environment variable can potentially interfere with other applications, or even with the system itself. I'm not sure if this is a safe idea.

tech189:
Is there a config file you could edit in pre_install to point to the cacert file if it's installed?

KaranKad:
> This environment variable can potentially interfere with other applications, or even with the system itself. I'm not sure if this is a safe idea.

After looking at it, if a user wants to have a custom location for `SSL_CERT_FILE` it might interfere.
> Is there a config file you could edit in pre_install to point to the cacert file if it's installed?

I searched for it but couldn't find any documentation for it.

I'll close this PR

PR #3469, opened 2022-04-01
yosys: Remove autoupdate, add archive URL & outdated note

tech189:
- Relates to Excavator unable to perform checkver: https://github.com/ScoopInstaller/Main/runs/5784628664?check_suite_focus=true#step:3:321
- It looks as if the site where Yosys was previously hosted has been hacked/bought in order to promote some online casinos: https://clifford.at/
- The download page previously looked like this: https://web.archive.org/web/20210622104453/http://www.clifford.at/yosys/download.html
- On the old download page, there was a GitHub link to https://github.com/cliffordwolf/yosys, which redirects to https://github.com/YosysHQ/yosys and links to their new homepage https://yosyshq.net/yosys/
- On their website you can find old links to Yosys' download page - https://yosyshq.net/yosys/download.html - which now says that Yosys is best downloaded as part of OSS CAD Suite: https://github.com/YosysHQ/oss-cad-suite-build
- I made a PR for adding OSS CAD Suite to versions https://github.com/ScoopInstaller/Versions/pull/475, so I made a note to tell users of this manifest to update by using the package in Versions (this version was released 26 Aug 2019, latest is 0.15)
- This manifest will continue to work because the download has been moved to the one hosted on archive.org, it has the same hash so it should be identical (I have tested it works)

### DIFF
--- a/bucket/yosys.json
+++ b/bucket/yosys.json
@@ -1,7 +1,7 @@
 {
     "version": "0.9",
     "description": "Framework for Verilog RTL synthesis",
-    "homepage": "http://www.clifford.at/yosys/about.html",
+    "homepage": "https://yosyshq.net/yosys/",
     "license": "ISC",
     "notes": "Separate binaries are no longer released for Yosys, but updated versions are included in OSS CAD Suite (versions/oss-cad-suite-nightly)",
     "url": "https://web.archive.org/web/20210622104453/http://www.clifford.at/yosys/nogit/win32/yosys-win32-mxebin-0.9.zip",
--- a/bucket/yosys.json
+++ b/bucket/yosys.json
@@ -3,19 +3,12 @@
     "description": "Framework for Verilog RTL synthesis",
     "homepage": "http://www.clifford.at/yosys/about.html",
     "license": "ISC",
-    "url": "http://www.clifford.at/yosys/nogit/win32/yosys-win32-mxebin-0.9.zip",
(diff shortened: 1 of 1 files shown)


tech189:
```
INFO: Hashes

Log of Output:

    Downloading https://web.archive.org/web/20210622104453/http://www.clifford.at/yosys/nogit/win32/yosys-win32-mxebin-0.9.zip (-1 B)... 
    yosys:  
    Mismatch found  
    	URL:		https://web.archive.org/web/20210622104453/http://www.clifford.at/yosys/nogit/win32/yosys-win32-mxebin-0.9.zip 
    	First bytes:	3C 21 44 4F 43 54 59 50 
    	Expected:	d638d860dff1f351c15ff0caeaa668742649677b874b1a9bdb0b6ce0e54840f5 
    	Actual:		77b184bd7dd1449b6755f2caccaf5599320f230b2367d52420e1ffa020d237d2
```

GitHub actions thinks there is a mismatch, maybe because there is a redirect in the download? Should I make a PR to add the zip to https://github.com/ScoopInstaller/Binary instead?

tech189:
Sadly not. But the download from archive.org has the same hash as in the manifest from 2019 when the site was still up. (https://github.com/ScoopInstaller/Main/commit/65d0ce7b763f016d4138b5b356e037c6bffbf35b) Is that sufficient?

rashil2000:
Yes, that's alright

PR #3472, opened 2022-04-03
Add dig.exe to bin list for bind.json

leonyu:
Add dig.exe to bin list for bind.json so it is searchable by `scoop search dig`. 

`dig` is a common tool used for advanced DNS lookup, it is part of `bind`, unless people know this, they might not know to install `bind`. Current `bind.json` adds bind `bin` path user's ENV, so no shim is necessary, this just make `dig` searchable.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/bind.json
+++ b/bucket/bind.json
@@ -21,7 +21,38 @@
             "Remove-Item \"$dir\\*\" -Exclude 'bin'"
         ]
     },
-    "env_add_path": "bin",
+    "bin": [
+        "bin\\arpaname.exe",
+        "bin\\ddns-confgen.exe",
+        "bin\\delv.exe",
+        "bin\\dig.exe",
+        "bin\\dnssec-cds.exe",
+        "bin\\dnssec-dsfromkey.exe",
+        "bin\\dnssec-importkey.exe",
+        "bin\\dnssec-keyfromlabel.exe",
+        "bin\\dnssec-keygen.exe",
+        "bin\\dnssec-revoke.exe",
+        "bin\\dnssec-settime.exe",
+        "bin\\dnssec-signzone.exe",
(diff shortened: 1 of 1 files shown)


rashil2000:
This should be avoided. It will lead to duplicated binaries in user's Path.

leonyu:
Any suggestion for desired behavior -- making `dig` searchable?

rasa (changes requested):
The line above
`"env_add_path": "bin",`
added all the files in bin to the path, so your change is redundant. Instead, what you could do is add every .exe to `"bin"` and remove the `"env_add_path": "bin",` line.

rashil2000:
> Any suggestion for desired behavior -- making `dig` searchable?

The proper way would be to add dig.exe (and related tools) in the description, and wait for https://github.com/ScoopInstaller/Scoop/issues/4412.

But for now @rasa's workaround should be good.

rasa (approved):

bucket/bind.json
```diff
@@ -21,7 +21,35 @@
...
+        "bin/host.exe",
+        "bin/mdig.exe",
+        "bin/named-checkconf.exe",
+        "bin/named-checkzone.exe",
+        "bin/named-compilezone.exe",
+        "bin/named-journalprint.exe",
+        "bin/named-rrchecker.exe",
+        "bin/named.exe",
+        "bin/nsec3hash.exe",
+        "bin/nslookup.exe",
```
rashil2000, line 47: To prevent conflict with `C:\WINDOWS\System32\nslookup.exe`

```suggestion
        [
            "bin/nslookup.exe",
            "nslookup-bind"
        ],
```

rashil2000 (commented):
Please use `\\` instead of `/` for all bins

PR #3482, opened 2022-04-08
uutils: add symlinks for various tools

neheb:
adds symlinks to the various tools. Similar to busybox.

- [x ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).


rashil2000:
Some people might not prefer creating so many shims, and many shims conflict with Windows internal tools. I would suggest creating a new manifest, similar to [busybox](https://github.com/ScoopInstaller/Main/blob/master/bucket/busybox.json) and [busybox-lean](https://github.com/ScoopInstaller/Main/blob/master/bucket/busybox-lean.json).

neheb:
Sort of done.

Which shims conflict with Windows tools?

rashil2000:
sort, date, whoami, echo, mkdir, rmdir, tar

neheb:
Some of those don't come with Windows < 10. busybox currently comes with all of those symlinks.

PR #3486, opened 2022-04-10
cmder: handle spaces in %USERPROFILE%

nkpvk:
If a Windows user has a name with spaces in it, `cmder` will fail to start.


tech189:
We can ignore the Autoupdate Hash Extraction failure for both manifests, the current release's hashes file is named `hashes.1.txt` which is probably a one-off (https://github.com/cmderdev/cmder/releases/download/v1.3.19/hashes.1.txt)

PR #3492, opened 2022-04-12
jdtls: Add version 1.10.0-202204081744

agorgl:
[+] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/jdtls.json
+++ b/bucket/jdtls.json
@@ -1,0 +1,29 @@
+{
+    "version": "1.11.0-202204140336",
+    "description": "Java language server",
+    "homepage": "https://projects.eclipse.org/projects/eclipse.jdt.ls",
+    "license": "EPL-2.0",
+    "suggest": {
+        "JDK": "java/openjdk",
+        "Python": "python"
+    },
+    "url": "https://download.eclipse.org/jdtls/snapshots/jdt-language-server-1.11.0-202204140336.tar.gz",
+    "hash": "946bb77d0863e3137600e8451845a3d853bf64e87a90c99e7bb936d9ca34e7fa",
+    "installer": {
+        "script": "shim (search_in_path python.exe) $global jdtls \"$dir\\bin\\jdtls\""
+    },
+    "uninstaller": {
+        "script": "rm_shim jdtls (shimdir $global) jdtls"
+    },
+    "checkver": {
+        "url": "https://download.eclipse.org/jdtls/snapshots/latest.txt",
+        "regex": "jdt-language-server-([\\d.]+)-(\\d+)",
+        "replace": "${1}-${2}"
+    },
+    "autoupdate": {
+        "url": "https://download.eclipse.org/jdtls/snapshots/jdt-language-server-$match1-$match2.tar.gz",
+        "hash": {
+            "url": "$url.sha256"
+        }
+    }
+}


tech189:
I'm not sure manually adding shims like this is a good idea - it won't work if Scoop's Python isn't installed or if you use pyenv or something similar. Probably best to at least depend on python like [poetry](https://github.com/ScoopInstaller/Main/blob/master/bucket/poetry.json).

bucket/jdtls.json
```diff
@@ -0,0 +1,29 @@
...
+        "Python3": "python"
+    },
+    "url": "https://download.eclipse.org/jdtls/snapshots/jdt-language-server-1.10.0-202204081744.tar.gz",
+    "hash": "da4c832d89e1e66932eb809c5760043fe1abdbefc1a4492a8022edd354c873c5",
+    "installer": {
+        "script": "scoop shim add jdtls $scoopdir\\apps\\python\\current\\python.exe $dir\\bin\\jdtls"
+    },
+    "uninstaller": {
+        "script": "scoop shim rm jdtls"
+    },
```
rashil2000, line 17:
```suggestion
    "bin": "bin\\jdtls.py",
```

You can directly do this

agorgl: Yeah sorry I should had explained the reasoning behind this!

Currently the "bin" option in scoop generates a NON-exe shim if used like the above suggestion; this initially might seem ok, but it cannot be used with programs that need native executables to use it and not powershell scripts.

Such an example is the `neovim` language server client and any other program that relies on `libuv`'s `spawn` method (see here: https://github.com/libuv/libuv/blob/2a1b880f5439e074ef4d5c556f626b1044fb6781/src/win/process.c#L297)

The `depend` option seems a good idea, I will modify the manifest to depend on `python`.

Finally as a side note, the current PR will allow scoop + neovim users to setup painlessly their java development environment when this PR lands too!: https://github.com/neovim/nvim-lspconfig/pull/1833

agorgl:
Any updates on this?

bucket/jdtls.json
```diff
@@ -0,0 +1,29 @@
...
+    "homepage": "https://projects.eclipse.org/projects/eclipse.jdt.ls",
+    "license": "EPL v2.0",
+    "depends": "python",
+    "suggest": {
+        "JDK": "java/openjdk"
+    },
+    "url": "https://download.eclipse.org/jdtls/snapshots/jdt-language-server-1.10.0-202204081744.tar.gz",
+    "hash": "da4c832d89e1e66932eb809c5760043fe1abdbefc1a4492a8022edd354c873c5",
+    "installer": {
+        "script": "scoop shim add jdtls $scoopdir\\apps\\python\\current\\python.exe $dir\\bin\\jdtls"
```
rashil2000, line 13:
```suggestion
        "script": "shim python $global jdtls $dir\\bin\\jdtls"
```

bucket/jdtls.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "uninstaller": {
+        "script": "scoop shim rm jdtls"
```
rashil2000, line 16:
```suggestion
        "script": "rm_shim jdtls (shimdir $global) jdtls"
```

rashil2000:
Make Python as suggestion instead of dependency

bucket/jdtls.json
agorgl: This actually gives: `Can't shim 'python': couldn't find 'python'`

agorgl: Fixed!

bucket/jdtls.json
agorgl: Fixed!

agorgl:
Fixed!

bucket/jdtls.json
```diff
@@ -0,0 +1,29 @@
+{
+    "version": "1.10.0-202204081744",
+    "description": "Java language server",
+    "homepage": "https://projects.eclipse.org/projects/eclipse.jdt.ls",
+    "license": "EPL v2.0",
```
issaclin32, line 5:
```suggestion
    "license": "EPL-2.0",
```
See [SPDX license identifiers](https://spdx.org/licenses/).

bucket/jdtls.json
```diff
@@ -0,0 +1,29 @@
...
+    "suggest": {
+        "JDK": "java/openjdk",
+        "Python": "python"
+    },
+    "url": "https://download.eclipse.org/jdtls/snapshots/jdt-language-server-1.10.0-202204081744.tar.gz",
+    "hash": "da4c832d89e1e66932eb809c5760043fe1abdbefc1a4492a8022edd354c873c5",
+    "installer": {
+        "script": "shim \"$(currentdir python $global)\\python.exe\" $global jdtls \"$dir\\bin\\jdtls\""
```
rashil2000, line 13:
```suggestion
        "script": "shim (search_in_path python) $global jdtls \"$dir\\bin\\jdtls\""
```

A more robust way would be this

bucket/jdtls.json
agorgl: Fixed!

bucket/jdtls.json
agorgl: Yeap, much cleaner thanks! Had to change `(search_in_path python)` to `(search_in_path python.exe)` for this to work correctly!

agorgl:
Fixed all the suggestions, I think we are good to go!

agorgl:
Updated to latest version, lets run the checks again

rashil2000:
Great, thanks for your patience. 

For a first-time contribution, you've done excellent work!

PR #3513, opened 2022-04-19
nova: Add version 3.0.2

stevehipwell:
This PR adds Nova v3.0.2.

### DIFF
--- a/bucket/nova.json
+++ b/bucket/nova.json
@@ -1,0 +1,26 @@
+{
+    "version": "3.0.2",
+    "description": "Utility to scan your Kubernetes cluster for installed Helm charts, then cross-checks them against all known Helm repositories. If it finds an updated version of the chart you're using, or notices your current version is deprecated, it will let you know.",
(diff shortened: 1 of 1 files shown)


stevehipwell:
@rashil2000 the description is from the project README, `scoop` is a package manager and should not be in the business of writing copy for apps.

rashil2000:
App descriptions should be concise and to the point. The exact same information can be conveyed in much less words than the current one. Besides, have a look at what they use in their About section:

image

stevehipwell:
@rashil2000 where is that documented in a way that says what I've submitted here is not concise and to the point? Things like this are subjective and there aren't ever "correct" answers, I thought that the Scoop reboot, that I played a major part in starting, was going to get rid of petty nits in PRs. Unless something is explicitly documented and can be validated automatically we should be permissive rather than restrictive.

rashil2000:
@issaclin32 thoughts?

issaclin32:
I am fine with a longer "description". As long as it is clear enough and **does not look confusing** to users.

desc

Issue #3304, opened 2022-02-15
tor@0.4.6.9: download via aria2 failed


leochiukl:
#### Scoop Configuration

```json
{
    "lastupdate":  "2022-02-15T12:23:57.6198469+08:00",
    "SCOOP_REPO":  "https://github.com/lukesampson/scoop",
    "SCOOP_BRANCH":  "master",
    "aria2-enabled":  true,
    "aria2-warning-enabled":  false
}
```

niheaven:
Fixed in https://github.com/ScoopInstaller/Main/commit/a058307b0709000277b58d4bd2896b2dd265543a

PR #3514, opened 2022-04-19
tor: Update invalid download url (torbrowser version segment /11.0.7/ -> /11.0.10/)

timepigeon:
Url has been updated to the most recent torbrowser version 11.0.10 per https://dist.torproject.org/torbrowser/ since 11.0.7 isn't offered anymore and the current download link 404s: https://dist.torproject.org/torbrowser/11.0.7/tor-win64-0.4.6.10.zip .

The underlying tor version is the same `0.4.6.10` but I get different file hashes (that probably should be verified).  
  
Updating this part of the url seems to be necessary around every three months, given torbrowser following the ESR release schedule, while keeping the three latest stable versions hosted:

https://gitlab.torproject.org/tpo/applications/tor-browser/-/wikis/Release-Schedule
https://wiki.mozilla.org/Release_Management/Calendar
https://dist.torproject.org/torbrowser/

Relates to #3304
  

### DIFF
--- a/bucket/tor.json
+++ b/bucket/tor.json
@@ -9,12 +9,12 @@
     ],
     "architecture": {
         "64bit": {
-            "url": "https://dist.torproject.org/torbrowser/11.0.7/tor-win64-0.4.6.10.zip",
-            "hash": "8eb81cae2a54103483c2740c22c75bdd01f646a3c38e88a011da4e4bafbd34e6"
+            "url": "https://dist.torproject.org/torbrowser/11.0.10/tor-win64-0.4.6.10.zip",
+            "hash": "ecf6ccb6c39421d048bac95091758d3c48fd43b8e7fb1b3d7203910ef05a82e4"
         },
         "32bit": {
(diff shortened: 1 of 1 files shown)


issaclin32 (approved):

issaclin32:
@rashil2000 Should we change version pattern to `0.4.6.10-11.0.10` to avoid the problem?

rashil2000:
Sure, we could do that, but I'm confused about which part receives updates more often, 0.4.6.10 or 11.0.10?

rashil2000:
If the old URL becomes 404, then this is the only solution I guess. We can do it.

timepigeon:
Looking at chocolateys install script for tor ( https://community.chocolatey.org/packages/tor#files - click on 'show') , it's linking to https://https://archive.torproject.org/tor-package-archive/torbrowser/ instead of https://dist.torproject.org/torbrowser/, with the advantage that all torbrowser releases seem to keep being stored there, instead of only the last three,  which should remove the need to update the url without there being an actual update of core tor itself.  

I've asked on their IRC channel if there is an alternative way to reliably link to the latest stable windows release, if none exist I would update the PR to link to https://archive.torproject.org/

PR #3549, opened 2022-05-04
ffmpeg-yt-dlp: Add version 5.0.1-5-20220503

issaclin32:
closes https://github.com/ScoopInstaller/Extras/issues/7788

[FFMpeg-yt-dlp](https://github.com/yt-dlp/FFmpeg-Builds) contains FFMpeg builds of latest release branch with patches necessary for smooth integration with yt-dlp.

**NOTES**:
* The repo provides both latest **master** and **release** branches of *FFMpeg*. I choose the release branch because it is more stable.
* Stability of the url / hash: I think it is safe to add this to the bucket because:
    (1) Each new build is now put in a new release tag.
    (2) Links for old builds are still valid.
See [release page](https://github.com/yt-dlp/FFmpeg-Builds/releases) for details.

### DIFF
--- a/bucket/ffmpeg-yt-dlp.json
+++ b/bucket/ffmpeg-yt-dlp.json
@@ -1,0 +1,43 @@
+{
+    "version": "5.0.1-5-20220503",
+    "description": "FFMpeg builds of latest release branch with patches necessary for smooth integration with yt-dlp",
+    "homepage": "https://github.com/yt-dlp/FFmpeg-Builds",
+    "license": "MIT",
+    "suggest": {
+        "yt-dlp": "yt-dlp"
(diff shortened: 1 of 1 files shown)


issaclin32:
> How frequently will the manifest get updated?

It will be updated per day.
This will not cause hash errors because the manifest searches for the tag that contains build time (which still contains the latest build of the binary) instead of the "latest" tag.

rashil2000:
Hmm, IMO it updates too frequently for a stable package. How about we add it to Versions?

issaclin32:
superseded by https://github.com/ScoopInstaller/Versions/pull/507

PR #3554, opened 2022-05-05
sccache: Update to version 0.3.0

Redfire75369:
This PR updates the version of sccache from v0.2.14, released on 2020-12-22, to v0.2.15, released on 2021-01-13.

There will likely be a PR soon for v0.3.0, but prebuilt binaries are not yet available on the GitHub release for that.

### DIFF
--- a/bucket/sccache.json
+++ b/bucket/sccache.json
@@ -16,7 +16,7 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/mozilla/sccache/releases/download/v$version/sccache-v$version-x86_64-pc-windows-msvc.tar.gz",
-                "extract_dir": "sccache-$version-x86_64-pc-windows-msvc"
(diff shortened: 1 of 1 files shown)


Redfire75369:
I've updated it with v0.3.0, released on 2022-05-06. The prebuilt releases just got published.

rashil2000:
`"checkver": "github",` should also work here

Redfire75369:
I've done that, but I've found a slight issue with downloading specific versions. Because of the change in tag conventions in the releases, one has to type `0.x.y` until `0.2.14` (ex. `sccache@0.2.12`), but one has to type `v0.x.y` past that (ex. `sccache@0.2.15`). Is there any way to make that less annoying or is that not possible with how the checkver/autoupdate is set up?

rashil2000:
`"checkver": "github",` will take care of that. The default regex has `v|V` as an optional character. Regarding the URL, I believe they will not change the release format very frequently, so the manual process is fine as of now.

PR #3581, opened 2022-05-18
fix lynx depends: update the slproweb download url

JoaquinRChen:
fix lynx depends: update the slproweb download url

Closes #XXXX

Relates to #XXXX


tech189:
Should we make lynx depend on `openssl1-light` from the Versions bucket? That way we don't have to manually update it in the lynx manifest.

rashil2000:
We can't add a hard dependency on a different bucket.

JoaquinRChen:
> We can't add a hard dependency on a different bucket.
> 
> Regardless, lynx doesn't even work, so I'm debating whether to even keep it or deprecate it

Hi, it works well here?
image

PR #3588, opened 2022-05-18
wrangler2: Add version 2.0.5

JamesCullum:
Wrangler2 is the v2 family of the Cloudflare tool called "wrangler". It is not the same codebase and is maintained seperately, hence I would propose to keep the original "wrangler" for now. [List of differences](https://developers.cloudflare.com/workers/wrangler/compare-v1-v2/)


rashil2000:
Hi! 

Thanks for the PR. However, we don't accept language-specific packages here (npm, pip, cargo etc.). If the upstream repos release the tool as a single packaged binary, then it is acceptable.

PR #3624, opened 2022-05-28
avr-gcc: Manually shim avr-gcc-$version.exe

### LINKED ISSUE #3623: avr-gcc (12.1.0) failed to create shim for avr-gcc-11.1.0
## Bug Report

avr-gcc (12.1.0)

### Current Behaviour

```
Creating shim for 'avr-gcc-11.1.0'.
Can't shim 'bin\avr-gcc-11.1.0.exe': File doesn't exist.
```

### Expected Behaviour

emmm... just do it maybe.

### Additional context/output

clearly the cause is the hard coded version number in the file name, which has been changed to avr-gcc-12.1.0.exe.

### Possible Solution

Since I have barely seen anyone used this file, no damage will occur after just ignoring the file perhaps.

### System details

Windows 11 x64

PowerShell 7.2.2

```

### PR DESCRIPTION
**Pull request by issaclin32**
* Fixes #3623

* Both `avr-gcc.exe` and `avr-gcc-$version.exe` exists in the package.
   Although they are the same thing (SHA-256 is the same), I think `avr-gcc-$version.exe` exists for some reason.
  We should keep it as it is, rather than just ignoring `avr-gcc-$version.exe`.

### DIFF
--- a/bucket/avr-gcc.json
+++ b/bucket/avr-gcc.json
@@ -56,7 +56,7 @@
         "}",
         "$shim_exe = \"$scoopdir\\apps\\scoop\\current\\supporting\\shims\\$shim_build\\shim.exe\"",
         "Copy-Item $shim_exe \"$scoopdir\\shims\\avr-gcc-$version.exe\"",
-        "Set-Content \"$scoopdir\\shims\\avr-gcc-$version.shim\" \"path = `\"$dir\\avr-gcc-$version.exe`\"\" -Encoding UTF8"
+        "Set-Content \"$scoopdir\\shims\\avr-gcc-$version.shim\" \"path = `\"$dir\\bin\\avr-gcc-$version.exe`\"\" -Encoding UTF8"
     ],
     "uninstaller": {
         "script": "Remove-Item \"$scoopdir\\shims\\avr-gcc-$version.exe\", \"$scoopdir\\shims\\avr-gcc-$version.shim\""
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rashil2000** [comment]
The whole process can be simplified by `scoop shim add <name> <path>` and `scoop shim rm <name>` 😅

**Comment by issaclin32** [comment]
> The whole process can be simplified by `scoop shim add <name> <path>` and `scoop shim rm <name>` 😅

Oh... Thanks for the info.

File: bucket/avr-gcc.json [hunk 1]
```diff
@@ -47,6 +46,13 @@
         "bin\\avr-strings.exe",
         "bin\\avr-strip.exe"
     ],
+    "post_install": [
+        "# manually shim bin\\avr-gcc-$version.exe",
+        "scoop shim add \"avr-gcc-$version.exe\" \"$dir\\bin\\avr-gcc-$version.exe\""
```
rashil2000, line 51:
```suggestion
        "scoop shim add \"avr-gcc-$version\" \"$dir\\bin\\avr-gcc-$version.exe\""
```

bucket/avr-gcc.json
```diff
@@ -47,6 +46,13 @@
...
+    ],
+    "uninstaller": {
+        "script": "scoop shim rm \"avr-gcc-$version.exe\""
```
rashil2000, line 54:
```suggestion
        "script": "scoop shim rm \"avr-gcc-$version\""
```

rashil2000 (approved):

PR #3664, opened 2022-06-09
ant@1.10.12: Persist user libs

### LINKED ISSUE #3662: Ant: Import user persisted libs in $persist_dir/lib
## Bug Report

**Package Name:** ant

### Current Behaviour

No way to add user defined libs into `$dir/lib` that persist across updates.

### Expected Behaviour

Allow user persisted libs not included in the default ant install to be copied into the `$dir/lib` folder.

### Additional context/output

Some libs, such as `ant-contrib` are extras that don't come with the default `ant` install.

### Possible Solution

As suggested in https://github.com/ScoopInstaller/Scoop/discussions/4973#discussion-4128671, this can be done via a `pre_install` script that copies the user libs from `$persist_dir/lib`.

I have a PR ready for this, but as instructed by the PR template I'm opening this issue first.  For reference, here is my suggested change: https://github.com/dstcruz/Scoop-Main/commit/441fca7193edbe8dadabfb14848504eb106d4fbe

dstcruz:
Import any extra libs persisted by the user in $persist_dir\lib, such as ant-contrib, that
are not part of the default ant install.

Closes #3662

### DIFF
--- a/bucket/ant.json
+++ b/bucket/ant.json
@@ -11,7 +11,9 @@
     "extract_dir": "apache-ant-1.10.12",
     "pre_install": [
         "if (Test-Path \"$persist_dir\\lib\") {",
-        "    Copy-Item \"$persist_dir\\lib\\*\" \"$dir\\lib\"",
+        "    Get-ChildItem \"$persist_dir\\lib\" | ForEach-Object {",
+        "        if (!(Test-Path \"$dir\\lib\\$($_.Name)\")) { Copy-Item $_ \"$dir\\lib\" }",
+        "    }",
         "}"
     ],
     "env_add_path": "bin",
--- a/bucket/ant.json
+++ b/bucket/ant.json
@@ -10,15 +10,29 @@
     "hash": "sha512:7e6fbcc3563df4bd87c883ad86a161a71da2774e0ed71a1b3aad82cbff1a7656ed9a0acb5ce40652129376dfd79f1ef74ec3369c1067d412a63062fea62ceccd",
     "extract_dir": "apache-ant-1.10.12",
     "pre_install": [
+        "# Record built-in libs in builtin_libs.json",
+        "$builtin_libs = Get-ChildItem \"$dir\\lib\" | Select -ExpandProperty Name",
+        "$builtin_libs | ConvertTo-Json | Out-File \"$dir\\lib\\builtin_libs.json\" -Encoding Ascii",
+        "# Copy user libs to $dir",
         "if (Test-Path \"$persist_dir\\lib\") {",
-        "    Get-ChildItem \"$persist_dir\\lib\" | ForEach-Object {",
-        "        if (!(Test-Path \"$dir\\lib\\$($_.Name)\")) { Copy-Item $_ \"$dir\\lib\" }",
+        "    Get-ChildItem \"$persist_dir\\lib\" | Select -ExpandProperty Name | ForEach-Object {",
+        "        if (!(Test-Path \"$dir\\lib\\$_\")) { Copy-Item \"$persist_dir\\lib\\$_\" \"$dir\\lib\" }",
         "    }",
         "}"
     ],
     "env_add_path": "bin",
     "env_set": {
         "ANT_HOME": "$dir"
+    },
+    "uninstaller": {
+        "script": [
+            "# Only persist libs that were added by the user, but not built-in ones",
+            "ensure \"$persist_dir\\lib\" | Out-Null",
+            "$builtin_libs = Get-Content \"$dir\\lib\\builtin_libs.json\" | ConvertFrom-Json",
+            "Get-ChildItem \"$dir\\lib\" -Exclude builtin_libs.json | Select -ExpandProperty Name | ForEach-Object {",
+            "    if (!($builtin_libs -contains \"$_\")) { Copy-Item \"$dir\\lib\\$_\" \"$persist_dir\\lib\" }",
+            "}"
+        ]
     },
     "checkver": {
         "url": "https://ant.apache.org/bindownload.cgi",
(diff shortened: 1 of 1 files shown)


bucket/ant.json
```diff
@@ -9,6 +9,11 @@
     "url": "https://www.apache.org/dist/ant/binaries/apache-ant-1.10.12-bin.zip",
     "hash": "sha512:7e6fbcc3563df4bd87c883ad86a161a71da2774e0ed71a1b3aad82cbff1a7656ed9a0acb5ce40652129376dfd79f1ef74ec3369c1067d412a63062fea62ceccd",
     "extract_dir": "apache-ant-1.10.12",
+    "pre_install": [
+        "if (Test-Path -PathType Container \"$persist_dir\\lib\") {",
+        "    Copy-Item \"$persist_dir\\lib\\*\" \"$dir\\lib\"",
+        "}"
+    ],
```
issaclin32, line 16:
```suggestion
    ],
    "uninstaller": {
        "script": [
            "ensure \"$persist_dir\\lib\" | Out-Null",
            "Copy-Item \"$dir\\lib\\*\" \"$persist_dir\\lib\""
        ]
    },
```

You'll also need to copy the files to `$persist_dir` when uninstalling.

bucket/ant.json
```diff
@@ -9,6 +9,11 @@
     "url": "https://www.apache.org/dist/ant/binaries/apache-ant-1.10.12-bin.zip",
     "hash": "sha512:7e6fbcc3563df4bd87c883ad86a161a71da2774e0ed71a1b3aad82cbff1a7656ed9a0acb5ce40652129376dfd79f1ef74ec3369c1067d412a63062fea62ceccd",
     "extract_dir": "apache-ant-1.10.12",
+    "pre_install": [
+        "if (Test-Path -PathType Container \"$persist_dir\\lib\") {",
```
issaclin32, line 13:
```suggestion
        "if (Test-Path \"$persist_dir\\lib\") {",
```
`-PathType Container` looks redundant. I cannot think of any case that `lib` can be a file.

issaclin32:
I am not familiar with **Ant**'s workflow,
but when copying from `$persist_dir` to `$dir`, would it be a good idea to **skip overwrite**?

This will make the library in `lib` folder **keep up-to-date** if the library exists in official package. Otherwise we will be overwriting the new libraries (in `$dir`) with the old one (in `$persist_dir`).

To do this, we can use something like:
```
    "pre_install": [
        "if (Test-Path \"$persist_dir\\lib\") {",
        "    Get-ChildItem \"$persist_dir\\lib\" | ForEach-Object {",
        "        if (!(Test-Path \"$dir\\lib\\$_\")) { Copy-Item \"$persist_dir\\lib\\*\" \"$dir\\lib\" }",
        "    }",
        "}"
    ],
```

bucket/ant.json
dstcruz: I'm not sure we want to do this. Ant has most of its version specific implementations in `$dir/lib`. This would not be copying out only user-specified libs, but _all_ libs.

bucket/ant.json
dstcruz: Accepting this. The only reason I put it there was to be explicit about what we are looking for: a dir.

bucket/ant.json
issaclin32: Oh. I see, 
but my concern is that many users will just put their libs at `$dir\libs` instead of `$persist_dir\libs`, and see their libs disappear after updating the app.

A possible way to solve this is by putting **a list of built-in libs** in a JSON file, then, avoid copying them in `uninstaller`.

```
    "pre_install": [
        "# Copy user libs to $dir",
        "if (Test-Path \"$persist_dir\\lib\") {",
        "    Get-ChildItem \"$persist_dir\\lib\" | Select -ExpandProperty Name | ForEach-Object {",
        "        if (!(Test-Path \"$dir\\lib\\$_\")) { Copy-Item \"$persist_dir\\lib\\$_\" \"$dir\\lib\" }",
        "    }",
        "}",
        "# Record built-in libs in builtin_libs.json",
        "$builtin_libs = Get-ChildItem \"$dir\\lib\" | Select -ExpandProperty Name",
        "$builtin_libs | ConvertTo-Json | Out-File \"$dir\\builtin_libs.json\" -Encoding Ascii"
    ],
    "uninstaller": {
        "script": [
            "# Only persist libs that are added by the user, but not built-in ones",
            "ensure \"$persist_dir\\lib\" | Out-Null",
            "$builtin_libs = Get-Content \"$dir\\builtin_libs.json\" | ConvertFrom-Json",
            "Get-ChildItem \"$dir\\lib\" | Select -ExpandProperty Name | ForEach-Object {",
            "    if (!($builtin_libs -contains \"$_\")) { Copy-Item \"$dir\\lib\\$_\" \"$persist_dir\\lib\" }",
            "}"
        ]
    },
```

*p.s. The `Select -ExpandProperty Name` mainly exists to avoid errors when there are *spaces* in SCOOP_DIR.*

dstcruz: This works. I'll implement that. However, I'll record the list of built-in libs before copying the user-defined ones, instead of the other way around.

dstcruz: I'll leave the `-ExpandProperty` as suggested, but would the fact that we put quotes around the file names when we refer to them avoid the issue with spaces in the path?

issaclin32: Sorry that I didn't say it clearly. I tried to use `$_` in the script (which is a *DirectoryInfo* object), but error occur when I test it on a VM with Scoop directory that contains spaces.

issaclin32 (approved):

rashil2000 (approved):

PR #3669, opened 2022-06-13
winget: Remove leading period in version

rasa:
Seems they screwed up the latest tag.


niheaven:
Waiting for their correct version number should be the best way :cry:

Or the RegEx was used once (for current wrong version number).

rasa:
Yeah, I didn't think this out. Waiting is best.

PR #3693, opened 2022-06-21
(chore): Fix git conflicts with line endings

### LINKED ISSUE #3692: [Bug] error: cannot pull with rebase: You have unstaged changes.
#### Current Behavior
Sometimes, there will be some change in the working directory of a bucket repository. That blocks bucket updates.

Furthermore, the green text says `Scoop was updated successfully!` which I'm not sure I agree with.

#### Expected Behavior
As I understand it, the bucket repositories aren't meant for interaction so updating should be more forceful and discard local changes without asking.

#### Additional context/output

It can look like this (don't mind the “unsafe repository” stuff):

image

In this case, it was caused by these files:

image

But I think I've seen others as well previously.

#### Possible Solution

- `git reset --hard` first. I think there are edge cases where this is still problematic, like when there are path name collisions due to case insensitivity on Windows.
- use `git fetch && git checkout -f origin/master` or similar (relies on knowing which branch we want though)

### System details

**Windows version:** 10

**OS architecture:** 64bit

**PowerShell version:** 7.2.4

**Additional software:** git version 2.36.1.windows.1

#### Scoop Configuration
```json
{
  "lastupdate": "2022-06-21T08:15:59.0239435+02:00",
  "SCOOP_REPO": "https://github.com/lukesampson/scoop",
  "SCOOP_BRANCH": "master",
  "show_update_log": false
}

```

rashil2000:
Closes #3692

Relates to https://github.com/ScoopInstaller/Extras/pull/8285

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,10 +12,6 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{[Bb][Aa][Tt],[Cc][Mm][Dd]}]
-# DOS/Win *requires* BAT/CMD files to have CRLF newlines
-end_of_line = crlf
-
 [[Mm]akefile{,.*}]
 # TAB-style indentation
 indent_style = tab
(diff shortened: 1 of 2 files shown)


issaclin32:
related: https://github.com/ScoopInstaller/Extras/issues/7546

PR #3712, opened 2022-06-26
yt-dlp: Persistant configuration file

pukkandan:

[Configuration file](https://github.com/yt-dlp/yt-dlp#configuration) for yt-dlp can be placed alongside the executable in addition to [youtube-dl's supported locations](https://github.com/ytdl-org/youtube-dl#configuration). This is especially usefull for portable installs like using Scoop. So this PR adds `yt-dlp.conf` as a persistent file

I have also removed `vcredist2010` from dependencies since yt-dlp no longer uses it. Let me know if you'd rather have this in a separate PR

### DIFF
--- a/bucket/yt-dlp.json
+++ b/bucket/yt-dlp.json
@@ -17,6 +17,12 @@
         }
     },
     "bin": "yt-dlp.exe",
+    "pre_install": [
+        "$manifest.persist | ForEach-Object {",
+        "    if (-not (Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" -ItemType file | Out-Null }",
+        "}"
+    ],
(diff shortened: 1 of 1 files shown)


bucket/yt-dlp.json
```diff
@@ -18,6 +17,12 @@
         }
     },
     "bin": "yt-dlp.exe",
+    "pre_install": [
+        "$manifest.persist | ForEach-Object {",
+        "    if (-not (Test-Path \"$persist_dir\\$_\")) { New-Item \"$dir\\$_\" -ItemType file | Out-Null }",
+        "}"
+    ],
```
pukkandan, line 24: This was copied from `everything.json` to prevent scoop from creating a directory called `yt-dlp.conf`. I don't know if there is another proper way to do this

PR #3717, opened 2022-06-28
grep: Rename binary before shimming

sitiom:
Rename binary to `grep.exe` so that the filename that grep shows up is the same as its command.

Before:
image
After:
image

### DIFF
--- a/bucket/grep.json
+++ b/bucket/grep.json
@@ -23,7 +23,7 @@
             "extract_dir": "tools\\install\\grep-windows-3.7"
         }
     },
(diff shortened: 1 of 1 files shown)


bucket/grep.json
```diff
@@ -8,24 +8,22 @@
...
-                    "grep-3.7-x64.exe",
+                    "grep.exe",
                     "fgrep",
                     "-F"
                 ]
             ],
             "extract_dir": "tools\\install\\grep-windows-3.7"
         }
     },
+    "pre_install": "Move-Item \"$dir\\grep-$version-x64.exe\" \"$dir\\grep.exe\"",
```
niheaven, line 26: Why not using `Rename-Item`?

niheaven (approved):

PR #3727, opened 2022-06-30
clash.meta: Add version 1.11.2

issaclin32:
**[Clash.Meta](https://github.com/MetaCubeX/Clash.Meta)** is a fork of Clash ([main/clash](https://github.com/ScoopInstaller/Main/blob/master/bucket/clash.json)), a rule-based tunnel in Go.

**NOTES**:
* Currently the repo has **794 stars** and 4K forks.
* config location: `$Env:UserProfile\.config\clash`.

### DIFF
--- a/bucket/clash.meta.json
+++ b/bucket/clash.meta.json
@@ -1,6 +1,6 @@
 {
     "version": "1.11.2",
-    "description": "Fork of Clash, a rule-based tunnel in Go.",
+    "description": "Fork of Clash, a rule-based tunnel.",
     "homepage": "https://github.com/MetaCubeX/Clash.Meta",
     "license": "GPL-3.0-or-later",
     "architecture": {
--- a/bucket/clash.meta.json
+++ b/bucket/clash.meta.json
@@ -1,0 +1,35 @@
+{
+    "version": "1.11.2",
+    "description": "Fork of Clash, a rule-based tunnel in Go.",
+    "homepage": "https://github.com/MetaCubeX/Clash.Meta",
+    "license": "GPL-3.0-or-later",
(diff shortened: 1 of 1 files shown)


bucket/clash.meta.json
```diff
@@ -0,0 +1,35 @@
...
+            "url": "https://github.com/MetaCubeX/Clash.Meta/releases/download/v1.11.2/Clash.Meta-windows-386-v1.11.2.zip",
+            "hash": "396a7d2295e93700eca433f277b52eb82e797a3426d57b183d6ceb40b9ae66aa"
+        }
+    },
+    "pre_install": "Get-ChildItem \"$dir\\Clash.Meta*.exe\" | Rename-Item -NewName 'Clash.Meta.exe'",
+    "bin": [
+        "Clash.Meta.exe",
+        [
+            "Clash.Meta.exe",
+            "clash"
```
rashil2000, line 21: Wouldn't this conflict with main/clash's binaries?

rashil2000 (approved):

bucket/clash.meta.json
issaclin32: Yes, but my thought is,
Since *Clash.Meta* is a fork of *Clash*:

(1) It should be a rare case for user to install both of them. Even if so, they can run `scoop reset clash` to solve the problem. (so that the shim `clash` points to *Clash*, and `clash.meta` points to *Clash.Meta*)

(2) Users will expect the shim `clash` to work for *Clash.Meta* as well.

rashil2000: Hmm, understood

PR #3754, opened 2022-07-16
qrencode: Add version 4.1.1

sitiom:
Commonly used CLI utility to generate QR Codes from text.

### DIFF
--- a/bucket/qrencode.json
+++ b/bucket/qrencode.json
@@ -1,0 +1,16 @@
+{
+    "version": "4.1.1",
+    "description": "Encode input data in a QR Code symbol",
+    "homepage": "https://fukuchi.org/works/qrencode/",
+    "license": "LGPL-2.1-only",
+    "url": "https://downloads.sourceforge.net/project/qrencode-for-windows/QREncode-4.1.1_Win32%28static%29.zip",
+    "hash": "sha1:eb3afc0f87bf9f1c9af143fcd205c29c07f605bf",
+    "extract_dir": "QREncode-4.1.1_Win32",
+    "bin": "qrencode.exe",
+    "checkver": {
+        "github": "https://github.com/fukuchi/libqrencode"
+    },
+    "autoupdate": {
+        "url": "https://downloads.sourceforge.net/project/qrencode-for-windows/QREncode-$version_Win32%28static%29.zip"
+    }
+}
--- a/bucket/qrencode.json
+++ b/bucket/qrencode.json
@@ -34,10 +34,11 @@
             "$base_url = 'https://packages.msys2.org/package/'",
             "$qrencode_page = (Invoke-WebRequest \"${base_url}libqrencode\" -UseBasicParsing).Content",
             "$runtime_page = (Invoke-WebRequest \"${base_url}msys2-runtime\" -UseBasicParsing).Content",
+            "$version_regex = \"Version:.+\\s+<.+>([\\d.-]+)\"",
             "",
-            "if(!($qrencode_page -match \"https://mirror.msys2.org/msys/x86_64/libqrencode-([\\d.-]+)-x86_64.pkg.tar.zst\")) { break }",
+            "if(!($qrencode_page -match $version_regex)) { error \"No match found for version\"; break }",
             "$scriptver = $matches[1]",
-            "if(!($runtime_page -match \"https://mirror.msys2.org/msys/x86_64/msys2-runtime-([\\d.-]+)-x86_64.pkg.tar.zst\")) { break }",
+            "if(!($runtime_page -match $version_regex)) { error \"No match found for version\"; break }",
             "$msys2_runtime_version = $matches[1]",
             "",
             "Write-Output $scriptver $msys2_runtime_version"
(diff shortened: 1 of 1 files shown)


rashil2000:
Is there any indication that the GitHub repo and SourceForge repo are officially related?

sitiom:
> Is there any indication that the GitHub repo and SourceForge repo are officially related?

No, but it's the only available source. I guess it's similar to [`grep`](https://github.com/ScoopInstaller/Main/blob/master/bucket/grep.json), which gets its source from the chocolatey package, which gets it from https://github.com/mbuilov/grep-windows/. Should I use the [msys2 source](https://packages.msys2.org/base/mingw-w64-qrencode) instead?

rashil2000:
Checking https://sourceforge.net/p/qrencode-for-windows/tickets/1/, I'm skeptical about the author releasing any 64bit binaries, or worse, any versions newer than 4.1.1 (when they come at the GitHub repo)...

sitiom:
Switched to [`libqrencode`](https://packages.msys2.org/package/libqrencode) from msys2. Additionally, that version does not print garbage on utf8 compared to the mingw-w64 version (https://github.com/fukuchi/libqrencode/issues/200)

bucket/qrencode.json
```diff
@@ -0,0 +1,53 @@
...
+                "https://mirror.msys2.org/msys/x86_64/msys2-runtime-3.3.5-3-x86_64.pkg.tar.zst"
+            ],
+            "hash": [
+                "ead222db63fadf4c6ec67f6b7a95761eb9d0c4ffd92e18129ab11f7944e1e583",
+                "9e99486501905294eb01d7024d2cb96f658b609232878a990e469b5448ea4e11"
+            ]
+        }
+    },
+    "pre_install": [
+        "$files = @('qrencode.exe', 'msys-qrencode-4.dll', 'msys-2.0.dll');",
```
rashil2000, line 19: We should also keep these files, for the sake of completeness:
```
/usr/share/licenses/libqrencode/COPYING
/usr/share/man/man1/qrencode.1.gz
```

sitiom: Done

sitiom:
PNG output seems to be disabled, which is available in the MinGW version (https://github.com/msys2/MSYS2-packages/issues/2214). But the MinGW version cannot display utf8 properly. I guess that is more important, and you can alternatively export it to SVG and use something like ImageMagick to convert it to PNG.
image

rashil2000:
I did some testing around: https://github.com/msys2/MSYS2-packages/issues/2214#issuecomment-1186512575

sitiom:
Mingw or msys 🤔

rashil2000:
Mingw (you can check the command's path, it starts with `/mingw64/bin`)

sitiom:
I mean, I was wondering what to use for this manifest, considering the drawbacks of each option 😅

rashil2000:
The mingw one has multiple dependencies so it'll be very difficult to manage.

Hmm, I guess the current manifest (msys2) is our only option now...

bucket/qrencode.json
```diff
@@ -0,0 +1,69 @@
...
+        "                \"$dir\\usr\\bin\\msys-qrencode-4.dll\",",
+        "                \"$dir\\usr\\bin\\msys-2.0.dll\",",
+        "                \"$dir\\usr\\share\\licenses\\libqrencode\\COPYING\",",
+        "                \"$dir\\usr\\share\\man\\man1\\qrencode.1.gz\")",
+        "",
+        "Get-ChildItem \"$dir\\usr\" -Recurse",
+        "    | Select-Object -ExpandProperty FullName",
+        "    | Where-Object {$keep_files -notcontains $_ }",
+        "    | Sort-Object Length -Descending",
+        "    | ForEach-Object { if ((Test-Path $_ -Type Container) -and (Get-ChildItem $_)) { return } Remove-Item $_ }"
```
rashil2000, line 30: There are three residual files remaining after this (in root folder).

I think this can be simplified by using `Get-ChildItem "$dir" -Recurse -Exclude $keep_files | Remove-Item -Force`

sitiom: > I think this can be simplified by using `Get-ChildItem "$dir" -Recurse -Exclude $keep_files | Remove-Item -Force`

It doesn't work when you try to specify files in folder and subfolders. `-Force` only removes hidden and read-only files. It doesn't stop the prompt and would require the `-Recurse` parameter. `Sort-Object Length -Descending` ensures that files are deleted first before folders are emptied out and can be deleted.

rashil2000: Ah, I see

sitiom:
> The mingw one has multiple dependencies so it'll be very difficult to manage.

I'm more inclined to switch to MinGW as it is future-proof (The latest Windows Terminal preview works, conhost might be fixed), has png support, and has 32-bit builds.

Is it also possible that other sub-dependencies are also introduced due to some dependencies bundling their own tools? I have reduced the dependencies to just libpng, libwinpthread, and zlib in a Sandbox and it seems to work:
image

sitiom:
A quick look using [dependencies](https://github.com/lucasg/Dependencies) tells us otherwise:

image

PR #3780, opened 2022-07-25
sets helix runtime path

abhijeetChawla:
This makes sure that all the theme and grammar are readily available when the user install helix
the personal config of config.toml and language.toml are still stored in AppData/Roaming/helix

### DIFF
--- a/bucket/helix.json
+++ b/bucket/helix.json
@@ -17,6 +17,9 @@
             "helix"
         ]
     ],
+    "env_set":{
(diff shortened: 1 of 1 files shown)


rashil2000:
This is not required, as helix automatically picks up the runtime directory besides its binary. You can confirm this by running `hx --health`.

abhijeetChawla:
So it would be more appropriate to move the runtime folder to %AppData%/helix` instead?

PR #3787, opened 2022-07-28
wapm-cli: Add version 0.5.4

### LINKED ISSUE #3785: WAPM is not installed with wasmer
https://github.com/ScoopInstaller/Main/blob/aa2eee1b1e4ca4008165ead195d29e85faf25543/bucket/wasmer.json#L16-L19

Hello, I would've liked WAPM to be installed with WASMER but I see that it is not added by default. Can you please add it to the WASMER version?

Edit: Wax can't even run because it requires WAPM, which is not installed by default.

@syrusakbary
@Ash258
@r15ch13 
@HUMORCE

HUMORCE:
Closes #3785

### DIFF
--- a/bucket/wapm-cli.json
+++ b/bucket/wapm-cli.json
@@ -1,0 +1,29 @@
+{
+    "version": "0.5.4",
+    "description": "WebAssembly package manager.",
+    "homepage": "https://wapm.io/",
+    "license": "MIT",
+    "suggest": {
+        "wasmer": "wasmer"
+    },
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/wasmerio/wapm-cli/releases/download/v0.5.4/wapm-cli-windows-amd64.tar.gz",
+            "hash": "8f198c2c31b435152eda047035c2e49d4f9cf8d2d95714d37521590392b0ecda"
+        }
+    },
+    "bin": [
+        "bin\\wapm.exe",
+        "bin\\wax.cmd"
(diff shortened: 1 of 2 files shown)


wmstack:
Hmm, I believe that something would be still off about `wasmer`'s app manifest: last I tried the `wax` cli doesn't function unless `wapm` is installed.

So `wapm-cli` should be more than just a suggestion, it should be a requirement for the correct functioning of wasmer. I believe that, since people might want to use wasmer by itself, that wax should be installed with the `wapm-cli`, and `wasmer` should be installed alone.

HUMORCE:
it's two things, imo.

wmstack:
> it's two things, imo.

Not sure what you mean, do you mean that wax should be a separate manifest? Well, sure. I guess why not. Consider adding it as another suggestion to wasmer as well.

HUMORCE:
I mean that `wapm-cli` should not be a dependency of `wasmer`. it's all.

wmstack:
I hadn't seen that the `wax` cli had moved to the wasm manifest. All good.

rashil2000 (approved):

wmstack:
What is up with the windows powershell checks? Is this special for this pull request or does this always happen?

rashil2000:
> What is up with the windows powershell checks? Is this special for this pull request or does this always happen?
> 
> Edit: I mean why they are failing.

Upstream issue, can be ignored for now

PR #3802, opened 2022-08-03
edgedb: Add edgedb version 2.0.2+61917d6

r15ch13:
Add edgedb version 2.0.2+61917d6

### DIFF
--- a/bucket/edgedb.json
+++ b/bucket/edgedb.json
@@ -1,0 +1,28 @@
+{
+    "version": "2.0.2+61917d6",
+    "description": "A graph-relational database",
+    "homepage": "https://www.edgedb.com/",
+    "license": "Apache-2.0|MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://packages.edgedb.com/archive/x86_64-pc-windows-msvc/edgedb-cli-2.0.2+61917d6.exe#/edgedb.exe",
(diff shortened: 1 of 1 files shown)


bucket/edgedb.json
```diff
@@ -0,0 +1,28 @@
...
+        "jsonpath": "$..packages[-1:].version"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://packages.edgedb.com/archive/x86_64-pc-windows-msvc/edgedb-cli-$version.exe#/edgedb.exe"
+            }
+        },
+        "hash": {
+            "url": "https://packages.edgedb.com/archive/.jsonindexes/x86_64-pc-windows-msvc.json",
```
rashil2000, line 24: This URL seems x86_64 specific, so shouldn't it go inside the architecture block?

r15ch13: It's sometimes inside and sometimes it is not. :smile: Examples: 
https://github.com/ScoopInstaller/Main/blob/master/bucket/containerd.json
https://github.com/ScoopInstaller/Main/blob/master/bucket/chronograf.json

rashil2000: Ah yes, pre-existing inconsistencies are there 😅

But I try to put architecture-specific URLs inside, and generic URLs outside, as a convention.

PR #3809, opened 2022-08-08
chore: Update Info-Zip licenses

sitiom:
[Info-ZIP](https://spdx.org/licenses/Info-ZIP.html) has its own entry in the SPDX.

### DIFF
--- a/bucket/unzip.json
+++ b/bucket/unzip.json
@@ -2,10 +2,7 @@
     "version": "6.00",
     "description": "Unzip compression utility",
     "homepage": "http://www.info-zip.org/",
-    "license": {
-        "identifier": "Freeware",
-        "url": "http://infozip.sourceforge.net/doc/LICENSE"
-    },
+    "license": "Info-ZIP",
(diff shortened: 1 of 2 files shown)


bucket/unzip.json
```diff
@@ -2,10 +2,7 @@
     "version": "6.00",
     "description": "Unzip compression utility",
     "homepage": "http://www.info-zip.org/",
-    "license": {
-        "identifier": "Freeware",
-        "url": "http://infozip.sourceforge.net/doc/LICENSE"
-    },
```
rasa, line 8: Do we have a standard practice/rule to leave off the url if it's a known SPDX license? The url in this case would be http://infozip.sourceforge.net/license.html found on https://spdx.org/licenses/Info-ZIP.html.

sitiom: That's what most of the manifests in the official buckets do, so I'm just following it for consistency.

rasa (approved):

PR #3817, opened 2022-08-11
forcebindip: Add checkver

brian6932:
I don't see why not 🤷


rashil2000:
Checkver is removed from manifests which haven't been updated in a very long time (5 years in this case). This saves Excavator bandwidth.

So we don't need it here.

PR #3845, opened 2022-08-21
rustup: Update note

b4imetu:
Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/rustup.json
+++ b/bucket/rustup.json
@@ -3,7 +3,13 @@
     "description": "Manage multiple rust installations with ease",
     "homepage": "https://rustup.rs",
     "license": "MIT|Apache-2.0",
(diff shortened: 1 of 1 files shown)


rashil2000:
Actually, the note from this manifest should be removed entirely, because MSVC tools are not required for the GNU toolchain

b4imetu:
> Actually, the note from this manifest should be removed entirely, because MSVC tools are not required for the GNU toolchain

🤣So that's it. It can be replaced with other tool dependencies

rashil2000:
What do you mean?

b4imetu:
> What do you mean?

[Link](https://www.reddit.com/r/rust/comments/a63dlt/comment/ebrbi8d/)
I thought rust-gnu was dependent on the installation of MinGW to run, but it turns out that it can be run directly🤣
Never mind. I will delete the notes🤣

PR #3858, opened 2022-08-25
Add Determined AI CLI manifest

sirredbeard:
Adds the [Determined AI](https://github.com/determined-ai/determined) CLI manifest to Main

Determined has 1.8k ⭐

This manifest makes use of a [simple pass-through determined.exe file](https://github.com/sirredbeard/det-helper) but installs the main components via Python and pip.

[Auto-update function check](https://gist.github.com/sirredbeard/a615dc1276aa029f9584617d12ce52f8) passes.

### DIFF
--- a/bucket/determined.json
+++ b/bucket/determined.json
@@ -1,0 +1,28 @@
+{
+    "version": "0.19.1",
+    "description": "The Determined CLI is a command line tool that lets you launch new experiments and interact with a Determined cluster.",
+    "homepage": "https://github.com/determined-ai/determined",
+    "license": "Apache-2.0",
+    "depends": "python",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/sirredbeard/det-helper/releases/download/0.0.1/determined.exe",
+            "hash": "B8DAAB778AC4E657F5DB909BC541C1D286B9A8E186C438079A29C82872150C01"
+        }
+    },
+    "installer": {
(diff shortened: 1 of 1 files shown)


rashil2000:
Hi!

We don't accept language-specific packages here in the Main bucket, because they are better managed by their respective package managers - PIP in this case. Unless the package is specifically distributed as a single, independent binary, that is.

sirredbeard:
It is not distributed as a separate executable. Would it be accepted into extras?

rashil2000:
Sorry, neither in Extras - the reason is the same - Scoop is not good at managing version-specific, transitive dependencies that PIP or NPM packages require, and installing them through Scoop will give users the wrong impression that it is indeed managing those dependencies itself.

sirredbeard:
I see. In this case when a new release is tagged in GitHub, the CLI is also [pushed to PyPi](https://github.com/determined-ai/determined/blob/74dd27f393fcbccf1cb27b9419aed5e63e9193d8/.circleci/config.yml#L1173), so detection of a new release on GitHub would be a valid signal to upgrade via pip.

sirredbeard:
I have created a third-party [Scoop bucket](https://github.com/sirredbeard/determined-bucket) and noted the issue you raised.

rashil2000:
Thanks for the info! People searching for Determined package on Scoop will find the thread helpful!

PR #3867, opened 2022-09-02
fix: lf: add lfcd to shims

### LINKED ISSUE #3796: [Request] lf: add lfcd to shims
[lf](https://github.com/gokcehan/lf) [manifest](https://github.com/ScoopInstaller/Main/blob/master/bucket/lf.json) should add [lfcd.ps1](https://github.com/gokcehan/lf/blob/master/etc/lfcd.ps1) and/or [lfcd.cmd](https://github.com/gokcehan/lf/blob/master/etc/lfcd.cmd) (there are also versions for other shells) to shims. `lfcd` provides a way to change working directory to the last directory in `lf` on exit. Which is a common [use case](https://github.com/gokcehan/lf/issues/631).

Valinor:
see the issue for details on the typos fixed

fixes issue #3796 

Closes #3796 

### DIFF
--- a/bucket/lf.json
+++ b/bucket/lf.json
@@ -5,17 +5,35 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/gokcehan/lf/releases/download/r27/lf-windows-amd64.zip",
-            "hash": "118d6f2096e35ea7703f2b881a9d724806af8a5575c7f788d723f9797fa20948"
+            "url": [
+                "https://github.com/gokcehan/lf/releases/download/r27/lf-windows-amd64.zip",
+                "https://raw.githubusercontent.com/gokcehan/lf/r27/etc/lfcd.cmd",
+                "https://raw.githubusercontent.com/gokcehan/lf/r27/etc/lfcd.ps1"
+            ],
+            "hash": [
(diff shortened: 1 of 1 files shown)


rashil2000:
Would be better to download these in the URL field itself. The ps1 script should also be added

rashil2000:
We should copy the .cmd and .ps1 scripts (as you had done earlier), so that they are directly accessible and don't get overwritten

Valinor:
> We should copy the .cmd and .ps1 scripts (as you had done earlier), so that they are directly accessible and don't get overwritten

instead of use post_install, I use the **bin** part to create shims from cmd and ps1, isn't that a better solution?

rashil2000:
No, because:
- `lfcd.cmd` will create `lfcd.cmd` and `lfcd` (sh script)
- `lfcd.ps1` will create `lfcd.cmd` (which will overwrite the one created above), `lfcd` (sh script) ((which will again overwrite the one created above) and `lfcd.ps1`

So you see there'll be no use of creating shell-specific shims this way. That's why we need to copy them instead.

tech189:
@rashil2000 I restored the shim copy to `pre_install` and moved the removal to `post_uninstall`. It's in `post_uninstall` so that if you uninstall while `lf.exe` is still running and try again after closing it, Scoop won't complain that the shims don't exist because they've already been deleted. Could you take a look?

rashil2000:
Perfect!

PR #3882, opened 2022-09-07
nircmd: Add `nircmd.exe` to `bin`

sitiom:
> It's recommended to copy the executable of NirCmd (nircmd.exe) to your windows directory, or to any other folder listed in your PATH environment variable, so you won't need to type the full path of nircmd in each time that you want to use it.


rashil2000:
What's the difference between the two executables?

sitiom:
`nircmd.exe` is a windowless application while `nircmdc.exe` is a console application. Otherwise, they have the same command line arguments. `nircmd.exe` is used more to run commands in a hidden window.

PR #3888, opened 2022-09-07
v2ray: Fix install for v5

### LINKED ISSUE #3885: v2ray@5.1.0: install failed
Scoop Configuration

```json
{
  "lastUpdate": "2022-09-07T17:21:11.1113918+08:00",
  "rootPath": "D:\\Apps\\scoop",
  "globalPath": "D:\\ProgramData\\scoop",
  "SCOOP_REPO": "https://github.com/ScoopInstaller/Scoop",
  "SCOOP_BRANCH": "master",
  "aria2-enabled": false,
  "alias": {}
}
```

Ouput

```plain
> scoop install v2ray
Installing 'v2ray' (5.1.0) [64bit] from main bucket
Loading v2ray-windows-64.zip from cache
Checking hash of v2ray-windows-64.zip ... ok.
Extracting v2ray-windows-64.zip ... done.
Running installer script...
Verifying V2Ray binary...
&: The term 'D:\Apps\scoop\apps\v2ray\5.1.0\v2ctl.exe' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
V2Ray binary is corrupted!
```

AkinoKaede:
Closes #3885

### DIFF
--- a/bucket/v2ray.json
+++ b/bucket/v2ray.json
@@ -15,23 +15,19 @@
     },
     "installer": {
         "script": [
-            "#v2ctl.exe cannot recognize signature file when there are spaces in path. This is a workaround to eliminate spaces.",
+            "#v2ray.exe cannot recognize signature file when there are spaces in path. This is a workaround to eliminate spaces.",
             "Write-Host 'Verifying V2Ray binary...' -f Yellow",
             "$sig_dir = \"$Env:SystemRoot\\Temp\\v2ray\"",
             "ensure $sig_dir | Out-Null",
             "Invoke-WebRequest \"https://github.com/v2fly/v2ray-core/releases/download/v$version/Release\" -OutFile $sig_dir\\Release",
-            "if (!(Invoke-Expression \"&`\"$dir\\v2ctl.exe`\" verify --sig=$sig_dir\\Release `\"$dir\\v2ray.exe`\"\" | Select-String 'OK')) {",
+            "if (!(Invoke-Expression \"&`\"$dir\\v2ray.exe`\" verify --sig=$sig_dir\\Release `\"$dir\\v2ray.exe`\"\" | Select-String 'OK')) {",
             "    Write-Host 'V2Ray binary is corrupted!' -f Red",
             "    break",
             "}",
             "Remove-Item $sig_dir -Force -Recurse | Out-Null"
         ]
     },
-    "bin": [
-        "v2ctl.exe",
-        "v2ray.exe",
-        "wv2ray.exe"
-    ],
+    "bin": "v2ray.exe",
     "persist": "config.json",
     "checkver": {
         "github": "https://github.com/v2fly/v2ray-core"


rashil2000:
I think we can remove the `installer` field entirely. We are already verifying the package through its SHA512 hash.

AkinoKaede:
> I think we can remove the field entirely. We are already verifying the package through its SHA512 hash.`installer`

We use the signature to verfiy the identity of publisher.

bucket/v2ray.json
```diff
@@ -15,23 +15,19 @@
     },
     "installer": {
         "script": [
-            "#v2ctl.exe cannot recognize signature file when there are spaces in path. This is a workaround to eliminate spaces.",
+            "#v2ray.exe cannot recognize signature file when there are spaces in path. This is a workaround to eliminate spaces.",
             "Write-Host 'Verifying V2Ray binary...' -f Yellow",
-            "$sig_dir = \"$Env:SystemRoot\\Temp\\v2ray\"",
+            "$sig_dir = \"$Env:TEMP\\v2ray\"",
```
rashil2000, line 20: Was the issue with spaces in path fixed? Because $Env:TEMP can still contain space if the user's name has space.

rashil2000:
> > I think we can remove the field entirely. We are already verifying the package through its SHA512 hash.`installer`
> 
> We use the signature to verfiy the identity of publisher, so I think we should not remove it.

We are directly downloading the package from the publisher, so I don't get why this holds?

bucket/v2ray.json
AkinoKaede: sorry i didn't notice this.

AkinoKaede:
> > > I think we can remove the field entirely. We are already verifying the package through its SHA512 hash.`installer`
> > 
> > 
> > We use the signature to verfiy the identity of publisher, so I think we should not remove it.
> 
> We are directly downloading the package from the publisher, so I don't get why this holds?
The private key is only held by a small number of developers. For exmaple, I can create a release for V2Ray but I cannot sign it.

rashil2000:
The method for verification used by the script in the manifest is using SHA256 hashes - there's no key signing

AkinoKaede:
> The method for verification used by the script in the manifest is using SHA256 hashes (which is what the Release file contains) - there's no key signing

You can get the signature from line 2 of `Release` file.

rashil2000:
> You can get the signature from line 2 of `Release` file.

Which component's signature is that?

AkinoKaede:
> > You can get the signature from line 2 of `Release` file.
> 
> Which component's signature is that?

Files listed in the `Release` file.

rashil2000:
Sorry I'm really lost here: how do all of those files have a single signature?

AkinoKaede:
> Sorry I'm really lost here: how do all of those files have a single signature?

To be precise, the signature in the `Release` file is its own, and the file contains SHA256 hashes of other files.

rashil2000:
I have 2 further questions.


> To be precise, the signature in the `Release` file is its own, and the file contains SHA256 hashes and version of other files.

- This is not possible. A file cannot contain its own signature. Ref: https://security.stackexchange.com/questions/3851/can-a-file-contain-its-md5sum-inside-it. How are V2Ray developers achieving this?
- Since the Release file is downloaded separately as part of the `install` field - it is not really adding any layer of security with respect to verification of binaries, right? And since that signature on Line 2 is for the Release file (and not for v2ray.exe), it is useless (here, in this context) because we only verify v2ray.exe

AkinoKaede:
> I have 2 further questions.
> 
> > To be precise, the signature in the `Release` file is its own,
> 
> * This is not possible. A file cannot contain its own signature. Ref: https://security.stackexchange.com/questions/3851/can-a-file-contain-its-md5sum-inside-it. How are V2Ray developers achieving this?
> 
> > and the file contains SHA256 hashes and version of other files.
> 
> * Since the Release file is downloaded separately as part of the `install` field - it is not really adding any layer of security with respect to verification of binaries, right? And since that signature on Line 2 is for the Release file (and not for v2ray.exe), it is useless (here, in this context) because we only verify v2ray.exe
> * One more loophole is that you are using v2ray.exe to verify v2ray.exe _itself_, which kind of defeats the purpose, doesn't it?

1. The file does not contain its own signature, the first two lines are not included in the signature.
2. it can verify

AkinoKaede:
In fact, VSign (the sign tool for V2Ray) is an implemention of signify, which is a sign tool for OpenBSD.

<https://www.openbsd.org/papers/bsdcan-signify.html>

PR #3897, opened 2022-09-10
btop: Add version 1.0.3

brian6932:
btop-LHM: Add version 1.0.2

Popular system monitor on Linux, and new to Windows, since it is more than popular enough on the [original repo](https://github.com/aristocratos/btop), I thought PRing to main should work


bucket/btop-LHM.json
```diff
@@ -0,0 +1,32 @@
+{
```
rasa, line 1: We don't have many, if any, files with upper-case characters in the filenames. Can we lowercase these?

rasa (changes requested):

brian6932:
Sure no problem, I just copied the original names

rashil2000:
- Add only one variant - the LHM one (and rename it to btop.json)
- Use the `persist` field instead of manual copying. For first time install, create an empty conf file if it doesn't exist, like https://github.com/ScoopInstaller/Main/blob/87a6c5f0bb98ff3524e2b89957c07f96d5b4bfb1/bucket/mpxplay.json#L8

rasa:
> * Add only one variant - the LHM one (and rename it to btop.json)

@rashil2000 Given that one's of scoop's primary goals is to install/run apps as a non-admin user, I think keeping both versions makes sense. And keep the non-admin version be the more simply named manifest. Your thoughts?

rasa:
@rashil2000 Note that for the related btop4win, we *just* added a non-admin version at https://github.com/ScoopInstaller/Extras/pull/9200 and an admin version at https://github.com/ScoopInstaller/Versions/pull/635 . Is it our policy to push apps that require UAC to our Versions bucket?

brian6932:
The thing with using persist, is that the program generates it's own config, so if I make an empty file, it wouldn't be able to generate it, if there are no configs found, this is why I thought this approach was best

brian6932:
Ok after testing, it does indeed overwrite the blank file, I'll fix that

rashil2000:
> @rashil2000 Given that one's of scoop's primary goals is to install/run apps as a non-admin user, I think keeping both versions makes sense. And keep the non-admin version as the more simply named manifest. Your thoughts?

For any other application, I would agree. But this is a system resource monitor that contains information that shouldn't normally be viewable by an unprivileged user. Moreover, the non-admin version shows incomplete _and incorrect_ information (w.r.t. CPU freqs), which I feel is something that shouldn't be there in the Main bucket.

rashil2000:
Dang it, I did not see the PRs in the Extras and Versions bucket. I guess this one can be closed then.

brian6932:
Is this not something that is better fit for main though?

brian6932:
Also I think it might be a good idea to have a `pre_uninstall` where `btop.sys` is moved to something like temp or recycle bin, because it tends to cause some issues 🤔

brian6932:
With that it should uninstall without error now

brian6932:
If a different directory is preferred for this, for example the scoop cache directory, lmk

sitiom (changes requested):
> Is this not something that is better fit for main though? [Bottom is in main]

The main bucket has [certain criteria](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) and btop does not fit one of the criteria:
> a reasonably well-known and widely used developer tool (e.g. if it's a GitHub project, it should have at least 500 stars and 150 forks)

Alternate versions like this one should go to the [Versions](https://github.com/ScoopInstaller/Versions) bucket. Here is my already existing PR for it: https://github.com/ScoopInstaller/Versions/pull/635.

Additionally, the manifest should be named `btop4win` as the [project](https://github.com/aristocratos/btop4win) is separate from [btop](https://github.com/aristocratos/btop) and even the executable name is called `btop4win`.

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
...
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/aristocratos/btop4win/releases/download/v1.0.2/btop4win-LHM-x64.zip",
+            "hash": "898b0725918e58926851304195e795b881abd1c2353b0537e62ea1a273777a66",
+            "extract_dir": "btop4win"
+        }
+    },
+    "pre_install": [
+        "Get-ChildItem \"$dir\\btop4win.exe\" | Rename-Item -NewName { $_.name -Replace '4win', '' }",
```
sitiom, line 18: We should keep the original executable name (`btop4win.exe`) as is. If you like to add `btop` as a command, add it to `bin` instead as a shim alongside with `btop4win`.

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
...
+        "if (!(Test-Path \"$persist_dir\\btop.conf\")) { New-Item \"$dir\\btop.conf\" | Out-Null }"
+    ],
+    "pre_uninstall": [
+        "$sys = \"$dir\\btop.sys\"",
+        "if (!(Test-Path $sys)) { return }",
+        "$tempPath = '~\\appdata\\temp\\btop.sys'",
+        "if (Test-Path $tempPath) { $i = 0; While (Test-Path $tempPath) { $i++; $tempPath = \"$($tempPath | Split-Path)\\btop$i.sys\" } }",
+        "Move-Item $sys $tempPath"
+    ],
```
sitiom, line 27: I do not see why this is necessary

sitiom (changes requested):

brian6932:
>a reasonably well-known and widely used developer tool (e.g. if it's a GitHub project, it should have at least 500 stars and 150 forks)

btop 100% fits this category, btop4win is literally [btop](https://github.com/aristocratos/btop), it is made by the same developer. When you combine the stats of the repos, it far exceeds that requirement.

>Issues like what? Why is this necessary?

Because `btop.sys` cannot be deleted after being used, it will error out. You can try it yourself without a `pre_uninstall` and run the app and try to uninstall it after.

>Additionally, the manifest should be named btop4win as the [project](https://github.com/aristocratos/btop4win) is separate from [btop](https://github.com/aristocratos/btop) and even the executable name is called btop4win.

Why should it? it's btop for Windows, it's literally the repo where the Windows versions of btop are hosted, there is no other btop on Windows.

brian6932:
Please don't waste anyone's time in this thread, you aren't contributing anything

sitiom:
No need to be negative, I'm trying to help.

> btop 100% fits this category, btop4win is literally [btop](https://github.com/aristocratos/btop), it is made by the same developer. When you combine the stats of the repos, it far exceeds that requirement.

Okay, that's an exception if you count that as well. But alternate versions should still be put in the Versions bucket.

> Because `btop.sys` cannot be deleted after being used, it will error out. You can try it yourself without a `pre_uninstall` and run the app and try to uninstall it after.

I cannot reproduce this for some reason.

> Why should it? it's btop for Windows, it's literally the repo where the Windows versions of btop are hosted, there is no other btop on Windows.

The codebase is separate from the btop project, otherwise the Windows releases would be hosted in the same repo.

bucket/btop-lhm.json
sitiom: @brian6932 Do not close this, this hasn't been resolved yet.

brian6932:
There are many programs that use separate repos for Windows installations. this is not a valid reasoning to not use the name of the actual program. It is btop, there is no reason to add 4win, scoop is a Windows package manager.

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
...
+    "persist": "btop.conf",
+    "bin": "btop.exe",
```
sitiom, line 29: Do not rename the original executable, instead create a shim named `btop`

```suggestion
    "pre_install": "if (!(Test-Path \"$persist_dir\\btop.conf\")) { New-Item \"$dir\\btop.conf\" | Out-Null }",
    "pre_uninstall": [
        "$sys = \"$dir\\btop.sys\"",
        "if (!(Test-Path $sys)) { return }",
        "$tempPath = '~\\appdata\\temp\\btop.sys'",
        "if (Test-Path $tempPath) { $i = 0; While (Test-Path $tempPath) { $i++; $tempPath = \"$($tempPath | Split-Path)\\btop$i.sys\" } }",
        "Move-Item $sys $tempPath"
    ],
    "persist": "btop.conf",
    "bin": [
        "btop4win.exe",
        [
            "btop4win.exe",
            "btop"
        ]
    ],
```

sitiom (changes requested):

brian6932:
Renaming the executable is preferable, to adding a bin shim, especially in a case where there is no other btop on Windows.
There is add more complexity to the launch of the bin.

sitiom:
Please comment in the suggestion thread and not in the main thread itself. Going back:

> Renaming the executable is preferable to adding a renamed bin shim, especially in a case where there is no other btop on Windows. There is add more complexity to the launch of the bin.

The author intends the name of the executable to be `btop4win.exe`. Even the [command-line options](https://github.com/aristocratos/btop4win#command-line-options) in the readme state `btop4win.exe`. People who install this manifest might be confused of the renaming.

brian6932:
image
Are you sure about that?
Did you even try this manifest before commenting in here at all?

sitiom:
> image
> Are you sure about that?

Okay, that's an inconsistency between the readme and the command line output, but then the binaries are released as `btop4win.exe` instead of `btop.exe`.

> Did you even try this manifest before commenting in here at all?

Of course, I have added btop4win to the Extras bucket (https://github.com/ScoopInstaller/Extras/pull/9200) before you made your PR. Have you looked at that before making this PR?

sitiom (changes requested):
`btop-lhm` should be in the [Versions](https://github.com/ScoopInstaller/Versions) bucket, as this is an alternate version.

rashil2000:
- Renaming the binary to btop.exe is fine, as shown in the commandline help text. The '4win' suffix is redundant.
- The btop.sys step is unnecessary. It should be removed.
- To reach a consensus, I'm okay with adding both manifests to Main bucket. There are some instances of this presently (imagemagick, for example).

brian6932:
@rashil2000 This is the problem with removing `pre_uninstall` on `btop-lhm`:
image

brian6932:
This does not occur with the `pre_uninstall` I made

rashil2000:
How is moving it to a different location technically different from deleting it?

brian6932:
@rashil2000 I am not sure, but when I run `Remove-Item` it does still errors, even with `-Force`

rashil2000:
How about running the uninstall process as admin? Remove-Item should work in that case.

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
+{
+    "version": "1.0.2",
+    "description": "btop++ for windows",
```
rollingmoai, line 3: Specify description
```suggestion
    "description": "A monitor of system resources, btop++ for Windows",
```

bucket/btop.json
```diff
@@ -0,0 +1,31 @@
+{
+    "version": "1.0.2",
+    "description": "btop++ for windows",
```
rollingmoai, line 3: Specify description
```suggestion
    "description": "A monitor of system resources, btop++ for Windows",
```

brian6932:
@rashil2000 It's all admin, `.sys` files are apparently drivers, and would thus be in your kernel access, which I don't believe a user can just delete

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
...
+        "Has full GPU monitoring, etc. using Libre Hardware Monitor, DLL's included in the package.",
+        "Requires admin rights to run."
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/aristocratos/btop4win/releases/download/v1.0.2/btop4win-LHM-x64.zip",
+            "hash": "898b0725918e58926851304195e795b881abd1c2353b0537e62ea1a273777a66",
+            "extract_dir": "btop4win"
+        }
+    },
```
rollingmoai, line 16: Nitpick: move this out to architecture in case of future 32-bit and ARM versions.

```suggestion
            "hash": "898b0725918e58926851304195e795b881abd1c2353b0537e62ea1a273777a66"
        }
    },
    "extract_dir": "btop4win",
```

bucket/btop.json
```diff
@@ -0,0 +1,31 @@
...
+        "Has no support for GPU monitoring, CPU temperature monitoring and does not have accurate CPU clock monitoring.",
+        "Does not require admin rights to run. (But it's still higly recommended to run as admin, otherwise some process information will be missing.)"
+    ],
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/aristocratos/btop4win/releases/download/v1.0.2/btop4win-x64.zip",
+            "hash": "df0f7ef62dabe1d6776cba76d03e9922b440f55d210707cc3691b8abea5b3220",
+            "extract_dir": "btop4win"
+        }
+    },
```
rollingmoai, line 16: Nitpick: move this out to architecture in case of future 32-bit and ARM versions.

```suggestion
            "hash": "df0f7ef62dabe1d6776cba76d03e9922b440f55d210707cc3691b8abea5b3220"
        }
    },
    "extract_dir": "btop4win",
```

rollingmoai (changes requested):

brian6932:
@rollingmoai If there are future 32bit versions, then `extract_dir` can be moved when they are added.
As far as descriptions go, it's not a bad call to add the description of the original repo as well 👍

brian6932:
@rashil2000 If you find any solution to this lmk, I could personally not find a way to delete it 🤷

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
+{
+    "version": "1.0.2",
+    "description": "A monitor of resources, btop++ for windows",
```
rollingmoai, line 3: Might be good to also note the differences between the two manifests in the description field (like most manifests here):

```suggestion
    "description": "A monitor of resources, btop++ for windows - version with full monitoring capabilities using LibreHardwareMonitor",
```

rollingmoai (changes requested):

rashil2000:
> @rashil2000 If you find any solution to this lmk, I could personally not find a way to delete it 🤷

Use pre_uninstall like this: https://github.com/ScoopInstaller/Extras/blob/b2d8efd20976ee3ddb688a67cdb05e528518f071/bucket/gtk-sharp.json#L12

rollingmoai:
> Use pre_uninstall like this: [ScoopInstaller/Extras@`b2d8efd`/bucket/gtk-sharp.json#L12](https://github.com/ScoopInstaller/Extras/blob/b2d8efd20976ee3ddb688a67cdb05e528518f071/bucket/gtk-sharp.json#L12) to Remove-Item as admin

Then assuming I don't have admin rights, I cannot uninstall the manifest after installing it.

bucket/btop-lhm.json
brian6932: This is already in the notes

rollingmoai: It would be helpful for people searching though https://scoop.sh/, see example:
image

rashil2000:
> > Use pre_uninstall like this: [ScoopInstaller/Extras@`b2d8efd`/bucket/gtk-sharp.json#L12](https://github.com/ScoopInstaller/Extras/blob/b2d8efd20976ee3ddb688a67cdb05e528518f071/bucket/gtk-sharp.json#L12) to Remove-Item as admin
> 
> Then assuming I don't have admin rights, I cannot uninstall the manifest after installing it.

In that case installing should also be done as admin

brian6932:
@rashil2000 Yea it's not working, this is what happens when I do that:
image

rollingmoai:
Update to 1.0.3: https://github.com/aristocratos/btop4win/releases/tag/v1.0.3

rashil2000:
> With this being said, I really think moving it is the best option

No, if the file is not deletable by any means, it should be reported in their repo. Unless they fix that, the LHM version cannot be added to the bucket.

brian6932:
I have found 1 way to remove it, however it is not by any means a normal way:
You have to use `rm btop.sys`  from `msys2`, as that does not respect the Windows kernel's restrictions

brian6932:
@rashil2000 To clarify, if another instance of the program is started, and a new `btop.sys` is created, the old one becomes removable, or if you restart your PC, the previous ones become removable. This is why I think temp is honestly not a bad solution.

rashil2000:
Scoop should cleanly install and uninstall files, and leave no residual files, in any location.

brian6932:
Yes I understand, but I'm not sure if there is any other way to handle this, maybe @aristocratos can help 🤷

rashil2000:
Ideally the program should delete any files that it creates when it exits. And it should certainly **not** leave any _undeletable_ files behind.

rollingmoai:
> @rashil2000 To clarify, if another instance of the program is started, and a new `btop.sys` is created, the old one becomes removable, or if you restart your PC, the previous ones become removable. This is why I think temp is honestly not a bad solution.

[`processhacker`](https://github.com/ScoopInstaller/Extras/blob/master/bucket/processhacker.json) and [`processhacker-nightly`](https://github.com/ScoopInstaller/Versions/blob/master/bucket/processhacker-nightly.json) also behaves like this (this time it is `kprocesshacker.sys`), I usually just restart before uninstalling it.

aristocratos:
@brian6932 
Hi :)
The .sys file is the WinRing kernel driver used by LibreHardwareMonitorLib. Not sure why it isn't immediately removable, but likely because of weird Windows kernel stuff... (Can look into it, but if it is fixable would likely have to be done in https://github.com/LibreHardwareMonitor/LibreHardwareMonitor)
But as you said, moving the file to Windows temp and letting Windows take care of removing it when possible is an alternative.

brian6932:
I ran in to an issue with this:
image

brian6932:
@rashil2000 Issue should now be resolved

sitiom (changes requested):
Also persist the `themes` folder for any custom themes put or made by the user (such as https://github.com/catppuccin/btop).

brian6932:
I think that's a bad idea, if themes receive an update
It's best if the user just uses any arbitrary directory to point to their custom theme of choice

aristocratos:
@brian6932 The windows version only uses the themes folder in same folder as the .exe
Just missed updating the documentation...

brian6932:
Ok, I'll persist it then

sitiom:
> @brian6932 The windows version only uses the themes folder in same folder as the .exe

Can you make it support themes in the `.config` folder? https://github.com/ScoopInstaller/Main/pull/3897#issuecomment-1242970822 is a valid point. Scoop will keep persisting the old version of themes if they get updated on a newer version.

aristocratos:
@sitiom Create an issue in https://github.com/aristocratos/btop4win and I can add support for a `user-themes` folder or something like that when I've got time.

bucket/btop-lhm.json
```diff
@@ -18,7 +18,7 @@
         "Get-ChildItem \"$dir\\btop4win.exe\" | Rename-Item -NewName { $_.name -Replace '4win', '' }",
         "if (!(Test-Path \"$persist_dir\\btop.conf\")) { New-Item \"$dir\\btop.conf\" | Out-Null }"
     ],
-    "pre_uninstall": "sc stop R0btop | Out-Null",
+    "pre_uninstall": "sc.exe stop R0btop | Out-Null",
```
sitiom, line 21: Add admin rights check to match suggestion in https://github.com/ScoopInstaller/Main/pull/3897#issuecomment-1242946583
```suggestion
    "pre_uninstall": [
        "if (!(is_admin)) { error \"$app requires admin rights to $cmd\"; break }",
        "sc.exe stop R0btop | Out-Null"
    ],
```

sitiom (changes requested):

bucket/btop-lhm.json
```diff
@@ -0,0 +1,38 @@
+{
+    "version": "1.0.3",
+    "description": "A monitor of resources, btop++ for windows",
+    "homepage": "https://github.com/aristocratos/btop4win",
+    "license": "Apache-2.0",
+    "notes": [
+        "Has full GPU monitoring, etc. using Libre Hardware Monitor, DLL's included in the package.",
+        "Requires admin rights to run."
+    ],
```
rashil2000, line 9:
```suggestion
    "notes": "This application requires admin rights to run.",
```

bucket/btop.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "1.0.3",
+    "description": "A monitor of resources, btop++ for windows",
```
rashil2000, line 3:
```suggestion
    "description": "Resource monitor that shows usage and stats for processor, memory, disks, network, processes and services.",
```

bucket/btop.json
```diff
@@ -0,0 +1,34 @@
...
+    "homepage": "https://github.com/aristocratos/btop4win",
+    "license": "Apache-2.0",
+    "notes": [
+        "Has no support for GPU monitoring, CPU temperature monitoring and does not have accurate CPU clock monitoring.",
+        "Does not require admin rights to run. (But it's still higly recommended to run as admin, otherwise some process information will be missing.)"
+    ],
```
rashil2000, line 9:
```suggestion
```

Not required here

rashil2000 (changes requested):

bucket/btop-lhm.json
brian6932: I think keeping the notes from the repo is better, as they explain more what the differences are between the manifests

PR #3903, opened 2022-09-11
wgcf: Add version 2.2.15

### LINKED ISSUE #3902: [Request] wgcf
### Criteria

- [x] a non-GUI tool
- [x] a reasonably well-known and widely used developer tool (e.g. if it's a GitHub project, it should have at least 500 stars and 150 forks)

### Information

**Name:** wgcf

**Description:** unofficial CLI for Cloudflare Warp (Wireguard VPN service)

**Homepage:** https://github.com/ViRb3/wgcf

**Download link(s):** https://github.com/ViRb3/wgcf/releases
**Some indication of popularity/repute:** >2.1k stars, 260 forks

ftk:
Closes #3902 

### DIFF
--- a/bucket/wgcf.json
+++ b/bucket/wgcf.json
@@ -1,0 +1,40 @@
+{
+    "version": "2.2.15",
+    "description": "Unofficial CLI for Cloudflare Warp (Wireguard VPN service)",
+    "homepage": "https://github.com/ViRb3/wgcf",
+    "license": "MIT",
+    "architecture": {
+        "32bit": {
+            "url": "https://github.com/ViRb3/wgcf/releases/download/v2.2.15/wgcf_2.2.15_windows_386.exe#/wgcf.exe",
+            "hash": "ad58ec6bbe700238359624e68c26f460befefe0b8a3fba1b37a9f01f0553046b"
+        },
+        "64bit": {
+            "url": "https://github.com/ViRb3/wgcf/releases/download/v2.2.15/wgcf_2.2.15_windows_amd64.exe#/wgcf.exe",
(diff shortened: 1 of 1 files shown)


rashil2000:
What's the default location for wgcf-account.toml file?

ftk:
It's read from working directory by default. Its location is printed before each operation and passing --config to shim (second --config option) will change it without error.

rashil2000:
But is it recommended by them to save it to a particular static location? 

Because if we do that users will not be able to pass the `--config` flag to the binary (as it's already used).

ftk:
No, the tool doesn't care where you save the config.

As I said, passing second --config will change the location of used config file. That's why I think adding persist might be worth it.

rashil2000:
Ah, I get it now.

bucket/wgcf.json
```diff
@@ -0,0 +1,40 @@
...
+            "hash": "2aba25924a6db2703a344aff1f5b5573af8cd19d82a4718117675a0bc59d285d"
+        }
+    },
+    "pre_install": "if (!(Test-Path \"$persist_dir\\wgcf-account.toml\")) { New-Item \"$dir\\wgcf-account.toml\" | Out-Null }",
+    "bin": [
+        [
+            "wgcf.exe",
+            "wgcf",
+            "--config",
+            "$persist_dir/wgcf-account.toml"
```
rashil2000, line 22:
```suggestion
            "\"$persist_dir\\wgcf-account.toml\""
```
Would be a good idea to put it in quotes in case the Scoop folder contains space

PR #3912, opened 2022-09-13
retdec: Add version 4.0

Dany-coder778:
Added retdec.json in the bucket

### DIFF
--- a/bucket/retdec.json
+++ b/bucket/retdec.json
@@ -4,6 +4,7 @@
     "homepage": "https://github.com/avast/retdec",
     "license": "MIT",
     "notes": [
+        "The python executables were replaced with *.cmd by the shim creation.",
         "You could install the following programs:",
         " - UPX (Optional: if you want to use UPX unpacker in the preprocessing stage);",
         " - Graphviz (Optional: if you want to generate call or control flow graphs)."
--- a/bucket/retdec.json
+++ b/bucket/retdec.json
@@ -2,10 +2,7 @@
     "version": "4.0",
     "description": "A retargetable machine-code decompiler based on LLVM.",
     "homepage": "https://github.com/avast/retdec",
-    "license": {
-        "identifier": "MIT",
-        "url": "https://github.com/avast/retdec/blob/master/LICENSE"
-    },
+    "license": "MIT",
     "notes": [
         "It's recommended to install any version of Python >= 3.4",
         " (for ex. you can type 'scoop install python').",
--- a/bucket/retdec.json
+++ b/bucket/retdec.json
@@ -29,7 +29,6 @@
         "bin\\retdec-bin2llvmir.exe",
         "bin\\retdec-bin2pat.exe",
         "bin\\retdec-config.exe",
-        "bin\\retdec-config.py",
         "bin\\retdec-decompiler.py",
         "bin\\retdec-fileinfo.exe",
         "bin\\retdec-fileinfo.py",
--- a/bucket/retdec.json
+++ b/bucket/retdec.json
@@ -4,7 +4,6 @@
     "homepage": "https://github.com/avast/retdec",
     "license": "MIT",
     "notes": [
-        "The python executables were replaced with *.cmd by the shim creation.",
         "You could install the following programs:",
         " - UPX (Optional: if you want to use UPX unpacker in the preprocessing stage);",
         " - Graphviz (Optional: if you want to generate call or control flow graphs)."
--- a/bucket/retdec.json
+++ b/bucket/retdec.json
@@ -4,13 +4,14 @@
     "homepage": "https://github.com/avast/retdec",
     "license": "MIT",
     "notes": [
-        "It's recommended to install any version of Python >= 3.4",
-        " (for ex. you can type 'scoop install python').",
         "You could install the following programs:",
         " - UPX (Optional: if you want to use UPX unpacker in the preprocessing stage);",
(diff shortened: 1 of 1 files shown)


bucket/retdec.json
```diff
@@ -0,0 +1,40 @@
...
+        "url": "https://github.com/avast/retdec/blob/master/LICENSE"
+    },
+    "notes": [
+        "It's recommended to install any version of Python >= 3.4",
+        " (for ex. you can type 'scoop install python').",
+        "You could install the following programs:",
+        " - UPX (Optional: if you want to use UPX unpacker in the preprocessing stage);",
+        " - Graphviz (Optional: if you want to generate call or control flow graphs)."
+    ],
+    "depends": "vcredist2022",
```
rashil2000, line 16:
```suggestion
    "notes": [
        "You could install the following programs:",
        " - UPX (Optional: if you want to use UPX unpacker in the preprocessing stage);",
        " - Graphviz (Optional: if you want to generate call or control flow graphs)."
    ],
    "suggest": {
        "Python 3.4+": "python",
        "Visual C++ Redistributable": "extras/vcredist2022"
    },
```

bucket/retdec.json
```diff
@@ -0,0 +1,40 @@
...
+            "url": "https://github.com/avast/retdec/releases/download/v4.0/retdec-v4.0-windows-64b.7z",
+            "hash": "5fc667da3af3530cf3d393f96627987c3e46da62f4b426e78cdcc25a79416a31"
+        },
+        "32bit": {
+            "url": "https://github.com/avast/retdec/releases/download/v4.0/retdec-v4.0-windows-32b.7z",
+            "hash": "c9b6fc91007c2158209ac730d981a99006a960a27e3b0f54e5618872f34e1afe"
+        }
+    },
+    "extract_dir": "retdec",
+    "env_add_path": "bin",
```
rashil2000, line 28: Please create an array of .exe files (i.e. `bin` field) instead of `env_add_path` here

rashil2000 (changes requested):

bucket/retdec.json
Dany-coder778: Ok, done

bucket/retdec.json
```diff
@@ -0,0 +1,60 @@
...
+        }
+    },
+    "extract_dir": "retdec",
+    "bin": [
+        "bin\\retdec-archive-decompiler.py",
+        "bin\\retdec-ar-extractor.exe",
+        "bin\\retdec-bin2llvmir.exe",
+        "bin\\retdec-bin2pat.exe",
+        "bin\\retdec-config.exe",
+        "bin\\retdec-config.py",
```
rashil2000, line 33:
```suggestion
```

exe for this already exists

bucket/retdec.json
```diff
@@ -0,0 +1,60 @@
...
+        "bin\\retdec-fileinfo.py",
+        "bin\\retdec-getsig.exe",
+        "bin\\retdec-idb2pat.exe",
+        "bin\\retdec-llvmir2hll.exe",
+        "bin\\retdec-macho-extractor.exe",
+        "bin\\retdec-pat2yara.exe",
+        "bin\\retdec-signature-from-library-creator.py",
+        "bin\\retdec-stacofin.exe",
+        "bin\\retdec-unpacker.exe",
+        "bin\\retdec-unpacker.py",
```
rashil2000, line 45:
```suggestion
```

exe for this already exists

Dany-coder778: The *.py files are wrapper for the *.exe

bucket/retdec.json
```diff
@@ -0,0 +1,60 @@
+{
+    "version": "4.0",
+    "description": "A retargetable machine-code decompiler based on LLVM.",
+    "homepage": "https://github.com/avast/retdec",
+    "license": "MIT",
+    "notes": [
+        "The python executables were replaced with *.cmd by the shim creation.",
```
rashil2000, line 7:
```suggestion
```

This note is not required (it's implied)

Dany-coder778: The user needs to know how to call the wrapper

bucket/retdec.json
rashil2000: Then the exe file should be removed from the array

Dany-coder778: But it could also be useful, I don't know why there are two choice

rashil2000: Only one of them can be used, because of the preference order in the PATHEXT variable - so there's no use of two shims. The other will never get invoked.

Dany-coder778: I think this applies to all these duplicates

bucket/retdec.json
Dany-coder778: I believe this is fine because the script seems to set environment variables for other app python scripts

bucket/retdec.json
rashil2000: Hmm, okay. What about the help texts of retdec-unpacker?

Dany-coder778: They are more similar, but it seems that the exe has more functions
[out-unpacker.py-help.txt](https://github.com/ScoopInstaller/Main/files/9567021/out-unpacker.py-help.txt)
[out-unpacker.exe-help.txt](https://github.com/ScoopInstaller/Main/files/9567022/out-unpacker.exe-help.txt)

rashil2000: Okay

PR #3917, opened 2022-09-13
qpdf: Update to version 11.0.0

Dany-coder778:
Modified homepage on github repository

### DIFF
--- a/bucket/qpdf.json
+++ b/bucket/qpdf.json
@@ -1,33 +1,32 @@
 {
-    "version": "10.6.3",
+    "version": "11.0.0",
     "description": "A command-line program that does structural, content-preserving transformations on PDF files.",
-    "homepage": "http://qpdf.sourceforge.net/",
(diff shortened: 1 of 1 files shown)


rashil2000:
The releases are also present on SourceForge, right? 

We avoid using GitHub unless it's the only option, because we have a limited quota on Excavator for GitHub URLs.

bucket/qpdf.json
```diff
@@ -1,19 +1,20 @@
 {
-    "version": "10.6.3",
+    "version": "11.0.0",
     "description": "A command-line program that does structural, content-preserving transformations on PDF files.",
-    "homepage": "http://qpdf.sourceforge.net/",
+    "homepage": "https://github.com/qpdf/qpdf",
```
rashil2000, line 4:
```suggestion
    "homepage": "https://qpdf.sourceforge.net/",
```

PR #3918, opened 2022-09-14
cmder: fix broken homepage url

SiraKen:
Fixed broken link


SiraKen:
My changes do not pass checks because `hashes.txt` is not found in the latest version (v1.3.19). There is a file named `hashes.1.txt` instead of `hashes.txt`. I think this problem will be fixed in the next release probably.

PR #3943, opened 2022-09-20
add jqp 0.1.0

Slach:
jqp - A TUI playground to experiment with jq


rashil2000:
This project is way too new to be added to the Main bucket.

PR #3976, opened 2022-09-26
tfsec: Add version 1.28.0

Dany-coder778:
Added tfsec.json in the bucket

### DIFF
--- a/bucket/tfsec.json
+++ b/bucket/tfsec.json
@@ -5,22 +5,25 @@
     "license": "MIT",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


rashil2000:
Did you change the line endings or something? The whole file seems to be edited.

Dany-coder778:
I've added one binary in "bin" key.
Maybe by mistake I also changed the hash link, but anyway it works the same.

Dany-coder778:
> I've added one binary in "bin" key.
Because now it is unpacking an archive that includes an executable that was previously excluded

PR #3982, opened 2022-09-27
Github CLI: Add version 2.16.1

marcolongol:


niheaven:
See https://github.com/ScoopInstaller/Main/blob/master/bucket/gh.json

marcolongol:
there is no way I would have seen it using `scoop search gh`
image
`scoop search github` also doen't return anything that looks like the cli package, perhaps we could rename `gh.json` to `github-cli.json`?
image

PR #4013, opened 2022-10-11
7zip: Add `7zFM.exe` and `7zG.exe` to bin entry

musjj:
- `7zFM.exe`: Useful if you want to open the GUI file manager from the command line.
- `7zG.exe`: This is similar to `7z.exe`, except that it displays the progress in a separate GUI window, instead of the console.

### DIFF
--- a/bucket/7zip.json
+++ b/bucket/7zip.json
@@ -26,7 +26,17 @@
         "    Set-Content \"$dir\\$_\" $content -Encoding Ascii",
         "}"
     ],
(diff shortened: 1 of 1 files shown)


rashil2000:
Do both of these binaries accept commandline arguments?

musjj:
`7zFM.exe` accepts a single file argument, while `7zG.exe` should accept the same arguments as `7z.exe`.

PR #4025, opened 2022-10-14
7zip: Add ARM64 arch

niheaven:
7-zip doesn't provides MSI installer for ARM64, so use 7zr as bridge.

### DIFF
--- a/bucket/7zip.json
+++ b/bucket/7zip.json
@@ -22,7 +22,7 @@
                 "$7zr = Join-Path $env:TMP '7zr.exe'",
                 "Invoke-WebRequest https://www.7-zip.org/a/7zr.exe -OutFile $7zr",
                 "Invoke-ExternalCommand $7zr @('x', \"$dir\\$fname\", \"-o$dir\", '-y') | Out-Null",
-                "Remove-Item \"$dir\\Uninstall.exe\", \"$dir\\*-arm64.exe\""
+                "Remove-Item \"$dir\\Uninstall.exe\", \"$dir\\*-arm64.exe\", $7zr"
             ]
         }
     },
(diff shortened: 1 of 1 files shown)


sitiom (changes requested):
Remove `Uninstall.exe` left from the ARM64 installer

sitiom:
> I'm not using arm64 now

You can force install arm64 by passing `-a arm64`

>  so only `Uninstall.exe` should be removed?

`Uninstall.exe` and the installer itself (`7z2201-arm64.exe`)

niheaven:
Aha, yes, only test the installer.

bucket/7zip.json
```diff
@@ -6,15 +6,26 @@
...
+            "extract_dir": "Files\\7-Zip"
+        },
+        "arm64": {
+            "url": "https://www.7-zip.org/a/7z2201-arm64.exe",
+            "hash": "700dea3e4012319a09ccadfce91cf090334cfe658d0bdc42204e77acbea1ef99",
+            "pre_install": [
+                "$7zr = Join-Path $env:TMP '7zr.exe'",
+                "Invoke-WebRequest https://www.7-zip.org/a/7zr.exe -OutFile $7zr",
+                "Invoke-ExternalCommand $7zr @('x', \"$dir\\$fname\", \"-o$dir\", '-y') | Out-Null",
+                "Remove-Item \"$dir\\Uninstall.exe\", \"$dir\\*-arm64.exe\""
```
rashil2000, line 25:
```suggestion
                "Remove-Item \"$dir\\Uninstall.exe\", \"$dir\\*-arm64.exe\", $7zr"
```

niheaven: `$7zr` is in temp folder, we needn't remove it, right?

rashil2000: I'd prefer removing it - sometimes the auto cleanup of temp directory does not happen, and other times, people might reinstall 7zip so there'll be a conflicting file while downloading.

PR #4042, opened 2022-10-19
generate-domains-blocklist: Add version Jan.22.2022

Zliced13:


tech189:
Is there a use case for having a separate package just for this script? If it's included in the release of the program itself you could make a package for the program and add a shim for this `.py` script too. Can you use this script without the program itself?

Zliced13:
No it's not included in dnscrypt-proxy, it's completely separate. It's mainly for generating domain lists from host files.

rashil2000:
This should be included as part of https://github.com/ScoopInstaller/Main/blob/master/bucket/dnscrypt-proxy.json

PR #4067, opened 2022-10-27
Chez Scheme: Add version 9.5.8

ColeDeanShepherd:
Adding an auto-updating app manifest for Chez Scheme v9.5.8.


rashil2000:
Building packages is not allowed in Scoop's official buckets. Only pre-packaged applications. 

You can either run the installer (e.g. [here](https://github.com/maokwen/scoop-bucket/blob/master/bucket/chez-scheme.json)) or extract the exe using Dark (e.g. [here](https://github.com/dodorz/scoop/blob/master/ChezScheme.json)).

PR #4068, opened 2022-10-27
Idris 2: Add version 0.5.1

ColeDeanShepherd:
Adding an auto-updating app manifest for Chez Scheme v9.5.8.

### DIFF
--- a/bucket/idris2.json
+++ b/bucket/idris2.json
@@ -1,0 +1,48 @@
+{
+    "version": "0.5.1",
+    "description": "Pure functional programming language with dependent types",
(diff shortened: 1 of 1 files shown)


ColeDeanShepherd:
Note: This depends on my other PR #4067

rashil2000:
Building packages is not allowed in Scoop's official buckets. Only pre-packaged applications. 

It looks like prebuilt binaries are not available for Idris 2.

Idris 1 is already present in the Main bucket though (https://github.com/ScoopInstaller/Main/blob/master/bucket/idris.json).

PR #4086, opened 2022-11-03
resvg: Add version 0.25.0

brian6932:

### DIFF
--- a/bucket/resvg.json
+++ b/bucket/resvg.json
@@ -5,16 +5,28 @@
     "license": "MPL-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/RazrFalcon/resvg/releases/download/v0.25.0/resvg-win64.zip",
-            "hash": "01ae2caa627d2e376333113ab8cc79e0b89737c0c1f09ccb02e608c45f5083a6"
+            "url": [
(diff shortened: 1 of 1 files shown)


rashil2000:
What's the function of each of these utilities?
image

Shouldn't they be included too?

brian6932:
@rashil2000 a lot of those are either gui, or seperate programs that use the library, the cli util is the current one

brian6932:
usvg is a cli util, optionally you could make a package for that in main

rashil2000:
You can add multiple URLs in a manifest and make bins for them, see https://github.com/ScoopInstaller/Versions/blob/master/bucket/gcc-msys.json for example.

rashil2000:
I think we should add viewsvg (create a bin for it) and resvg-explorer-extension (no need for bin here) too

brian6932:
@rashil2000 Main is a scoop bucket for developer/CLI programs, viewsvg, and resvg-explorer-extension are gui and codecs for gui programs respectively, is this wise?
Maybe a separate manifest should be made for viewsvg and resvg-explorer-extension on `extras` or `versions`?

rashil2000:
I think it's fine: the primary app is resvg, which is CLI. Its helper apps can be GUI. (for example, main/7zip)

niheaven:
viewsvg is 32.8MB while resvg is 1.11MB and usvg is 779kB. Put it in extras IMO.

Issue #3832, opened 2022-08-16
Posgresql not working psql to log in


jhoriascos:
## Bug Report
**Package Name:** posgresql

Hello,

I have installed the latest version of postgresql (14.5) with scoop then I got the installation successful message.
 
Then I have configured with the notes or information provided by the package on how to build a new PostgreSQL service? and how to start a PostgreSQL service?

In the end when in the CMD, I go to run 'psql -U postgres'. I get the following error:

`psql: error: falló la conexión al servidor en «localhost» (::1), puerto 5432: FATAL:  role "postgres" does not exist`

### System details

**Windows version:** 11

**OS architecture:** 64bit

**PowerShell version:**  5.1.22000.832

![image](https://user-images.githubusercontent.com/66481268/184995219-856a4f85-84e3-400c-b31d-aa6014df0855.png)

![image](https://user-images.githubusercontent.com/66481268/184995081-b0861193-8abe-4510-ad8a-eaa8a350e2d5.png)

how can I enter the psql with the postgres user, I do not know if I need to configure something and also if possible can you help me to load the Admin 4 that comes with the same package that I have noticed that it does not load the data.

rashil2000:
You need to create the use using the `createuser` command before using psql. For more details see https://stackoverflow.com/questions/15301826/psql-fatal-role-postgres-does-not-exist

jhoriascos:
I used the command *`createuser -s postgres,`* but I didn't get any result. I keep getting the same error.

![image](https://user-images.githubusercontent.com/66481268/185177982-a2f02c3f-80be-4b89-8585-945bd6569b8b.png)

rashil2000:
Does this issue happen if you install Postgresql from their official installer?

jhoriascos:
No, I use the same version on my Macbook Pro with brew and it's fine, but I recently installed coop as application manager for windows and I thought I made a mistake in some step of the installer.  I found a solution that specifies delete the data folder, create a new data with permissions and after that the psql command is fine, but on inspection I notice that in the windows root C: it has installed a new psql and not in the postgresql application folder that scoop created and I also found that it deletes the admin 4 shorcut. 

https://aeadedoyin.medium.com/getting-started-with-postgresql-on-windows-201906131300-ee75f066df78

rashil2000:
[...]

PR #4095, opened 2022-11-06
Fix "FATAL: role "postgres" does not exist"

YauHsien:
A problem, caused by using the old line `Invoke-ExternalCommand -FilePath \"$dir\\bin\\initdb.exe\" -ArgumentList \"--username=postgres --encoding=UTF8 --locale=en --lc-collate=C\" | Out-Null`, occurs in Windows so that people could not login by the default superuser Id "postgres"

By migration to the new line `Invoke-ExternalCommand -FilePath \"$dir\\bin\\initdb.exe\" -ArgumentList @('--username=postgres', '--encoding=UTF8', '--locale=en', '--lc-collate=C') | Out-Null`, the issue is eliminated.

Relates to #3832

### DIFF
--- a/bucket/postgresql.json
+++ b/bucket/postgresql.json
@@ -22,7 +22,7 @@
     ],
     "post_install": [
         "if (!(Test-Path \"$dir\\data\\pg_hba.conf\")) {",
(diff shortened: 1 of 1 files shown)


rashil2000:
Wow, thanks for catching this!

PR #4100, opened 2022-11-07
nanazip: Add version 2.0.396

niheaven:
7-Zip fork with lots of features. https://github.com/M2Team/NanaZip

image

This app has a Windows 10/11 style context menu item, but I don't know how to register it.


rashil2000:
I think we should keep the original name of the binaries, otherwise it will confuse users with main/7zip

niheaven:
It is alternative of 7-zip, and when installed from Microsoft Store, the bin name is 7z/7zg by linking.

rashil2000:
Oh, I see.

PR #4135, opened 2022-11-18
sftpgo: persist configuration across updates

xmrv:
The following files have to be copied to the new version's root directory every time the app gets updated:

id_ecdsa
id_ecdsa.pub
id_ed25519
id_ed25519.pub
id_rsa
id_rsa.pub
sftpgo.db
sftpgo.json

By adding the files above to persist, the config files get copied automatically.


rashil2000:
Do these files exist by default?

xmrv:
Yes they do.

rashil2000:
I don't see them...
image

Only existing files can be persisted, or else you need to create them first in the `pre_install` step.

xmrv:
id_* files are created by the app at runtime if they don't already exist.

sftpgo.db and sftpgo.json are created at install.

Regardless, I understand the files must already exist in order to be persisted. I'll add a new commit to create those files beforehand.

xmrv:
Since Windows 10 builds earlier than 1809 do not have a built-in SSH client, the keys cannot be created in a way that doesn't require the installation of an SSH client at `pre_install`. So I'm only adding the config files to persist and none of the keys.

rashil2000:
I see. 

Could you also do one more thing? This package is available for 32bit and arm64 variants as well, it's just that the binary for them is located in separate folders. Can you add a pre_install step for these architectures to move the binary to the root folder? (and replace the one that's present in root folder)

xmrv:
Done.

PR #4177, opened 2022-11-28
(chore): Add new 'hash-error' issue template

niheaven:
Enable blank issues when the new template is not used by default in Scoop Core.

I'll commit to other official buckets when this is merged.

Please see https://github.com/niheaven/scoop-sysinternals/issues/new?template=hash-error.yml&title=process-explorer%4017.02%3a+hash+check+failed for the new template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Scoop Community Support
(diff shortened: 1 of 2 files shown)


rashil2000:
IMO, hash check is a pretty simple issue with a straightforward fix (which the bot handles very well), why do we need a template for it?

Note that we would then need a template for "decompress error" and "download failed" too, because currently they are also not working.

niheaven:
Users will always use blank issue if they could, then the bug reporting system makes no sense. They don't provide app name, PS version, config, and what the bug is.

PR #4203, opened 2022-12-07
Update imagemagick-lean.json

ddarfantasy:
Make autoupdate aware of tail version changes since [this change](https://github.com/ImageMagick/ImageMagick/discussions/5749).

### DIFF
--- a/bucket/imagemagick-lean.json
+++ b/bucket/imagemagick-lean.json
@@ -1,16 +1,16 @@
 {
-    "version": "7.1.0",
+    "version": "7.1.0-53",
     "description": "Create, edit, compose, and convert 200+ of bitmap images formats.",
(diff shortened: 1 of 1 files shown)


rashil2000:
So the checkver is still picking up `7.1.0` as the version...

ddarfantasy:
> So the checkver is still picking up `7.1.0` as the version...

I left it unchanged, it's the same as before and is consistent with the main imagemagick package. I have no idea why automated test give this result...

ddarfantasy:
Oh, I see, the old zip file is still available, so the checkver thinks the old zip is newer (since old zip name is larger)...
I've asked developers to delete the old zips...

rashil2000:
Cool, thanks for following up with the devs!

Issue #1606, opened 2020-11-23
gopass: missing shims for binaries other than gopass.exe


leogott:
The manifest for gopass should be updated to include the other binaries as well

PR #4212, opened 2022-12-11
nircmd: Add to path instead of shimming

sitiom:
`nircmd.exe` creates a brief ghost window on silent commands (such as `execmd`) due to the shim; adding it directly to path fixes the problem.

### DIFF
--- a/bucket/nircmd.json
+++ b/bucket/nircmd.json
@@ -13,10 +13,7 @@
             "hash": "5071b54669bb1e88422c6c340204b0b3a0ffd07e2ac1d747ccbd1447abc92948"
         }
     },
-    "bin": [
-        "nircmd.exe",
-        "nircmdc.exe"
-    ],
+    "env_add_path": ".",
     "checkver": "<td>NirCmd v([\\d.]+)",
     "autoupdate": {
         "architecture": {


rasa:
`nircmd.exe` is a Windows GUI application.
`nircmdc.exe` is a Windows CLI application.

rashil2000:
Instead of adding to path (leading to pollution), how about we copy the two bins to the shim folder (and remove them on uninstall)?

rasa:
> Instead of adding to path (leading to pollution), how about we copy the two bins to the shim folder (and remove them on uninstall)?

@rashil2000 That's an excellent idea, and I am surprised that isn't the default, or at least a global option to either shim or copy the app.

sitiom:
> Instead of adding to path (leading to pollution), how about we copy the two bins to the shim folder (and remove them on uninstall)?

Is it really worth it? I'd like to make the manifest as simple as possible, are there any other apps that do this?

> @rashil2000 That's an excellent idea, and I am surprised that isn't the default, or at least a global option to either shim or copy the app.

I agree that it should be a built-in option if it would be done.

sitiom:
> @sitiom I think it is important, as Windows has a pretty short limit on the length of the PATH variable. It used to be 2048

According to the [docs](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables?redirectedfrom=MSDN):
> The maximum size of a user-defined environment variable is **32,767 characters**. There is no technical limitation on the size of the environment block. However, there are practical limits depending on the mechanism used to access the block. For example, a batch file cannot set a variable that is longer than the maximum command line length.

rasa:
> The maximum size of a user-defined environment variable is **32,767 characters**.

cmd.exe's limit is 8,192, but I get your point.

rashil2000:
> @rashil2000 That's an excellent idea, and I am surprised that isn't the default, or at least a global option to either shim or copy the app.

It isn't the default because many times executables aren't standalone - they might require files such as `.dll`'s or folders such as `lib`/`libexec`/`usr`/`share` etc. to be present in the same location as themselves. So copying just the executable won't work.

For statically compiled apps like `nircmd`, we can do this :)

sitiom:
The problem with copying the binaries to the shim folder is that it will ignore running processes in that folder:
image

Which should normally behave this way:
image

rasa:
> The problem with copying the binaries to the shim folder is that it will ignore running processes in that folder:

That is true, but I think that's less of a concern than adding to the PATH. 

Perhaps we add a switch to the scoop installer that adds the app's `current` directory (or wherever the `bin` files live) to the PATH regarding of what the manifest says?

rashil2000:
I guess a better option would be to have an additional field like `copybin`, which will:
1. Copy the binaries to shim folder on installation and delete them on uninstallation
2. During uninstallation, check if those binaries are in use.

chawyehsu:
The issue is basically identical to https://github.com/ScoopInstaller/Scoop/issues/1606. Copying exes out of install dir and to shims dir is another polluting way on the other hand. I'd merge this for a quick fix and refer to #1606 for discussing the general fix.

PR #4224, opened 2022-12-13
rust, rustup: Use MSVC by default

### LINKED ISSUE #931: rust and rustup should be msvc by default
The [official Rust download page](https://www.rust-lang.org/tools/install) links to the MSVC version. This is what how most users will be installing it so scoop should match that.

There is also a good reason for that default. See the [official explanation](https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/msvc-toolchain-support.html) (emphasis mine):

> At the release of Rust 1.0, we only supported the GNU toolchain on Windows. With the release of Rust 1.2, we introduced initial support for the MSVC toolchain. After that, as support matured, we eventually made it the default choice for Windows users.
>
> The difference between the two matters for interacting with C. If you're using a library built with one toolchain or another, you need to match that with the appropriate Rust toolchain. **If you're not sure, go with MSVC; it's the default for good reason.**

Therefore `rust` should install the MSVC version and there should be a separate `rust-gnu`. Right now it is the other way around, presumably for historical reasons. To preserve backwards compatibility, maybe start by adding `rust-gnu` and slowly move towards deprecating `rust`. The same applies to `rustup`.

Thoughts?

comp500:
Changes rust and rustup packages to use MSVC toolchain by default, as this is generally the better supported (and default) choice.

In the rustup package, the MSVC version is always used (the GCC build of rustup provides no benefits, and the MSVC version can be used to install GCC toolchains). The --default-host flag has been removed, so new installs will use the rustup default (MSVC) but existing installs will have a configuration file indicating GCC is the default; a note was added to the rustup packages stating how to change this.

The rust-gnu package was added, to allow using the GCC toolchain without rustup, and a note was added to all rust packages indicating that rustup can be used to more easily manage multiple toolchains (and is the primary install method; see https://forge.rust-lang.org/infra/other-installation-methods.html).

Closes #931.

### DIFF
--- a/bucket/rust-gnu.json
+++ b/bucket/rust-gnu.json
@@ -1,0 +1,42 @@
+{
+    "version": "1.65.0",
+    "description": "A language empowering everyone to build reliable and efficient software. (GNU toolchain)",
+    "homepage": "https://www.rust-lang.org",
+    "license": "MIT|Apache-2.0",
+	"notes": [
+		"Use the rustup package instead for easier management of multiple toolchains, including beta/nightly releases."
+	],
+    "architecture": {
+        "64bit": {
+            "url": "https://static.rust-lang.org/dist/rust-1.65.0-x86_64-pc-windows-gnu.msi",
+            "hash": "598b61d3f432d2a6b54733019fd886f95bdc08d2bedf2fcbcf6b01d09eea208b"
+        },
+        "32bit": {
+            "url": "https://static.rust-lang.org/dist/rust-1.65.0-i686-pc-windows-gnu.msi",
+            "hash": "5e1a7d3f35ec33b4153ffdb3157e5ed45d9b6f3727b6e032bae5f794f2ad7344"
+        }
+    },
+    "extract_dir": "Rust",
+    "bin": [
+        "bin\\rustc.exe",
+        "bin\\rustdoc.exe",
+        "bin\\cargo.exe"
+    ],
+    "checkver": {
+        "url": "https://www.rust-lang.org/",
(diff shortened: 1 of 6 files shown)


rasa (approved):

rasa:
Whitespace issues:
https://github.com/ScoopInstaller/Main/actions/runs/3688451634/jobs/6245761114#step:4:29

rashil2000:
The `rust-msvc` manifest can be removed (as it is identical to `rust`).

For rustup, we can have two files: `rustup` and `rustup-gnu`. Both files download the same file, but the `-gnu` manifest used the additional flag specifying the triplet while installation.

niheaven:
Add `rustup-gnu` and keep the two `-msvc` ones for back-compatibility.

rashil2000:
We can make a breaking change now as the linked issue has been present for more than two years, so no need for back-compat IMO 🙂

niheaven:
I mean, someone is using rustup-msvc or rust-msvc (like me)...

rashil2000:
Having all 3 variants will cause confusion honestly.

rustup-msvc and rust-msvc should be easily replaceable in your environment as they are independent.

niheaven:
Keep them until we have some solutions, IMO. Not everyone checks the update log or uses `scoop status` and `scoop info` to get changes.

comp500:
I've added the `rustup-gnu` package defaulting to the GCC toolchain; because the default host is set in the install command it will be reset on every install/update of the package, so I added a note to make this clear to users.

PR #4284, opened 2022-12-28
tldr: Download the help pages in post_install

grafst:
if we don't do that on install it does it on first usage, which can result in a delay. I think an initial download of all the pages at install time is good UX.  I'm open to critique or improvements.
let me know what you think.

### DIFF
--- a/bucket/tldr.json
+++ b/bucket/tldr.json
@@ -10,6 +10,9 @@
         }
     },
     "bin": "tldr.exe",
+    "post_install": [
+	    "tldr --update"
+    ],
     "autoupdate": {
         "architecture": {
             "64bit": {


tech189:
Thank you for your contribution! I think this would be a good addition, but maybe we should print a message before the update command is run to explain to the user what is happening. And maybe silence the command's output too?

HUMORCE:
This should not be included as part of installation, IMO

Another well-known package as an example: `main/msys2`

HUMORCE:
make it as option:
```
if(test-path \"$persist_dir\\KEEP_UPDATE\") { tldr --update }
```

grafst:
Ok, thanks for the feedback.

of course I can add a status message and hide the output.
 
I respectfully disagree that it should not be done at install time. The
app is not usable at all without the actual tldr-pages.
What if  I install the app and the next time I open it I don't have an
Internet-Connection? 

Can you explain what the $persist_dir\\KEEP_UPDATE would do? I'm to dumb
to get it 🙈


On December 29, 2022, Markus Hilsenbeck ***@***.***>
wrote:
> make it as option:
>
> if(test-path \"$persist_dir\\KEEP_UPDATE\") { tldr --update }

—
Reply to this email directly, view it on GitHub
<https://github.com/ScoopInstaller/Main/pull/4284#issuecomment-
1367358473>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/ACAUEDZR6PQXPMOMZIRNWVLWPWMIPANCNFSM6AAAAAATK7JIRY>.
You are receiving this because you authored the thread.Message ID:
***@***.***>


--
Sent with HEY <https://hey.com/sent> — Email as it should be

HUMORCE:
> Can you explain what the $persist_dir\\KEEP_UPDATE would do?

check if file `KEEP_UPDATE` exists in `<SCOOP>\persist\tldr`.

This manifest does not persist any thing, so we need `ensure $persist_dir` first.

If you do this, add `notes` too.

rashil2000:
I feel users should do this themselves. See the Brewfile for example: https://github.com/isacikgoz/homebrew-taps/blob/master/tldr.rb

PR #4351, opened 2023-01-13
chart-testing: Add version 3.7.1

stevehipwell:
This PR adds the Helm [chart-testing](https://github.com/helm/chart-testing) CLI.

### DIFF
--- a/bucket/helm-chart-testing.json
+++ b/bucket/helm-chart-testing.json
@@ -1,0 +1,36 @@
+{
+    "version": "3.7.1",
+    "description": "Tool for testing Helm charts. It is meant to be used for linting and testing pull requests. It automatically detects charts changed against the target branch.",
+    "homepage": "https://github.com/helm/chart-testing/",
+    "license": "Apache-2.0",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/helm/chart-testing/releases/download/v3.7.1/chart-testing_3.7.1_windows_amd64.zip",
+            "hash": "218f6672e2145fc270e166c132c46db36005fb990d07df01635958d9daf99ab8"
+        },
+        "arm64": {
+            "url": "https://github.com/helm/chart-testing/releases/download/v3.7.1/chart-testing_3.7.1_windows_arm64.zip",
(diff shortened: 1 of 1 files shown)


rashil2000 (changes requested):
- Arm64 variant
- `chart-testing.json` -> `helm-chart-tester.json`

stevehipwell:
@rashil2000 I've added the `arm64` version and renamed the manifest to `helm-chart-testing` (it would be incorrect to rename it to `helm-chart-tester` as that's not it's name).

bucket/helm-chart-testing.json
```diff
@@ -0,0 +1,35 @@
...
+            "url": "https://github.com/helm/chart-testing/releases/download/v3.7.1/chart-testing_3.7.1_windows_amd64.zip",
+            "hash": "218f6672e2145fc270e166c132c46db36005fb990d07df01635958d9daf99ab8"
+        },
+        "arm64": {
+            "url": "https://github.com/helm/chart-testing/releases/download/v3.7.1/chart-testing_3.7.1_windows_arm64.zip",
+            "hash": "f26d5ed42e1586c6553abdab0f369981169b363906125078edc350292c6fee83"
+        }
+    },
+    "post_install": [
+        "New-Item -ItemType Directory -Path \"~/.ct\" -Force | Out-Null",
```
rashil2000, line 17: Is this post_install necessary? We don't want to make changes, that aren't strictly required for the functioning of the app, outside of the installation folder.

stevehipwell: Yes, it's 100% required or else `ct lint` and `ct lint-and-install` will not work.

rashil2000: According to the docs:

> Note that linting requires config file for [yamllint](https://github.com/adrienverge/yamllint) and [yamale](https://github.com/23andMe/Yamale). If not specified, these files are search in the current directory, the .ct directory in current directory, `$HOME/.ct`, and `/etc/ct`, in that order.

stevehipwell: I strongly disagree, these files are primarily tool scoped and the other installation methods make sure they exist. Without these files the tool doesn't work without significant user interaction.

PR #4381, opened 2023-01-25
winget-ps: Add version 1.4.10173

sitiom:
PowerShell Module that comes with the latest release of Winget.

### DIFF
--- a/bucket/winget-ps.json
+++ b/bucket/winget-ps.json
@@ -5,8 +5,7 @@
     "license": "MIT",
     "suggest": {
         "PowerShell Core": "pwsh",
-        "WinGet": "winget",
-        "VCRedist2022": "extras/vcredist2022"
+        "WinGet": "winget"
     },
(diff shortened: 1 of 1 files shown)


bucket/winget-ps.json
```diff
@@ -0,0 +1,25 @@
+{
+    "version": "1.4.10173",
+    "description": "A PowerShell Module for the Windows Package Manager Client.",
+    "homepage": "https://github.com/microsoft/winget-cli/tree/master/src/PowerShell/Microsoft.WinGet.Client",
+    "license": "MIT",
+    "suggest": {
+        "PowerShell Core": "pwsh",
+        "WinGet": "winget",
+        "VCRedist2022": "extras/vcredist2022"
```
rashil2000, line 9: Is this required? It's already present in the winget manifest.

sitiom: The PS  Module requires vcredist, but they would also get vcredist for Winget. So I guess it can be removed as a transitive dependency?

rashil2000: Yes

PR #4409, opened 2023-02-01
(chore): Explicit declaration of source(bucket name) of suggested manifests

HUMORCE:
Relates to https://github.com/ScoopInstaller/Scoop/issues/5131

### DIFF
--- a/bucket/aws-ecs.json
+++ b/bucket/aws-ecs.json
@@ -4,7 +4,7 @@
     "homepage": "https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_CLI.html",
     "license": "Apache-2.0",
     "suggest": {
-        "AWS CLI": "aws"
+        "AWS CLI": "main/aws"
     },
     "architecture": {
         "64bit": {
--- a/bucket/aws-sam-cli.json
+++ b/bucket/aws-sam-cli.json
@@ -4,7 +4,7 @@
     "homepage": "https://aws.amazon.com/serverless/sam/",
     "license": "Apache-2.0",
     "suggest": {
-        "AWS CLI": "aws"
+        "AWS CLI": "main/aws"
     },
     "architecture": {
         "64bit": {
--- a/bucket/bat.json
+++ b/bucket/bat.json
@@ -4,7 +4,7 @@
     "homepage": "https://github.com/sharkdp/bat",
     "license": "Apache-2.0",
     "suggest": {
-        "less": "less"
+        "less": "main/less"
     },
     "architecture": {
         "64bit": {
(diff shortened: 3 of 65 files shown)


rashil2000:
This seems to be a huge change: a simpler alternative would be to just prepend `main/` (if nothing else is there) in Scoop core itself.

HUMORCE:
> This seems to be a huge change: a simpler alternative would be to just prepend main/ (if nothing else is there) in Scoop core itself.

Not really, many manifests from `extras/versions` also do not clearly identify.

Implement it in Scoop Core may breaks **3rd-party buckets and specific scenarios**. this is also the reason why I have not changed `depends`.

Adding bucket name for suggested manifests does not affect the existing functionality and we can restore them easily.

rashil2000:
> Not really, many manifests from extras/versions also do not clearly identify.

Yes, this is precisely why I'm advocating against this. We follow a convention - a simple one - if the suggestion is from the default bucket of Scoop (i.e. Main), then no need to specify bucket, if it is from elsewhere (even official ones) we need to specify bucket.

> Implement it in Scoop Core may breaks 3rd-party buckets and specific scenarios.

That's on them - they did not follow the convention. It has been present since a long time.

----

Not to mention that I'm against commits that modify so many files at once - it pollutes the git history.

----

Also, how is it related to https://github.com/ScoopInstaller/Scoop/issues/5131? I mean how exactly does the `suggest` field let users customize bucket sorting order?

HUMORCE:
> a simple one - if the suggestion is from the default bucket of Scoop (i.e. Main), then no need to specify bucket, if it is from elsewhere (even official ones) we need to specify bucket.
> Also, how is it related to https://github.com/ScoopInstaller/Scoop/issues/5131? I mean how exactly does the suggest field let users customize bucket sorting order?

Scoop does not have bucket sorting settings, and the main bucket does not have the highest weight when manifest selecting.

On this premise, as a package manager Scoop must prompt|install exactly the packages that are actually suggested|depended when writing the manifest. IMO

rashil2000:
Hmm, @niheaven any thoughts on this?

HUMORCE:
https://github.com/ScoopInstaller/Main/commit/377d913de9bec3b7125cdd43279c00027268b0d3, https://github.com/ScoopInstaller/Main/commit/438628c11de950afbdffe9e54c74797ec3dd175f, https://github.com/ScoopInstaller/Main/commit/7c4d47ac374b3c3d4ba9d0bd7353c74bde145ab9, https://github.com/ScoopInstaller/Main/commit/dbab8f8853d691967833ce368a8888fe5e94f646, https://github.com/ScoopInstaller/Main/commit/204313896c0995c4ad319be6ee5fe66f76197921, https://github.com/ScoopInstaller/Main/commit/8b3ddd74f3d1d4e6bf12db2f929cc8723036b672

PR #4417, opened 2023-02-04
dfu-util: Add version 0.11

brian6932:

### DIFF
--- a/bucket/dfu-util.json
+++ b/bucket/dfu-util.json
@@ -1,0 +1,30 @@
+{
+    "version": "0.11",
+    "description": "DFU is intended to download and upload firmware to/from devices connected over USB",
+    "license": "GPL-2.0-only",
+    "homepage": "https://dfu-util.sourceforge.net/",
+    "url": "https://dfu-util.sourceforge.net/releases/dfu-util-0.11-binaries.tar.xz",
+    "hash": "6450de30a7dcd8d8c1273f43f0b153f054fd24d85f7f38296b1ad8edbd2ddb25",
+    "pre_install": [
+        "if ($architecture -eq 'arm64') { error \"$app does not support $architecture\"; break }",
+        "$path = \"$dir\\dfu-util-$version-binaries\"",
+        "Move-Item \"$path\\win$($architecture.Substring(0, 2))\\*\" \"$dir\"",
+        "Remove-Item -Recurse $path"
+    ],
+    "bin": [
+        "dfu-prefix.exe",
+        "dfu-suffix.exe",
+        "dfu-util-static.exe",
+        "dfu-util.exe",
+        "lsusb-static.exe",
+        "lsusb.exe"
+    ],
+    "checkver": {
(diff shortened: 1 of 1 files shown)


bucket/dfu-util.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "0.11",
+    "description": "DFU is intended to download and upload firmware to/from devices connected over USB",
+    "license": "GPL-2.0-only",
+    "homepage": "https://dfu-util.sourceforge.net/",
+    "url": "https://dfu-util.sourceforge.net/releases/dfu-util-0.11-binaries.tar.xz",
+    "hash": "6450de30a7dcd8d8c1273f43f0b153f054fd24d85f7f38296b1ad8edbd2ddb25",
+    "pre_install": [
+        "if ($architecture -eq 'arm64') { error \"$app does not support $architecture\"; break }",
```
rashil2000, line 9: Arm machines can run 64bit binaries, so we can fallback to 64bit arch on Arm.

bucket/dfu-util.json
```diff
@@ -0,0 +1,30 @@
...
+    "bin": [
+        "dfu-prefix.exe",
+        "dfu-suffix.exe",
+        "dfu-util-static.exe",
+        "dfu-util.exe",
+        "lsusb-static.exe",
+        "lsusb.exe"
+    ],
+    "checkver": {
+        "url": "https://gitlab.com/api/v4/projects/188452/repository/tags",
```
rashil2000, line 23: You can use https://dfu-util.sourceforge.net/releases/, I think it will be simpler.

brian6932: Yes although you'd load an entire webpage body, instead of a smaller json which also has all the previous versions, seems better to use this imo.
Less regex necessary, and if they decide to change the site layout this will still work.

brian6932: Yes although you'd load an entire webpage body, instead of a smaller json which also has all the previous versions, seems better to use this imo.
Less regex necessary, and if they decide to change the site layout this will still work.

bucket/dfu-util.json
```diff
@@ -0,0 +1,30 @@
+{
+    "version": "0.11",
+    "description": "DFU is intended to download and upload firmware to/from devices connected over USB",
+    "license": "GPL-2.0-only",
+    "homepage": "https://dfu-util.sourceforge.net/",
+    "url": "https://dfu-util.sourceforge.net/releases/dfu-util-0.11-binaries.tar.xz",
+    "hash": "6450de30a7dcd8d8c1273f43f0b153f054fd24d85f7f38296b1ad8edbd2ddb25",
+    "pre_install": [
+        "if ($architecture -eq 'arm64') { $architecture = '64bit' }",
```
rashil2000, line 9: $architecture is an internal variable, so don't modify it. Just read it/check its value.

rashil2000:
What's the difference between the normal and `-static` binaries? Shouldn't we only need one of them?

PR #4418, opened 2023-02-04
xxd: Add version 3503723768

brian6932:
Gives you the option to install the latest xxd version without installing vim, if this manifest is a better fit for the versions bucket, lmk, I wasn't too sure which would be better.

### DIFF
--- a/bucket/xxd.json
+++ b/bucket/xxd.json
@@ -1,0 +1,24 @@
+{
+    "version": "3503723768",
+    "description": "A hexdump utility by Juergen Weigert",
+    "homepage": "https://github.com/ckormanyos/xxd",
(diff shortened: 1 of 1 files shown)


rashil2000:
Sorry, this project does not meet the popularity criteria for _any_ official bucket yet. 

Besides, the Actions artifacts have a retention time period, so after a couple of months the download link might not work.

brian6932:
@rashil2000 I mean it's apart of vim, xxd is pretty well known, it doesn't have stars it's not on GitHub, I'm perfectly fine with PRing to versions if that's preferred.

rashil2000:
I feel there are a lot of independent hexdump utilities, which are more suitable for package managers (some are already present in Main bucket).

PR #4470, opened 2023-02-20
git-quick-stats: Add version 2.4.1

### LINKED ISSUE #4226: [Request]: GIT quick statistics
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Non-GUI tool
- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

GIT quick statistics

### Description

Simple way to access various statistics in git repository.

### Homepage

https://git-quick-stats.sh/

### Download Link(s)

https://github.com/arzzen/git-quick-stats

5.6k stars and more than 200 forks on GitHub

StarsbySea:
Closes #4226

Install and Uninstall
Run well

### DIFF
--- a/bucket/git-quick-stats.json
+++ b/bucket/git-quick-stats.json
@@ -5,14 +5,12 @@
     "license": "MIT",
     "notes": "Run 'git config --global alias.quick-stats \"! $dir\\git-quick-stats\"' to configure alias.",
     "suggest": "git",
-    "url": "https://github.com/arzzen/git-quick-stats/archive/refs/tags/2.4.1.zip#/dl.zip",
-    "hash": "e9c9d65d1695dad96edc4a95ff76c5468c6d714b90e578926bd764fee77899b0",
-    "extract_dir": "git-quick-stats-2.4.1",
+    "url": "https://github.com/arzzen/git-quick-stats/raw/2.4.1/git-quick-stats",
+    "hash": "31ba61a03662bd6dc2e4416a8079f66cf2ee6e52db16daa2ca0797dc9f6968df",
     "checkver": {
         "github": "https://github.com/arzzen/git-quick-stats"
     },
     "autoupdate": {
-        "url": "https://github.com/arzzen/git-quick-stats/archive/refs/tags/$version.zip#/dl.zip",
-        "extract_dir": "git-quick-stats-2.4.1"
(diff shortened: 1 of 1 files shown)


rashil2000:
- You can just directly download the bash script from https://raw.githubusercontent.com/arzzen/git-quick-stats/2.4.1/git-quick-stats instead of the whole archive. 
- It's not a good idea to modify user's configuration files automatically. Instead, you can ask them to run that same command in the `notes` property
- Use `suggest` instead of `depends` for Git

StarsbySea:
* As this is considered one of the official installation methods, https://gist.github.com/arzzen/35e09866dfdadf2108b2420045739245, I don't really intend to modify it just to save a few MB of traffic. What if the script adds a new dependency?
* Modified as per comments
* Modified as per comments

rashil2000:
> * As this is considered one of the official installation methods, https://gist.github.com/arzzen/35e09866dfdadf2108b2420045739245, I don't really intend to modify it just to save a few MB of traffic. What if the script adds a new dependency?

Highly unlikely. Tools written in Bash are always self-contained shell scripts. Besides we can always make a change if they do add a dependency.

bucket/git-quick-stats.json
```diff
@@ -0,0 +1,16 @@
+{
+    "version": "2.4.1",
+    "description": "Git quick statistics is a simple and efficient way to access various statistics in git repository.",
+    "homepage": "https://git-quick-stats.sh/",
+    "license": "MIT",
+    "notes": "Run 'git config --global alias.quick-stats \"! $dir\\git-quick-stats\"' to configure alias.",
+    "suggest": "git",
```
rashil2000, line 7:
```suggestion
    "suggest": {
        "Git": "git"
    },
```

rashil2000:
Hmm, looks good to me. 

One last thing - we can add a `"bin": "git-quick-stats",` property too, so that people can also use it without needing to change the config.

PR #4496, opened 2023-03-02
cpufetch: Update to version 1.03

kazzarin:
The latest release is a single binary that is is labeled as 64-bit, but is actually 32-bit.


rashil2000:
x86-64 variant should be inside the architecture 64bit field.

rashil2000:
Sorry, I see what you mean now.

PR #4500, opened 2023-03-03
r: Added post_install script so r-studio can recognize it.

Zliced13:
Closes https://github.com/ScoopInstaller/Extras/issues/10533

### DIFF
--- a/bucket/r.json
+++ b/bucket/r.json
@@ -60,7 +60,11 @@
         ]
     ],
     "persist": "site-library",
-    "pre_uninstall": "Remove-Item 'HKCU:\\Software\\R-core' -Recurse -Force",
+    "pre_uninstall": [
+        "if ($cmd -eq 'uninstall')",
+        "   Remove-Item 'HKCU:\\Software\\R-core' -ErrorAction 'SilentlyContinue' -Recurse -Force",
+        "}"
+    ],
     "checkver": {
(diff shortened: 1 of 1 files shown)


rashil2000:
Why did you move bin and shortcuts outside the architecture block?

Also, I do not advocate modifying registry in the manifest. It should be optional.

/cc @niheaven

Zliced13:
I moved the `bin` and `shortcuts` outside the `architecture` block because I have yet to see a 32bit version that comes with its own shortcuts and binaries. Also, I will make the registry changes optional.

rashil2000:
> I moved the `bin` and `shortcuts` outside the `architecture` block because I have yet to see a 32bit version that comes with its own shortcuts and binaries. 

The path for those bins and shortcuts contains the term 'x64', so it should be inside the architecture block.

> Also, I will make the registry changes optional.

Okay.

niheaven:
I don't like to change R's manifest, and have fixed the issue by https://github.com/ScoopInstaller/Extras/commit/fbcf30f6a858d2ae5a7483e094c0e5841c2c32fd (maybe, please test).

Local test passed.

Zliced13:
I'll test it later on.

Zliced13:
@niheaven it works, I closed both PRs.

PR #4567, opened 2023-03-16
go: Add `GOROOT` env variable

xmtq:
goland cannot detect GOROOT's path without this variable.


rashil2000:
Go's official installer does not set this environment variable, so Scoop won't either. It has been deprecated. 

GoLand's settings can be configured with a path for Go's SDK, see here: https://www.jetbrains.com/help/go/configuring-goroot-and-gopath.html

PR #4589, opened 2023-03-23
qemu: Add all .exes to `bin`

ous50:
qemu add bin, for easier execution in pwsh


bucket/qemu.json
```diff
@@ -16,6 +16,74 @@
     "pre_install": "Expand-7zipArchive \"$dir\\dl.7z_\" \"$dir\" -Removal -Switches '-xr!*.exe.nsis -x!$PLUGINSDIR'",
     "env_add_path": ".",
```
rasa, line 17: If we're adding all the .exes, shouldn't we remove the `"env_add_path": ".",` now?

rashil2000: You should try restarting the terminal for running the commands.

ous50: Oh yes, and rebooting does makes it work. thx :)

PR #4595, opened 2023-03-24
redis: Update to version 7.0.9

### LINKED ISSUE #4092: The latest version of redis is 7.0, why haven't you updated it?
scoop search redis
5.0.14 -> 7.0.5
https://github.com/redis/redis/releases

kazzarin:
The current source of unofficial binaries has not had a release in over a year, so I have changed the source to one that has.

https://github.com/zkteco-home/redis-windows - binaries are uploaded to git instead of releases; build process not open
https://github.com/redis-windows/redis-windows - automated build process; less popular

Used the first source for now since it appears to be more popular.

Closes #4092

### DIFF
--- a/bucket/redis.json
+++ b/bucket/redis.json
@@ -11,18 +11,17 @@
     },
     "bin": [
         "redis-benchmark.exe",
-        "redis-check-aof.exe",
-        "redis-check-rdb.exe",
         "redis-cli.exe",
         "redis-server.exe"
     ],
     "checkver": {
-        "github": "https://github.com/zkteco-home/redis-windows"
+        "url": "https://raw.githubusercontent.com/redis-windows/redis-windows/main/.github/redis_latest_version.txt",
+        "regex": "([\\d.]+)"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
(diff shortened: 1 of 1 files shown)


rashil2000:
I'm not sure if the author of the repo will allow this, given that he wants all users to star the repo before using the builds. 

> Warning:
> If you don't like(click star) and don't use it, otherwise it will be regarded as illegal use.

kazzarin:
I don't see how that's enforceable though. If that's problematic I can switch this to the other source instead.

(The second source also has a slightly newer version as well)

niheaven:
Change to second source IMO.

rashil2000:
What's the difference between current  and `with-service` variant? I think we should use that one.

kazzarin:
Service variant includes an additional program for registering as a windows service. On the downside, it bloats the overall filesize likely because it's bundling the .NET runtime.

rashil2000:
I'm not a user of Redis, but I feel installing databases as a service is a common use case for developers - what do you think?

rashil2000:
Feature flags already has a tracking issue here: https://github.com/ScoopInstaller/Scoop/issues/3736

niheaven:
Submit a separate PR in `versions` bucket, `with-service`, and let users to choose.

PR #4598, opened 2023-03-24
gsudo: Fix for artifact renamed on v2.0.6, added arm64.

gerardog:
gsudo v2.0.6 has arm64 binaries and the artifact was renamed, so this fix is needed.

### DIFF
--- a/bucket/gsudo.json
+++ b/bucket/gsudo.json
@@ -26,7 +26,6 @@
         }
     },
     "bin": [
-        "gsudo.exe",
         [
             "gsudo.exe",
             "sudo"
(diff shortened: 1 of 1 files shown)


niheaven:
@gerardog I've change the manifest:

- Use MSI installer to reduce download size
- Only shim exe
- Add autoupdate.hash
- Use psmodule

Could you test it? thx

gerardog:
I would go all in for this change. It looks good, but my only concern is about using shims instead of adding gsudo to PATH.

There are some additional problems using the shim, like:
- Windows DOSKEY issue where arrows wont show history (sometimes), specially nesting shells. cmd -> pwsh -> gsudo
- Some scenarios where the credentials cache break. (ok this one needs deep testing, and maybe could be fixed, but not 100% sure)

I think this was discussed several times and even someone said scoop could use symlinks instead of shims...

Shall we leave that particular change (shimming the exe, not adding gsudo to the PATH) for another PR? otherwise I'm not sure when I will be able to test it soon. Also if any issue is found would have to be fixed in other release anyway.

rashil2000 (approved):

Issue #4064, opened 2022-10-26
poetry: Can't shim 'venv\Scripts\poetry.exe': File doesn't exist.


ptruser:
## Bug Report

**Package Name:** poetry

### Current Behaviour

`ERROR 'poetry' isn't installed correctly.`

### Expected Behaviour

Successfully installation of `poetry`.

### Additional context/output

```
Installing 'poetry' (1.2.2) [64bit] from main bucket
Loading install-poetry.py from cache
Checking hash of install-poetry.py ... ok.
Running installer script...
Python Linking ~\scoop\apps\poetry\current => ~\scoop\apps\poetry\1.2.2
Creating shim for 'poetry'.
Can't shim 'venv\Scripts\poetry.exe': File doesn't exist.
```

### Possible Solution

The first find your python in PATH:

```bash
$ where python3

%userprofile%\AppData\Local\Microsoft\WindowsApps\python3.exe
%userprofile%\scoop\shims\python3.exe
```

Try to run `python3` from cli. Python3 from `WindowsApps` runnig Microsoft Store and propose to install python.

My solution is just remove `python3` and `python` from `WindowsApps` and use python installed via scoop.

### System details

**Windows version:** 10
**OS architecture:** 64bit

PR #4614, opened 2023-03-29
fix(poetry): fix poetry installer

CALMorACT:
Relates to #4064

1. add packaging management for description
2. fix install with Invoke-Expression
3. fix path with official advices
image

### DIFF
--- a/bucket/poetry.json
+++ b/bucket/poetry.json
@@ -12,7 +12,7 @@
             "Invoke-Expression \"$(scoop which python) `\"$dir\\$fname`\" --version $version\""
         ]
     },
-    "env_add_path":["bin"],
+    "bin": "bin\\poetry.exe",
     "checkver": {
         "github": "https://github.com/python-poetry/poetry/"
     },
--- a/bucket/poetry.json
+++ b/bucket/poetry.json
@@ -1,6 +1,6 @@
 {
     "version": "1.4.1",
-    "description": "Dependency Management for Python",
+    "description": "Python packaging and dependency management",
     "homepage": "https://python-poetry.org/",
     "license": "MIT",
     "depends": "python",
@@ -9,10 +9,10 @@
     "installer": {
         "script": [
             "$env:POETRY_HOME=\"$dir\"",
-            "python3 \"$dir\\$fname\" --version $version"
+            "Invoke-Expression \"$(scoop which python) `\"$dir\\$fname`\" --version $version\""
         ]
     },
-    "bin": "venv\\Scripts\\poetry.exe",
+    "env_add_path":["bin"],
     "checkver": {
         "github": "https://github.com/python-poetry/poetry/"
     },


rashil2000:
Does poetry install any additional binaries in the `bin` folder (i.e. when installing packages)?

CALMorACT:
> Does poetry install any additional binaries in the `bin` folder (i.e. when installing packages)?

Only poetry.exe, but the official advice is add it to `PATH` environment variable, I'm not sure it will be adding other programs in the future.

rashil2000:
We usually avoid adding to PATH as it leads to PATH pollution. Since there's only one binary there, we should stick to the `bin` field for now. 

Which problems does the Invoke-Expression part fix?

CALMorACT:
> We usually avoid adding to PATH as it leads to PATH pollution. Since there's only one binary there, we should stick to the `bin` field for now.
> 
> Which problems does the Invoke-Expression part fix?

1. OK, Using the `bin` field can remain unchanged until the poetry changes it.
2. With the issue  #4064, if we use Invoke-Expression, we can fix this issue. `python3` can' t support unless changing the MS store' s PATH for scoop' s python package.

bucket/poetry.json
```diff
@@ -9,10 +9,10 @@
     "installer": {
         "script": [
             "$env:POETRY_HOME=\"$dir\"",
-            "python3 \"$dir\\$fname\" --version $version"
+            "Invoke-Expression \"$(scoop which python) `\"$dir\\$fname`\" --version $version\""
         ]
     },
-    "bin": "venv\\Scripts\\poetry.exe",
+    "bin": "bin\\poetry.exe",
```
rashil2000, line 15:
```suggestion
    "bin": "venv\\Scripts\\poetry.exe",
```
The `bin\\poetry.exe` file is a symlink. It's better to directly shim `venv\\Scripts\\poetry.exe`

bucket/poetry.json
```diff
@@ -9,10 +9,10 @@
     "installer": {
         "script": [
             "$env:POETRY_HOME=\"$dir\"",
-            "python3 \"$dir\\$fname\" --version $version"
+            "Invoke-Expression \"$(scoop which python) `\"$dir\\$fname`\" --version $version\""
```
rashil2000, line 12: Unless you make the change mentioned in https://github.com/ScoopInstaller/Main/issues/4064#issuecomment-1292765454, using `scoop which python` is not going to make a difference - it will also point to the MStore one. 

So we don't need to use `scoop which python` here.

bucket/poetry.json
CALMorACT: the `bin\\poetry.exe` ' s size is 104 KB, maybe there are some other logics, and I think maybe bin is more readable than `venv\\scripts`.
Also, as the poetry installation says, the `bin` should be added to the `PATH` variable, If we use the `bin` field instead, I think it should be set to `bin\\poetry.exe`

rashil2000: Okay

bucket/poetry.json
CALMorACT: In `poetry.json`, we use the `depends` to specify the scoop to install the python of scoop.

However, due to the Microsoft Store's `Path` policy, our `Python3` commands will be overwritten.
So using `scoop which python` will effectively find the address of the python installed by scoop rather than jumping to the Microsoft Store

rashil2000: `scoop which python` picks up the first python.exe it finds in PATH - it does not care whether it was installed via Scoop or via MStore. So if the Scoop shim Path is preceded by MStore shim Path, it will return the MStore shim. 

This is why you need to read what the linked comment says, again.

CALMorACT: But in my computer, if I use `scoop which python`, it will return the true path for my python installed via scoop, but not return the MS store Path, so it can correctly identify the path of Python to install Poetry normally.
image
image

So, the `scoop which python` can help us find the python installed by scoop without change the `Path` variable orders.
Also, with the `depends: python`, it can solve the python path rightly to install the poetry.

rashil2000: What is the output of `$env:Path.Split(';')` on your system?

CALMorACT: I'm sure that if I go back to the original `python3` approach, I'll be redirected to MS store.
image

CALMorACT: I find that, there are two similar `Path` for MS store.
image

I'm not sure if this is a unique case for me, but it' s really weird.🤣🤣

PR #4651, opened 2023-04-11
cppcryptfsctl: Add version 1.4.4.1

bharatvaj:

Adding this to Main since it is recommended by gocryptfs

### DIFF
--- a/bucket/cppcryptfsctl.json
+++ b/bucket/cppcryptfsctl.json
@@ -1,6 +1,7 @@
 {
     "version": "1.4.4.1",
     "license": "MIT",
+	"description": "cppcryptfs is an implementation of the gocryptfs encrypted overlay filesystem in C++ for Windows.",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


HUMORCE:
We should download both `cppcryptfs` and `cppcryptfsctl`.

rashil2000:
> We should download both `cppcryptfs` and `cppcryptfsctl`.

Yes, I think so too

bharatvaj:
> > We should download both `cppcryptfs` and `cppcryptfsctl`.
> 
> Yes, I think so too. And you can rename `cppcryptfsctl32.exe` in the URL field itself, using `#/cppcryptfsctl.exe`

Hey thanks, will make the changes

bharatvaj:
Apologies, should've done this sooner. I did the mentioned changes, hope everything is alright.

rashil2000:
- You forgot to modify the autoupdate field.
- The bins are common for both 32 and 64bit, so it can be moved out of the architecture field.

PR #4705, opened 2023-04-21
jbang: Add version 0.106.3

goostleek:


goostleek:
\verify

HUMORCE (approved):

HUMORCE (changes requested):
`env_add_path` instead of `bin` in this case.

PR #4728, opened 2023-04-25
Fix upstream URL for mdcat

swsnr:
I renamed my Github account a while ago, this change updates the URLs accordingly.

Also note that the `autoupdate.hash.url` field is not correct anymore; I no longer have SHA512 hashes in the release pipeline but only b2 under `B2SUMS.txt`.  Haven't updated that field though, because I don't understand what it's for, and whether b2 hashes are okay there.

### DIFF
--- a/bucket/mdcat.json
+++ b/bucket/mdcat.json
@@ -1,23 +1,23 @@
 {
     "version": "2.0.3",
     "description": "cat for markdown",
-    "homepage": "https://github.com/lunaryorn/mdcat",
+    "homepage": "https://github.com/swsnr/mdcat",
     "license": "MPL-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/lunaryorn/mdcat/releases/download/mdcat-2.0.3/mdcat-2.0.3-x86_64-pc-windows-msvc.zip",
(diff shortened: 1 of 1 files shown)


bucket/mdcat.json
```diff
@@ -1,23 +1,23 @@
...
         "regex": "mdcat-([\\d.]+)"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/lunaryorn/mdcat/releases/download/mdcat-$version/mdcat-$version-x86_64-pc-windows-msvc.zip"
+                "url": "https://github.com/swsnr/mdcat/releases/download/mdcat-$version/mdcat-$version-x86_64-pc-windows-msvc.zip"
             }
         },
         "hash": {
```
aliesbelik, line 23: I wonder it makes sense to remove `hash` section from `autoupdate` completely since BLAKE2 option is [not supported](https://scoop-docs.vercel.app/docs/concepts/App-Manifests.html#optional-properties) for Scoop manifests.

HUMORCE: I did not notice this, Scoop does not support checksum BLAKE2.

HUMORCE (approved):

PR #4772, opened 2023-05-04
llvm: Add arm64

kazzarin:
The project has binaries for ARM which are not built for every release (or built much later than the other windows binaries) seen here: https://github.com/llvm/llvm-project/releases
So I modified checkver in an attempt to accommodate the irregular release schedule.


niheaven (approved):

rashil2000:
I don't think we should keep different version for the Arm64 variant - it will be misleading because it won't be consistent with the  manifest version. We can just wait, till they build the Arm64 variant, and then update the manifest.

kazzarin:
>which are not built for every release

Well the builds caught up, so I guess it's not a concern for now.

>We can just wait, till they build the Arm64 variant, and then update the manifest.

Alright, I reverted the changes to checkver so it will work the same as before

niheaven:
They are building arm64's version later than 32bit or 64bit, and maybe this is a temporal solution now?

PR #4846, opened 2023-05-29
zvm: add version 0.2.2

deevus:
Adds zvm (Zig Version Manager): https://github.com/tristanisham/zvm

Notably I generated this manifest with ChatGPT 4. I find creating manifests a bit tedious, but reviewing and testing them is really simple. 

I gave ChatGPT the github repo address and the URLs of the release files and it happily generated a valid manifest. I asked it to generate the previous version with blank hashes so that I could use the autoupdate script. Worked perfectly 👌 All I had to update was the project description.

### DIFF
--- a/bucket/zvm.json
+++ b/bucket/zvm.json
@@ -1,0 +1,56 @@
+{
+    "homepage": "https://github.com/tristanisham/zvm",
+    "version": "0.2.2",
+    "description": "Zig Version Manager (zvm) is a tool for managing your Zig installs",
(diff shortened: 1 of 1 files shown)


deevus:
Upon some more testing I had to add a junction to ~/.zvm since the binary has that hard coded. The post install script will create a junction from ~/.zvm to $dir/data which is persisted.

niheaven:
Bravo!

niheaven (approved):

PR #4931, opened 2023-07-06
dvc: update to 3.4 and new download URL

achimgaedke:
The scoop autoupdate looks at an outdated link.

Some sleuthing brought me to https://github.com/iterative/dvc-exe/blob/main/upload.py and that confirms that the AWS S3 bucket https link is used in future as well, see the Windows specific download link on homepage


achimgaedke:
I tested this manifest by installing and running the autoupdate version check.

PR #4999, opened 2023-08-01
gibo: Update to version 3.0.5

simonwhitaker:
v3.0 was a re-write in Go, hence we now need `gibo.exe` instead of `gibo.bat` on Windows. The version tag now also has a `v` prefix (e.g. `v3.0.5` compared with `2.2.8`).

Disclaimer: I don't have access to a Windows PC, so haven't been able to test this.

Closes #XXXX

Relates to #XXXX


HUMORCE (approved):
Tested locally.

PR #5098, opened 2023-09-12
fix(sing-box): Fix the nested directory

copper424:
Scoop fail to properly install the sing-box because the zipped package contains a directory instead of required files. Use pre_install script to move the binary to the correct direcotry.


chawyehsu:
`extract_dir` is preferred

copper424:
> `extract_dir` is preferred
Add commit using extract_dir instead

chawyehsu (approved):

PR #5215, opened 2023-10-25
chore (GitHub API): Use repo ID, instead of repo name

brian6932:
Repo names are subject to change, IDs aren't

Relates to https://github.com/ScoopInstaller/Extras/pull/12093 https://github.com/ScoopInstaller/Versions/pull/1429

### DIFF
--- a/bucket/arc.json
+++ b/bucket/arc.json
@@ -15,7 +15,7 @@
     },
     "bin": "arc.exe",
     "checkver": {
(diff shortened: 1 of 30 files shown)


rashil2000:
What if a repo gets transferred and /or renamed to another user/org? Does the ID still remain the same?

brian6932:
Yes

Issue #5251, opened 2023-11-08
[Request]: expose `openssl.exe` that comes with `git`


bersbersbers:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Non-GUI tool
- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

git

### Description

git from scoop

### Homepage

https://bjansen.github.io/scoop-apps/main/git/

### Download Link(s)

https://bjansen.github.io/scoop-apps/main/git/

### Some Indication of Popularity/Repute

Users looking for `openssl.exe` currently have to
- install https://bjansen.github.io/scoop-apps/main/openssl/, which takes several minutes to download 150 MB from a slow mirror,
- install https://bjansen.github.io/scoop-apps/main/git/, add a bucket, and install https://bjansen.github.io/scoop-apps/versions/openssl-light/

However, git comes with `openssl.exe` already, which could be exposed via a `shim`. I thinking something along the lines of https://github.com/bersbersbers/Main/commit/238a78b80510dca2542cb5830727516836facb88.

HUMORCE:
git for windows comes many of tools, why is `openssl` special?

here's an alternative: `scoop help shim`

bersbersbers:
> git for windows comes many of tools, why is `openssl` special?

`openssl` is an absolute pain to install (not only via `scoop`), with a 160MB installer hosted on a *very* slow server. I had CI time out several times due to this. But I guess you are right, `scoop shim add openssl %SCOOP%\apps\git\current\usr\bin\openssl.exe` should do the trick.

By the way, is there a way to add a shim using a command path relative to an installed app, such as `scoop shim add openssl usr\bin\openssl.exe --relative-to-app git`?

HUMORCE:
How about `scoop prefix`? e.g. `"$(scoop prefix git)\usr\bin\openssl.exe"`

bersbersbers:
Awesome, thanks!

PR #5305, opened 2023-11-30
git: expose `openssl.exe`

bersbersbers:
`git` conveniently comes with `openssl.exe`, which can be a pain to install otherwise. However, I just noticed this by accident because it is not exposed as a shim. This PR fixes that.


HUMORCE:
Responded to in #5251

PR #5315, opened 2023-12-03
pipx: Add version 1.3.3

sitiom:
Install and run Python applications in isolated environments

### DIFF
--- a/bucket/pipx.json
+++ b/bucket/pipx.json
@@ -1,13 +1,13 @@
 {
-    "version": "1.3.1",
+    "version": "1.3.3",
     "description": "Install and run Python applications in isolated environments",
     "homepage": "https://pipx.pypa.io/",
     "license": "MIT",
     "suggest": {
         "Python": "python"
     },
-    "url": "https://github.com/pypa/pipx/releases/download/1.3.1/pipx.pyz",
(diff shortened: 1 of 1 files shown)


bucket/pipx.json
```diff
@@ -0,0 +1,19 @@
...
+    "description": "Install and run Python applications in isolated environments",
+    "homepage": "https://pipx.pypa.io/",
+    "license": "MIT",
+    "suggest": {
+        "Python": "python"
+    },
+    "url": "https://github.com/pypa/pipx/releases/download/1.3.3/pipx.pyz",
+    "hash": "348b59b82b1ba9032513d36c2ea03ad3a34e45805ef8de7638a047673ee8b4cb",
+    "bin": "pipx.bat",
+    "pre_install": "Set-Content -Value '@python.exe \"%~dp0pipx.pyz\" %*' -Path \"$dir\\pipx.bat\"",
```
tech189, line 12: Perhaps we should use just `python` rather than `python.exe` so it also works with `pyenv`?

sitiom: How does `pyenv` work? Does it create a non-exe shim?

tech189: That's right, it creates `python.bat` shims in `scoop\persist\pyenv\pyenv-win\shims`.

tech189 (approved):

Issue #5345, opened 2023-12-12
kustomize@5.3.0 hash check failed


mikaello:
### Prerequisites

- [X] I have used the predefined issue title. (e.g. "xxx@xxx&#58; hash check failed")
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name and Version

kustomize@5.3.0

### Description

The latest version of kustomize in Scoop is[ 5.1.1](https://github.com/ScoopInstaller/Main/blob/3fa6d8f43defa4fc0fc7e0c2a1ef1d6c8755242e/bucket/kustomize.json#L2), it has been released multiple releases after this. The excavator logs errors with 404:

```
kustomize: 5.3.0 (scoop version is 5.1.1) autoupdate available
Autoupdating kustomize
Searching hash for kustomize_v5.3.0_windows_amd64.tar.gz in https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.3.0/checksums.txt
Could not find hash in https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.3.0/checksums.txt
Downloading kustomize_v5.3.0_windows_amd64.tar.gz to compute hashes!

VERBOSE: GET with 0-byte payload
VERBOSE: received 6709-byte response of content type application/json
VERBOSE: Content encoding: utf-8

The remote server returned an error: (404) Not Found.
URL https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.3.0/kustomize_v5.3.0_windows_amd64.tar.gz is not valid
ERROR Could not update kustomize, hash for kustomize_v5.3.0_windows_amd64.tar.gz failed!
```

Thas is because it uses it uses the wrong file extension (should be ZIP) in the URL it is fetching:

* Wrong: https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.3.0/kustomize_v5.3.0_windows_amd64.tar.gz
* Correct: https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.3.0/kustomize_v5.3.0_windows_amd64.zip

PR #5346, opened 2023-12-12
fix: kustomize has changed to .zip from .tar.gz

mikaello:
Closes https://github.com/ScoopInstaller/Main/issues/5345

This bug has been present since [Kustomize 5.2.0](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.2.0), which was the first version with .zip (see also https://github.com/kubernetes-sigs/kustomize/issues/5116).

### DIFF
--- a/bucket/kustomize.json
+++ b/bucket/kustomize.json
@@ -1,23 +1,31 @@
 {
-    "version": "5.1.1",
+    "version": "5.3.0",
     "description": "Customize raw, template-free YAML files for multiple purposes, leaving the original YAML untouched and usable as is.",
     "homepage": "https://github.com/kubernetes-sigs/kustomize",
     "license": "Apache-2.0",
     "architecture": {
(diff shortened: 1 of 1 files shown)


HUMORCE (approved):

HUMORCE (changes requested):
Fix checkver and update to latest, consider add arm64

mikaello:
> Fix checkver and update to latest, consider add arm64

What is wrong with `checkver`? 

Wouldn't the excavator update to latest given the correct `url`?

HUMORCE:
> What is wrong with checkver?

I'm not sure, but I updated it to make sure it would at least pass checks. Thank you.

HUMORCE (approved):

mikaello:
Thanks for fixing and merging, @HUMORCE

PR #5365, opened 2023-12-19
ssh3: Add version 0.1.5-rc5

brian6932:
Currently on Windows, it's only a client

### DIFF
--- a/bucket/ssh3.json
+++ b/bucket/ssh3.json
@@ -1,0 +1,42 @@
+{
+    "version": "0.1.4-bsd",
+    "description": "SSH3: faster and rich secure shell using HTTP/3",
+    "homepage": "https://github.com/francoismichel/ssh3",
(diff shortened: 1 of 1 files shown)


HUMORCE (approved):

HUMORCE (changes requested):
Current checkver will matching RC version, and what is different between `0.1.4` and `0.1.4-bsd`, its log is not clear.

brian6932:
@HUMORCE `bsd` has server support for openbsd, I should update the branch anyway, it's a bit stale

HUMORCE:
Server support for BSD is not available before 0.1.4-bsd/0.1.4-rc2, so we should allow the manifest to use RC version for now. Once an stable version released, update the checkver as soon

HUMORCE (approved):

PR #5372, opened 2023-12-23
openssl: Fix environment variables (2)

brian6932:
Should now work no matter what version of PowerShell or Windows you're on, and not require admin


brian6932:
The other option to a8321dee461a60854b0228cf4768266638afd1f8 is to randomize a string for it's type name to prevent any possible collisions, imo this is a less clean solution, as it still keeps the scope leak.
```pwsh
[char[]](Get-Random -Minimum 97 -Maximum 122 -Count 10) -join ''
```

HUMORCE (approved):

PR #5389, opened 2024-01-02
k9s: Persist config and logs

### LINKED ISSUE #5388: [Bug]: k9s does not use persist folder for config and logs
### Prerequisites

- [X] I have written a descriptive issue title.
- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [X] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

k9s

k9s is using %USERDIR%\AppData\Local\k9s to store its files. A Scoop persist folder would be better.

### Steps to Reproduce

```console
* Install k9s via Scoop
* Check where config is stored
```

### Possible Solution

Set env variables `K9S_CONFIG_DIR` and `K9S_LOGS_DIR` in the manifest. This is a breaking change for Scoop k9s users, but as the whole configuration structure was changes with v0.30 anyway, now is the best time to change it.

From the [v0.30 release notes](https://github.com/derailed/k9s/releases/tag/v0.30.0):

> K9s configurations are still managed using either xdg XDG_CONFIG_HOME or you can set K9S_CONFIG_DIR to specify a your preferred k9s configs location.

Best way I found is setting them OS-globally.
```
    "env_set": {
        "K9S_CONFIG_DIR": "$persist_dir/config",
        "K9S_LOGS_DIR": "$persist_dir/logs"
    },
```
I will open a PR with this solution.

If there is an option to only set the variable when running the binary, I would be happy to learn about it.

### Scoop and Buckets Version

```console
❯ scoop --version
Current Scoop version:
v0.3.1 - Released at 2022-11-15

'extras' bucket:
31b0a1d43 (HEAD -> master, origin/master, origin/HEAD) uiforetw: Update to version 1.59

'main' bucket:
939283bcd (HEAD -> master) minio: Update to version 2024-01-01T16-36-33Z
```

### Scoop Config

```console
❯ scoop config

last_update         root_path scoop_branch scoop_repo
-----------         --------- ------------ ----------
02.01.2024 08:39:19 C:\scoop  master       https://github.com/ScoopInstaller/Scoop
```

### PowerShell Version

```console
❯ $PSVersionTable

PSVersion                      7.4.0
GitCommitId                    7.4.0
OS                             Microsoft Windows 10.0.22631
```

### Additional Softwares

_No response_

gitolicious:
Scoop persist folder is used to persist k9s config and logs.

Closes #5388

### DIFF
--- a/bucket/k9s.json
+++ b/bucket/k9s.json
@@ -14,6 +14,10 @@
         }
     },
     "bin": "k9s.exe",
+    "env_set": {
+        "K9S_CONFIG_DIR": "$persist_dir/config",
+        "K9S_LOGS_DIR": "$persist_dir/logs"
+    },
     "checkver": "github",
     "autoupdate": {
         "architecture": {


HUMORCE (approved):

rashil2000:
There's no persist field in the manifest...

rashil2000:
>  the `scoop\persist\k9s` folder is created upon first run. I just verified with a fresh install. 

This folder shouldn't have been created if there's no persist field in the manifest.

HUMORCE:
> This folder shouldn't have been created if there's no persist field in the manifest...

If the app supports recursive creation, then Scoop doesn't need to pre-creates persist_dir of app by persist function. And doing so also avoids the situation where some apps don't support Junction as possible.

rashil2000:
Umm, I didn't quite get you, what is recursive creation?

HUMORCE:
I'm trying explaining to you why we don't need to define persist in this manifest.

persist field is not a mandatory requirement for persistence, It just depends on how we use the persisted resource.

rashil2000:
Okay. The reason why I asked this was that I couldn't find where the persist folder gets created during the install process (https://github.com/ScoopInstaller/Scoop/blob/develop/lib/install.ps1) if there's no persist field in the manifest.

gitolicious:
That's exactly what @HUMORCE meant with _"recursive creation"_. The folder is not created by Scoop on installation, but by k9s on first run (also see my comment above). _Recursive create_ means that k9s doesn't need the `$persist_dir` to exist, but creates all required above `$persist_dir/config` recursively, namely `.../persist`, `.../persist/k9s` and `.../persist/k9s/config`.

rashil2000:
Ohh, I understand now, makes sense. 

Thanks for the explanation!

PR #5413, opened 2024-01-10
Windows application driver fix

nicerloop:
Two fixes to streamline the use of Windows Application Driver aka WinAppDriver aka WAD:
- move supporting dlls from System folder next to WinAppDriver.exe: otherwise the supporting dlls are not in path and WAD fails to run
- set APPIUM_WAD_PATH environment variable with custom WAD location: appium-windows-driver can use WAD as installed by the installer, but needs to be informed of the location of the scoop installed version

These two fixes allow to transparently replace WAD installed by the installer with WAD installed by scoop.

### DIFF
--- a/bucket/windows-application-driver.json
+++ b/bucket/windows-application-driver.json
@@ -6,6 +6,7 @@
     "url": "https://github.com/Microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi",
     "hash": "a76a8f4e44b29bad331acf6b6c248fcc65324f502f28826ad2acd5f3c80857fe",
     "bin": "Windows Application Driver\\WinAppDriver.exe",
+    "post_install": "mv $dir\\System\\* $dir\\Windows*",
     "checkver": "github",
     "autoupdate": {
         "url": "https://github.com/Microsoft/WinAppDriver/releases/download/v$version/WindowsApplicationDriver_$version.msi"
--- a/bucket/windows-application-driver.json
+++ b/bucket/windows-application-driver.json
@@ -5,11 +5,13 @@
     "license": "MIT",
     "url": "https://github.com/Microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi",
     "hash": "a76a8f4e44b29bad331acf6b6c248fcc65324f502f28826ad2acd5f3c80857fe",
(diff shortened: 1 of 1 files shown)


bucket/windows-application-driver.json
```diff
@@ -6,6 +6,10 @@
     "url": "https://github.com/Microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi",
     "hash": "a76a8f4e44b29bad331acf6b6c248fcc65324f502f28826ad2acd5f3c80857fe",
     "bin": "Windows Application Driver\\WinAppDriver.exe",
+    "post_install": "mv $dir\\System\\* $dir\\Windows*",
```
HUMORCE, line 9:
```suggestion
    "post_install": [
        "Move-Item \"$dir\\System\\*\", \"$dir\\Windows Application Driver\\*\" -Destination \"$dir\"",
        "Remove-Item \"$dir\\System\", \"$dir\\Windows Application Driver\""
    ],
```

Move all files to the app root directory.

bucket/windows-application-driver.json
```diff
@@ -6,6 +6,10 @@
...
+    "env_set": {
+        "APPIUM_WAD_PATH": "$dir\\Windows Application Driver\\WinAppDriver.exe"
+    },
```
HUMORCE, line 12:
```suggestion
```
I think users should set it up manually when using it with appium. At least, the envvar`APPIUM_WAD_PATH` have not been created when install via its MSI installer.

nicerloop: I will replace with a note.

bucket/windows-application-driver.json
nicerloop: This will conflict with the creation of a bin shim. I will remove the bin shim.

HUMORCE (approved):

nicerloop:
I cannot install windows-application-driver with both the moved files and the bin shim:

Creating shim for 'WinAppDriver'.
Can't shim 'WinAppDriver.exe': File doesn't exist.

That is why I proposed to remove the bin shim, which is not necessary as the actual location is to be expose as an environment variable to Appium.

HUMORCE:
> Creating shim for 'WinAppDriver'.
Can't shim 'WinAppDriver.exe': File doesn't exist.

Fixed: https://github.com/ScoopInstaller/Main/commit/04607fabbecad5664e78e65bef455cb238a7b275

PR #5463, opened 2024-01-23
xtensa-esp32-elf: Rename into xtensa-esp-elf

pembem22:
- xtensa-esp32/esp32s2/esp32s3-elf were unified into xtensa-esp-elf, see the [release notes](https://github.com/espressif/crosstool-NG/releases/tag/esp-13.2.0_20230928).

- Update manifest accordingly.


tech189:
We should try to avoid renaming manifests as users might get stuck on the old one and not realise that their app is no longer being updated (there is no current way to migrate users to the updated name). I feel like this may be too minor a name change to warrant a rename. Especially as all the binaries are still `xtensa-esp32*-elf` anyway. What do you think @HUMORCE?

rasa:
> We should try to avoid renaming manifests as users might get stuck on the old one and not realise that their app is no longer being updated (there is no current way to migrate users to the updated name). I feel like this may be too minor a name change to warrant a rename. Especially as all the binaries are still `xtensa-esp32*-elf`

@tech189 @pembem22 I would tend to agree. If we need to rename, then add a new manifest, but add a note/suggests to the old one that a rename is in the works, to give the user some sort of warning, and timeline for the rename.

Or even better, a note in the original file that says what a better name for the manifest is. And then enhance our search tools to search in the notes field for this info? Or better yet, add an `aliases` field.

rasa (commented):
See comment in thread.

HUMORCE:
Scoop Core does not provide features like aliases, indicating alternatives, or disabling manifests, currently. I would prefer add a deadline and announcement to `nodes` of the original manifest and keep its functional(autoupdate, etc.), then create new one.

https://github.com/ScoopInstaller/Extras/pull/11116 https://github.com/ScoopInstaller/Extras/pull/11171 And a manifest that is plague by name changes: https://github.com/ScoopInstaller/Extras/pulls?q=is%3Apr+foxit

PR #5515, opened 2024-02-15
sbt@1.9.8: Fix checkver url

aliesbelik:
From excavator logs:

```
sbt: couldn't match '/sbt-([\d.]+)\.zip' in https://www.scala-sbt.org/download.html`
```

### DIFF
--- a/bucket/sbt.json
+++ b/bucket/sbt.json
@@ -8,7 +8,7 @@
     "extract_dir": "sbt",
     "bin": "bin\\sbt.bat",
     "checkver": {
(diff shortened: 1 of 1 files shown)


tech189:
Thanks, could you also persist the config files in the `conf` folder? I think they are loaded by `sbt.bat` but would be great if you could check!

aliesbelik:
Sorry, missed previous comment.
Made `conf` persist + set `SBT_HOME` env variable.

tech189 (approved):

PR #5541, opened 2024-02-26
zigup: Add version 2024_02_25

redactedscribe:
[zigup](https://github.com/marler8997/zigup) is a CLI tool for managing [Zig](https://ziglang.org/)'s compilers.

It does lack a little in stars and especially forks, but this manifest still seems best suited for Main over the Extras bucket.

### DIFF
--- a/bucket/zigup.json
+++ b/bucket/zigup.json
@@ -1,0 +1,18 @@
+{
+    "version": "2024_02_25",
+    "description": "Download and manage Zig compilers",
+    "homepage": "https://github.com/marler8997/zigup",
+    "license": {
+        "identifier": "MIT-0",
+        "url": "https://github.com/marler8997/zigup/blob/master/LICENSE"
+    },
+    "url": "https://github.com/marler8997/zigup/releases/download/v2024_02_25/zigup.windows-latest-x86_64.zip",
+    "hash": "9aa5c98b55203712af61242753871ad75cada479373713b4dc2804c21c4db561",
+    "bin": "zigup.exe",
+    "checkver": {
+        "github": "https://github.com/marler8997/zigup"
+    },
+    "autoupdate": {
+        "url": "https://github.com/marler8997/zigup/releases/download/v$version/zigup.windows-latest-x86_64.zip"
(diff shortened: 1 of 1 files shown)


redactedscribe:
065c925d seems like an odd fix to pass the checks, but it worked. Installing via the manifest locally worked prior to any fix.

bucket/zigup.json
```diff
@@ -0,0 +1,16 @@
...
+    "description": "Download and manage Zig compilers",
+    "homepage": "https://github.com/marler8997/zigup",
+    "license": {
+        "identifier": "MIT-0",
+        "url": "https://github.com/marler8997/zigup/blob/master/LICENSE"
+    },
+    "url": "https://github.com/marler8997/zigup/releases/download/v2024_02_25/zigup.windows-latest-x86_64.zip",
+    "hash": "9aa5c98b55203712af61242753871ad75cada479373713b4dc2804c21c4db561",
+    "bin": "zigup.exe",
+    "checkver": "github",
```
tech189, line 12:
```suggestion
    "checkver": {
        "url": "https://api.github.com/repos/marler8997/zigup/releases",
        "regex": "v([\\d_]+)"
    },
```
This should fix the checkver, it wasn't working because of underscores in the version number

tech189 (changes requested):

redactedscribe:
Thanks @tech189.

tech189:
Just realised that the downloaded versions of zig are not being persisted, I've made another commit to fix this - https://github.com/ScoopInstaller/Main/commit/0000de0f8e7c4c0afe62e999d7210c69116a8899. However, there is no way (--path-link doesn't work for me) to move the symlink `zig.exe` that is generated to a folder that Scoop can persist at the moment. See: https://github.com/marler8997/zigup/issues/114

tech189:
As of version `2024_03_13`, this is now fixed (solved by fixing https://github.com/marler8997/zigup/issues/114). I have pushed a commit that now aliases `zigup` with the `--path-link` argument to persist the symlink 🎉

PR #5552, opened 2024-02-27
wixtoolset: Update to v4

chawyehsu:
Don't know if it is proper to update to v4 directly but Wix 4 works totally different from its previous versions... This will be a breaking change I think. Wix 4 is required to unpack installers that are packed with Wix 4, which means `dark` will not work, taking `swift` as the example..

P.S. Wix 5 release is planning..

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/wixtoolset.json
+++ b/bucket/wixtoolset.json
@@ -1,36 +1,17 @@
 {
-    "version": "3.14",
+    "version": "4.0.4",
     "description": "Set of tools for windows installer creation.",
     "homepage": "https://wixtoolset.org/",
     "license": "MS-RL",
-    "url": "https://github.com/wixtoolset/wix3/releases/download/wix314rtm/wix314-binaries.zip",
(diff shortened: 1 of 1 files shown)


chawyehsu:
! DON'T merge until consensus is made.

niheaven:
This is why I made https://github.com/ScoopInstaller/Versions/pull/1606 and this could be updated to v4. But this may break some codes in Scoop Core (https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/lib/core.ps1#L318-L323 and https://github.com/ScoopInstaller/Scoop/blob/f93028001fbe5c78cc41f59e3814d2ac8e595724/libexec/scoop-checkup.ps1#L32-L34).

Nevertheless, update wix to 4 and I'll update core soon (most users are install dark directly IMO).

niheaven (approved):

niheaven:
This could be merged 'cos `wix burn extract` does the same thing as `dark`, see https://github.com/ScoopInstaller/Scoop/pull/5871

PR #5612, opened 2024-03-15
swift@5.10: Fix pre_install

### LINKED ISSUE #5581: swift@5.10: decompress error
```
Checking hash of swift-5.10-RELEASE-windows10.exe ... ok.
Uninstalling 'swift' (5.9.2)
Removing shim 'plutil.shim'.
Removing shim 'plutil.exe'.
Unlinking ~\scoop\apps\swift\current
Installing 'swift' (5.10) [64bit] from 'main' bucket
Loading swift-5.10-RELEASE-windows10.exe from cache.
Running pre_install script...
ERROR Exit code was 1620!
Failed to extract files from C:\Users\Simon\scoop\apps\swift\5.10\extract_folder\AttachedContainer\a0.
Log file:
  ~\scoop\apps\swift\5.10\extract_folder\AttachedContainer\msi.log

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/ScoopInstaller/Main/issues/new?title=swift%405.10%3a+decompress+error
```

The log file: `msi.log` is empty. 

It looks like the Swift team have changed how they make their archives. It could just be a new version, I honestly don't know. 

I reported the same issue here: https://github.com/ScoopInstaller/Versions/pull/1620

shiena:
Fix the pre_install script because the msi file and the path to be expanded have changed.

Closes #5581

### DIFF
--- a/bucket/swift.json
+++ b/bucket/swift.json
@@ -35,18 +35,20 @@
         "Move-Item -Path \"$dir\\Swift\\Platforms\\$ver\\Windows.platform\" -Destination \"$dir\\Swift\\Platforms\"",
         "Move-Item -Path \"$dir\\Swift\\Toolchains\\$($ver)+Asserts\\usr\" -Destination \"$dir\\Swift\\Toolchains\"",
         "Move-Item -Path \"$dir\\Swift\\Tools\\$ver\\*\" -Destination \"$dir\\Swift\\Tools\"",
-        "Remove-Item \"$dir\\Swift\\Platforms\\$ver\" -Recurse | Out-Null",
-        "Remove-Item \"$dir\\Swift\\Toolchains\\$ver+Asserts\" -Recurse | Out-Null",
-        "Remove-Item \"$dir\\Swift\\Tools\\$ver\" -Recurse | Out-Null",
-        "Remove-Item \"$dir\\extract_folder\" -Recurse | Out-Null"
+        "Remove-Item \"$dir\\Swift\\Platforms\\$ver\" | Out-Null",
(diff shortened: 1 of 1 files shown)


tech189 (changes requested):
Thank you for your fixes. However we can't leave a version number (`5.10.0`) hardcoded in `pre_install`, `env_add_path`, and `env_set` because it will change in the next release. We can't just use `$version` either because that is `5.10` not `5.10.0`. Perhaps we can rename all paths with `5.10.0` to `5.10`?

For reference if you need it, I found the longer version number can be extracted with `$xml.BurnManifest.Registration.Version`

shiena:
Thanks for the review. I did think that might be a problem. But the method you suggested is not sufficient because embedding the variable in env_add_path does not expand it. How about removing the version from the path instead?

**before**
```
"env_add_path": [
    "Swift\\Runtimes\\5.10.0\\usr\\bin",
    "Swift\\Toolchains\\5.10.0+Asserts\\usr\\bin",
    "Swift\\Tools\\5.10.0"
],
```

**after**
```
"env_add_path": [
    "Swift\\Runtimes\\usr\\bin",
    "Swift\\Toolchains\\Asserts\\usr\\bin",
    "Swift\\Tools"
],
```

tech189:
I think that would be the best option since Scoop already separates by version number in `scoop\apps\swift` so that version folder seems redundant to me. But please test as I do not personally use this package.

tech189 (approved):

tech189:
Thanks for your help!

PR #5644, opened 2024-03-25
boost: Update to version 1_85_0_b1_rc1

brian6932:
So I accidentally introduced the following error
```pwsh
Running pre_install script...
Rename-Item : Cannot process argument because the value of argument "path" is not valid. Change the value of the
"path" argument and run the operation again.
At line:1 char:1
+ Rename-Item "$dir\lib$($architecture.Substring(0, 2))-msvc-*.*" lib
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Rename-Item], PSArgumentException
    + FullyQualifiedErrorId : Argument,Microsoft.PowerShell.Commands.RenameItemCommand
```
I wasn't aware that wildcards couldn't be used within PowerShell 5's `Rename-Item`, and since I don't use PowerShell 5 on any regular basis, someone had to tell me 😅.

### DIFF
--- a/bucket/boost.json
+++ b/bucket/boost.json
@@ -14,7 +14,7 @@
         }
     },
     "innosetup": true,
-    "pre_install": "Rename-Item \"$dir/lib$($architecture.Substring(0, 2))-msvc-*.*\" lib",
+    "pre_install": "Get-ChildItem \"$dir/lib$($architecture.Substring(0, 2))-msvc-*.*\" | Rename-Item -NewName lib",
     "env_set": {
         "BOOST_ROOT": "$dir",
         "Boost_INCLUDE_DIR": "$dir\\boost"
--- a/bucket/boost.json
+++ b/bucket/boost.json
@@ -1,16 +1,16 @@
 {
-    "version": "1_84_0",
+    "version": "1_85_0_b1_rc1",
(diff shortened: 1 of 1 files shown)


niheaven:
For beta version, add it to https://github.com/ScoopInstaller/Versions/

brian6932:
It's a library, the experimental usage of it would be building it from it's head, these are generally more stable 🤷

niheaven:
Beta versions will never be in 'main' bucket except for ones that never publish stable versions even if it is stable enough. As you said, the beta versions of boost are published regularly, then make a beta manifest in 'version' bucket as other apps.

brian6932:
That's fine and all, but I'm asking you to reopen so I can reset to before that commit, only reason it was even altered was because tests were failing.

niheaven:
> That's fine and all, but I'm asking you to reopen so it can be reset to before that commit, only reason it was even altered was because tests were failing, the branch only has 1 commit currently.

I couldn't reopen this PR and don't know why 😂. So could you submit a new one? I'll review it soon.

PR #5655, opened 2024-03-27
yt-dlp: add plugins directory

silver886:
According to [document](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#installing-plugins), there is another path for loading plugins.

### DIFF
--- a/bucket/yt-dlp.json
+++ b/bucket/yt-dlp.json
@@ -20,7 +20,7 @@
     "pre_install": "if (-not (Test-Path \"$persist_dir\\yt-dlp.conf\")) { New-Item \"$dir\\yt-dlp.conf\" -ItemType file | Out-Null }",
     "persist": [
         "yt-dlp.conf",
-        "ytdlp_plugins"
+        "yt-dlp-plugins"
     ],
     "checkver": "github",
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


tech189 (commented):
Thanks for this, does this mean that plugins in future should be stored in `yt_dlp_plugins` but old ones still work if they are located in `ytdlp_plugins`?

If so, we could write a script to move plugins to the new folder.

silver886:
I haven't use old plugins and I'm not sure about that.
However, the plugin I used need to be placed in the new folder and it wouldn't work if be placed in the old directory.

I think we should leave the folder contents as is to avoid any impact to users.
The plugin structure seems different between these two plugin system.
Ref. https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/plugins.py#L138-L171

tech189:
Okay, I see. Isn't `yt_dlp_plugins` the new folder name though? Your PR has `yt-dlp-plugins` instead (with hyphens not underscores).

dyphire:
I'm having this issue and can only manually create the `yt-dlp-plugins` directory now. Hopefully this PR will be merged soon.

tech189 (approved):

PR #5657, opened 2024-03-27
nodejs-lts: fix checkver and add arm64 version

azrdn:
Due to the [redesign](https://nodejs.org/en/blog/announcements/diving-into-the-nodejs-website-redesign) of the Node.js website, the checkver of the nodejs-lts package is broken, this pull request fixes it, and also adds the arm64 version

### DIFF
--- a/bucket/nodejs-lts.json
+++ b/bucket/nodejs-lts.json
@@ -1,23 +1,23 @@
 {
-    "version": "20.12.0",
+    "version": "20.12.1",
     "description": "As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications. (Long Term Support)",
     "homepage": "https://nodejs.org",
     "license": "MIT",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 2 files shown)


eteeselink:
For what it's worth: I'm not sure how to review this, but I tested it by running this:

```sh
scoop install https://raw.githubusercontent.com/ScoopInstaller/Main/0f5202c216bf79be81549880c86f33eef1218be9/bucket/nodejs-lts.json
```
(that's the raw URL of the file in this PR)

and it worked flawlessly on my computer. Thanks!

f-cramer:
Seeing how long it takes to for those PRs to get merged, we should change the checkver script to instead get the latest version from https://nodejs.org/dist/index.json to no longer be dependent on the layout of nodejs.org.
See https://github.com/ScoopInstaller/Nonportable/pull/295 for my PR in the nonportable bucket that does exactly that.

Issue #5092, opened 2023-09-11
[Request]: Bun


azmankudus:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Non-GUI tool
- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

Bun

### Description

About Incredibly fast JavaScript runtime, bundler, test runner, and package manager – all in one

### Homepage

https://bun.sh

### Download Link(s)

https://github.com/oven-sh/bun/releases

### Some Indication of Popularity/Repute

Github Start 49.8k
Used by more than 55.6k users

rashil2000:
They don't provide Windows builds yet

sitiom:
> They don't provide Windows builds yet

They do now for canary:

- https://github.com/ScoopInstaller/Versions/pull/1449

sitiom:
> They don't provide Windows builds yet

They do now for canary:

- https://github.com/ScoopInstaller/Versions/pull/1449

Also, another important news https://github.com/oven-sh/bun/issues/43#issuecomment-1791719280:

> In 11 days, Windows support for Bun will land.

NemoStein:
They do now, fully native
https://bun.sh/blog/bun-v1.1

PR #5672, opened 2024-04-01
bun(-baseline): Add version 1.1.0

sitiom:
Closes https://github.com/ScoopInstaller/Main/issues/5092

### DIFF
--- a/bucket/bun-baseline.json
+++ b/bucket/bun-baseline.json
@@ -1,0 +1,38 @@
+{
+    "version": "1.1.0",
+    "description": "Incredibly fast JavaScript runtime, bundler, transpiler and package manager - all in one. (Baseline build)",
+    "homepage": "https://bun.sh/",
+    "license": "MIT",
+    "suggest": {
+        "vcredist": "extras/vcredist2022"
+    },
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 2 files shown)


sitiom:
CC @chawyehsu

chawyehsu (approved):

sitiom:
@chawyehsu Can you explain to me what it means to "leak" DLL imports? Perhaps the [`bun.sh/install.ps1`](https://bun.sh/install.ps1) script is leaking as well?
```powershell
  $IsBaseline = $ForceBaseline
  if (!$IsBaseline) {
    $IsBaseline = !( `
      Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);' `
        -Name 'Kernel32' -Namespace 'Win32' -PassThru `
    )::IsProcessorFeaturePresent(40);
  }
```

chawyehsu:
>  "leak" DLL imports?

That means leaking imported methods to the current PS session, literally. By _directly_ using `Add-Type` to load assemblies to current session there is no way to **unload** it unless starting a new session. In most cases, this won't cause issues as imported methods have namespaces specified, but one may not want to have imported methods persistent in current session.

Yeah it's leaking the imported method as well. You may validate `[Win32.Kernel32]::IsProcessorFeaturePresent` available after runing the install script.

PR #5673, opened 2024-04-02
helm-ls: Add version 0.0.14

fnuttens:
See https://github.com/mrjosh/helm-ls


aliesbelik:
The project doesn't match the [Main criteria](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket) (in terms of popularity/maturity).
Guess you could promote it to Extras bucket instead.

PR #5683, opened 2024-04-05
add boost-14.2 so can install for VS 2019 build tools

ahornby:
add boost-14.2 so can install for VS 2019 build tools

tested with: `scoop install C:\Users\alex\local\ScoopInstaller\bucket\boost-14.2.json`


niheaven:
Please move to versions bucket and rename to boost-msvc142

ahornby:
> Please move to versions bucket and rename to boost-msvc142

@niheaven done, raised as https://github.com/ScoopInstaller/Versions/pull/1669

PR #5694, opened 2024-04-08
haskell: use haskell.org instead of ghc github mirror and update to version 9.8.2

luca-schlecker:
The GitHub ghc mirror seems outdated (newest release there is 9.4.1).
This PR will thus switch to [ghc from haskell.org](https://www.haskell.org/ghc/download.html) and also update the haskell package to 9.8.2.

### DIFF
--- a/bucket/haskell.json
+++ b/bucket/haskell.json
@@ -12,7 +12,7 @@
     },
     "architecture": {
         "64bit": {
-            "url": "https://github.com/commercialhaskell/ghc/releases/download/ghc-9.4.1-release/ghc-9.4.1-x86_64-unknown-mingw32.tar.xz",
+            "url": "https://downloads.haskell.org/~ghc/9.4.1/ghc-9.4.1-x86_64-unknown-mingw32.tar.xz",
             "hash": "fdc2d78b8a978e712a4edcc1628e36a36be736d8202107d2d61999ac7a9dc5d0",
             "extract_dir": "ghc-9.4.1-x86_64-unknown-mingw32"
         }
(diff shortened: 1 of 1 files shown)


bucket/haskell.json
```diff
@@ -29,14 +29,14 @@
     ],
     "env_add_path": "lib\\bin",
     "checkver": {
-        "url": "https://api.github.com/repositories/46509594/tags",
-        "jsonpath": "$..name",
-        "regex": "ghc-([\\d\\.]+)-release"
+        "url": "https://downloads.haskell.org/~ghc/",
+        "regex": "(\\d\\.\\d\\.\\d)/",
```
tech189, line 33: What about the x.x.x.YYYYMMDD releases? Or if a hotfix to 9.8.2 is released like 9.8.2.1? This regex will miss them out. Maybe it's better to use the release posted as the current stable in https://www.haskell.org/ghc/download.html

tech189:
```suggestion
        "url": "https://downloads.haskell.org/~ghc/",
        "regex": ">([\\d.]+)/",
```
How about this to catch any number of point versions? This is the standard regex we use without catching versions with words in it like alpha, beta, etc.

luca-schlecker: Wouldn't this allow `..` to be parsed as a valid version?
Maybe something like `>(\\d(\\.\\d)*)/`?

luca-schlecker: I'll push my proposed regex so you can merge it quickly in case you agree with it, if not, I'll be happy to swap it out again. 😊

PR #5696, opened 2024-04-09
odbc2parquet: Add version 6.0.0

igitur:
Add new package odbc2parquet ( https://github.com/pacman82/odbc2parquet )

### DIFF
--- a/bucket/odbc2parquet.json
+++ b/bucket/odbc2parquet.json
@@ -1,0 +1,28 @@
+{
+    "version": "6.0.0",
+    "description": "A command line tool to query an ODBC data source and write the result into a parquet file.",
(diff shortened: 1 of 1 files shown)


niheaven:
Just use `github` as checkver.

igitur:
> Just use `github` as checkver.

Don't we need the regex to extract the correct version number from the tag name?

niheaven:
It's standard version number format in release/latest tag, so we don't need the regex. Ref: https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate#special-cases

igitur:
Thanks. I didn't expect it to handle the `v`. Will update soon.

niheaven (approved):

PR #5719, opened 2024-04-17
MAVEN: Update download URL

delassio:

### DIFF
--- a/bucket/maven.json
+++ b/bucket/maven.json
@@ -6,7 +6,7 @@
     "suggest": {
         "JDK": "java/openjdk"
     },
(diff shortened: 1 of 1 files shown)


niheaven:
Why do this change?

delassio:
> Why do this change?

changing the download server from archive to dlcdn server

maiq123:
The dlcdn server doesn't have older versions. Trying to install maven@3.6.0 but dlcdn doesn't have that. This change has made it difficult to install older versions of maven. Can we have a new feature of 'old_url' in the manifest to specify some and old url and if the package is not available in curren 'url' than it can also check the old url for the version?

Issue #5860, opened 2024-05-22
azure-cli@2.61.0: decompress error


MrPandey01:
```
azure-cli: 2.49.0 -> 2.61.0
Updating one outdated app:
Updating 'azure-cli' (2.49.0 -> 2.61.0)
Downloading new version
azure-cli-2.61.0-x64.msi (69.7 MB) [=================================================================================================================================================================================================] 100%
Checking hash of azure-cli-2.61.0-x64.msi ... ok.
Uninstalling 'azure-cli' (2.49.0)
Removing shim 'az'.
Removing shim 'az.cmd'.
Unlinking ~\scoop\apps\azure-cli\current
Installing 'azure-cli' (2.61.0) [64bit] from 'main' bucket
Loading azure-cli-2.61.0-x64.msi from cache
Extracting azure-cli-2.61.0-x64.msi ... ERROR Exit code was 1618!
Failed to extract files from C:\Users\arpande1\scoop\apps\azure-cli\2.61.0\azure-cli-2.61.0-x64.msi.
Log file:
  ~\scoop\apps\azure-cli\2.61.0\msi.log

Please try again or create a new issue by using the following link and paste your console output:
https://github.com/ScoopInstaller/Main/issues/new?title=azure-cli%402.61.0%3a+decompress+error
```

o-l-a-v:
I got error code 1304.

```log
=== Logging started: 06.06.2024  10:02:50 ===
Error 1304. Error writing to file: C:\Users\olav.birkeland\scoop\apps\azure-cli\2.61.0\_tmp\SourceDir\Microsoft SDKs\Azure\CLI2\Lib\site-packages\azure\cli\command_modules\dla\vendored_sdks\azure_mgmt_datalake_analytics\catalog\models\data_lake_analytics_catalog_secret_create_or_update_parameters.pyc. Verify that you have access to that directory.
=== Logging stopped: 06.06.2024  10:06:19 ===
```

What a mastadon of a file path:

* <https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/dla/vendored_sdks/azure_mgmt_datalake_analytics/catalog/models/data_lake_analytics_catalog_secret_create_or_update_parameters.py>

While installing using the ZIP went through fast with no issues. Lets hope #5885 gets merged.

Issue #5832, opened 2024-05-15
azure-cli@2.60.0: decompress error


aaryannnair:
Extracting azure-cli-2.60.0-x64.msi ... ERROR Exit code was 1603!
Failed to extract files from C:\Users\AaryanNair\scoop\apps\azure-cli\2.60.0\azure-cli-2.60.0-x64.msi.
Log file:
  ~\scoop\apps\azure-cli\2.60.0\msi.log

eagleeyetom:
Same happens with AWS. Reverting to 7-zip@23.0.1 fixed the issue. At least for me.

Issue #5744, opened 2024-04-24
azure-cli@2.59.0: decompress error


korfgunt:
Hello,
when using scoop to install azure-cli an error occurs during decompression

scoop install azure-cli
WARN Previous failed installation of azure-cli will be deleted.
ERROR You need administrator rights to uninstall global applications.
WARN Cleaning up the previous failed installation of azure-cli.
ERROR 'azure-cli' was not installed correctly.
'azure-cli' has been uninstalled.
Installing 'azure-cli' (2.59.0) [64bit] from the 'main' bucket
Load azure-cli-2.59.0-x64.msi from the cache
Extracting azure-cli-2.59.0-x64.msi ... ERROR Exit code was 1603!
Failed to extract files from C:\Gothaer\portable\scoop\apps\azure-cli\2.59.0\azure-cli-2.59.0-x64.msi.
Log file:
  C:\Gothaer\portable\scoop\apps\azure-cli\2.59.0\msi.log

The msi log contains this
=== Logging started: 24.04.2024 07:46:11 ===
Error 1304. error writing to the file: C:\Gothaer\portable\scoop\apps\azure-cli\2.59. 0\_tmp\SourceDir\Microsoft SDKs\Azure\CLI2\Lib\site-packages\azure\cli\command_modules\dla\vendored_sdks\azure_mgmt_datalake_analytics\catalog\models\data_lake_analytics_catalog_secret_create_or_update_parameters.pyc. Make sure that you have access to this directory.
=== Logging stopped: 04/24/2024 07:48:01 ===

Can you help?

Issue #5637, opened 2024-03-21
azure-cli@2.58.0: decompress error


BCoskun:

alixmacdonald10:
I had the same issue and managed to fix it. I detailed it in #5623

PR #5885, opened 2024-06-05
azure-cli: Use zip archive

o-l-a-v:
Switch Azure CLI to use ZIP instead of MSI.

Seems ZIP has been available since 2024-01-30, ref <https://github.com/Azure/azure-cli/pull/27911>.

Benefits:

* Should be faster to install and should create shorter paths which might help with #5300.
* Less complex to extract a ZIP vs. MSI.

~~Cons:~~ Edit: No cons as I see it.

* Only x64 is available as ZIP, not x86.
  * Related feature request: <https://github.com/Azure/azure-cli/issues/29104>
  * Edit: Not a problem, Azure CLI authors recommends against using x86.
* Download is much slower because Azure CLI ZIP is not available through a CDN, but an Azure Storage Account in US West only.
  * Related feature request: <https://github.com/Azure/azure-cli/issues/28856>
  * Edit: Not a con because ScoopInstaller/Main points to the same Azure Storage Account for MSI.

Most relevant issue: #5300. Other open issues:

* #5860
* #5832
* #5744
* #5637
* #5623

### DIFF
--- a/bucket/azure-cli.json
+++ b/bucket/azure-cli.json
@@ -9,20 +9,15 @@
     ],
     "architecture": {
         "64bit": {
-            "url": "https://azcliprod.blob.core.windows.net/msi/azure-cli-2.61.0-x64.msi",
-            "hash": "e6906997110d6acfb51a934b1d18b21fb2f314dab6b653644fd7d769e72e668c"
-        },
-        "32bit": {
-            "url": "https://azcliprod.blob.core.windows.net/msi/azure-cli-2.61.0.msi",
-            "hash": "f765b484f8279fdd909d884539ada5baf2aaf897e30c2ef3de33cdae42180956"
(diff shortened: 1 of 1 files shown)


silverkorn:
Notice: This breaks backward compatibility with installing older versions not having the ZIP archive.
On our end, it is required to make it work with legacy AKS credentials w/ `kubectl` & `kubelogin`.

I think we should not assume that new formats must be used too quickly, although I admit that it's been more than a year...

o-l-a-v:
You could maybe:

* Maintain your own bucket
* Download the app manifest at a given commit, then `scoop install <app_name>.json`

Would be nice if Scoop supported specifying `scoop install <bucket>/<app>@<commit>`, or `scoop install <url_to_file.json>`.

PR #6070, opened 2024-08-03
fnm: Remove leftovers when `uninstall --purge`

o-l-a-v:
fnm leaves behind two paths when uninstalled:

* `%LOCALAPPDATA%\fnm_multishells`
* `%APPDATA%\fnm`

Added ability to have them nuked when uninstalling with `--purge`.

Note: For some reason `[System.IO.Directory]::Delete()` failed to delete one of the directories with an error about access denied, but `Remove-Item -Recurse` worked. So using that.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/fnm.json
+++ b/bucket/fnm.json
@@ -1,4 +1,5 @@
 {
+    "$schema": "https://raw.githubusercontent.com/ScoopInstaller/Scoop/master/schema.json",
     "version": "1.37.1",
     "description": "Cross-platform Node.js version switcher",
     "homepage": "https://github.com/Schniz/fnm",
@@ -11,6 +12,19 @@
         }
     },
     "bin": "fnm.exe",
+    "post_uninstall": [
+        "if ($purge) {",
+        "    $Directories = [string[]](",
+        "        ('{0}\\fnm_multishells' -f $env:LOCALAPPDATA),",
+        "        ('{0}\\fnm' -f $env:APPDATA)",
+        "    )",
+        "    $Directories.ForEach{",
+        "        if ([System.IO.Directory]::Exists($_)) {",
+        "            $null = Remove-Item -Path $_ -Recurse -Force",
+        "        }",
+        "    }",
+        "}"
+    ],
     "checkver": "github",
     "autoupdate": {
         "architecture": {
--- a/bucket/fnm.json
+++ b/bucket/fnm.json
@@ -19,7 +19,7 @@
         "        ('{0}\\fnm' -f $env:APPDATA)",
         "    )",
         "    $Directories.ForEach{",
-        "        if ([System.IO.Directory]::Exists($_)) {",
+        "        if (Test-Path -Path $_ -PathType 'Container') {",
         "            $null = Remove-Item -Path $_ -Recurse -Force",
         "        }",
         "    }",


Lutra-Fs:
I guess we should set `FNM_DIR` up as well. See https://github.com/Schniz/fnm/blob/1ebc90decf58c2176dcbd4f8ac00ecb3adb1de12/docs/commands.md?plain=1#L33
and possibly add related persist option. It will affect `%APPDATA%\fnm` but we still need to handle `%LOCALAPPDATA%\fnm_multishells`

o-l-a-v:
I (think I) added your suggestions @Lutra-Fs.

* Setting env variable `FNM_DIR`
* Persist `aliases` and `node-versions` directories
* Purge will remove any `FNM_*` environment variables

What you think? 😊

Lutra-Fs:
I'll spin up a VM to install the updated manifest, try a couple of small projects, and then uninstall. If everything looks good, I'll go ahead and merge. One question: since the manifest uses `env_set`, do we also need to explicitly unset those variables in the `post_uninstall` script, or does Scoop handle cleanup automatically? I'm checking the source now to confirm.

Lutra-Fs:
Can you kindly add a note to tell the user that the new config location will be under our directory? See https://github.com/ScoopInstaller/Main/pull/6462#issuecomment-2812476661 for an example.

bucket/fnm.json
```diff
@@ -3,14 +3,51 @@
...
+        "    # Environment variables",
+        "    $EnvScope = [string]$(if($global){'Machine'}else{'User'})",
+        "    [System.Environment]::GetEnvironmentVariables($EnvScope).GetEnumerator().Where{",
+        "        $_.'Name' -like 'FNM_*'",
+        "    }.ForEach{",
+        "        [System.Environment]::SetEnvironmentVariable(",
+        "            $_.'Name',",
+        "            $EnvScope,",
+        "            $null",
+        "        )",
```
Lutra-Fs: The method signature for [`SetEnvironmentVariable`](https://learn.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=net-8.0) is:

```csharp
public static void SetEnvironmentVariable(string variable, string? value, EnvironmentVariableTarget target);
```

In the current code, the second argument is `$EnvScope` and the third is `$null`. This does not match the intended usage. The third parameter should be the environment variable target (`User` or `Machine`), and the value should be `$null` when you want to remove the variable.

For example:

```powershell
[Environment]::SetEnvironmentVariable($_.Name, $null, $EnvScope)
```

where `$EnvScope` is either `'User'` or `'Machine'` based on `--global`.

o-l-a-v:
One potential problem is that `FNM_DIR` is not set in current session / process after `scoop install fnm`. Related:

* <https://github.com/ScoopInstaller/Scoop/issues/6440>

I'll add info on that too.

o-l-a-v:
Worked around it by setting `$env:FNM_HOME` in `post_install`.

Lutra-Fs:
NVM, just noticed that you have fixed it. Reviewing the wrong version of the code 😢

Lutra-Fs (approved):

PR #6117, opened 2024-08-19
wcurl: Update manifest

sitiom:
- Fix `homepage` order
- Point `url` to the `wcurl` script instead of the entire archive.
- Move `depends` to `suggest` & remove `busybox` dependency (Scoop automatically creates a shim to bash)

### DIFF
--- a/bucket/zola.json
+++ b/bucket/zola.json
@@ -1,12 +1,12 @@
 {
-    "version": "0.23.5",
+    "version": "0.23.6",
     "description": "A fast static site generator in a single binary with everything built-in",
     "homepage": "https://www.getzola.org",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/getzola/zola/releases/download/v0.23.5/zola-v0.23.5-x86_64-pc-windows-msvc.zip",
-            "hash": "878fccd4ec4f4cfe7babbff1c5bae2abac9c47181ea383ccde2322dcdf445be2"
+            "url": "https://github.com/getzola/zola/releases/download/v0.23.6/zola-v0.23.6-x86_64-pc-windows-msvc.zip",
+            "hash": "79179284f816febc281adbedb27189806cf22f83e8be4d0fca6835fbeeb09587"
         }
     },
     "bin": "zola.exe",


sitiom (commented):
Issue: `wcurl` doesn't show output due to a bug with Scoop (https://github.com/ScoopInstaller/Scoop/issues/6112)

niheaven:
Issue: wcurl doesn't show output due to a bug with Scoop (https://github.com/ScoopInstaller/Scoop/issues/6112)

That's why the origin manifest use an explicit `sh` in shim creation. And why install such a big `git` app just for a `sh` shell?

BTW, `curl` and `sh` is depends, not suggests for this script.

sitiom:
I'm trying to copy the [neofetch](https://github.com/ScoopInstaller/Main/blob/master/bucket/neofetch.json) manifest here. `git` is just a suggestion, the machine may already have `bash` installed (WSL) which I assume the shim tries to find.

> BTW, `curl` and `sh` is depends, not suggests for this script.

`curl` is already built-in to Windows 10. Most "dependencies" in these buckets use `suggests` instead of `depends` to account for dependencies installed outside Scoop, don't they?

niheaven:
> Most "dependencies" in these buckets use suggest instead of depends to account for dependencies installed outside Scoop, don't they?

The additional feature depends is moved to suggests, but functional depends (`curl` for `wcurl`) is still depends, e.g., `r` for `rstudio`.

sitiom:
> The additional feature depends is moved to suggests, but functional depends (`curl` for `wcurl`) is still depends, e.g., `r` for `rstudio`.

`vcredist` as a functional dependency is listed as `suggest` though 🤔

sitiom:
> Issue: `wcurl` doesn't show output due to a bug with Scoop ([ScoopInstaller/Scoop#6112](https://github.com/ScoopInstaller/Scoop/issues/6112))

- Should be fixed by https://github.com/ScoopInstaller/Scoop/pull/6114

niheaven:
> vcredist as a functional dependency is listed as suggest though 🤔

It is common for many machines to have `vcredist` already installed, so `vcredist` is marked as `suggests`.

z-Fng:
Looks like this is stale (no updates in nearly two years). Safe to close?

sitiom:
I'm not sure why this was closed, continued in #8506

smokad2000-ops:
> [https://github.com/ScoopInstaller/Main/pull/6117#issuecomment-2296138462](https://github.com/ScoopInstaller/Main/pull/6117#issuecomment-2296138462) WSL's `bash` is very slow as I tried... So when I modify the manifest, I just keep `sh` and add `busybox` as depends.

To be honest, I share the same concern. Since Git no longer creates a bash shim (#7459), running bash will default to WSL's bash whenever WSL is enabled. Until this issue is resolved, I think it's best to keep the original manifest design. I'd suggest dropping these changes; otherwise, this PR will likely remain on hold.

Issue #5602, opened 2024-03-13
[Request]: git-absorb


KoltesDigital:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Non-GUI tool
- [X] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [X] English interface (or at least English documentation)
- [X] Latest stable version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

git-absorb

### Description

git absorb will automatically identify which commits are safe to modify, and which staged changes belong to each of those commits. It will then write fixup! commits for each of those changes. With the --and-rebase flag, these fixup commits will be automatically integrated into the corresponding ones.

### Homepage

https://github.com/tummychow/git-absorb

### Download Link(s)

https://github.com/tummychow/git-absorb/releases/download/0.6.12/git-absorb-0.6.12-x86_64-pc-windows-msvc.zip

### Some Indication of Popularity/Repute

3.1k stars
59 forks

KoltesDigital:
I'm willing to make a PR for that BTW. Didn't know if it goes to main or extras, but most git plugins seem to be in main.

selfdocumentingcode:
I think this issue can be closed.

KoltesDigital:
It seems so. I'm already using it!

PR #6205, opened 2024-09-21
git-absorb: Add version 0.6.15

daniel-liuzzi:
Relates to #5602 

### DIFF
--- a/bucket/git-absorb.json
+++ b/bucket/git-absorb.json
@@ -3,12 +3,20 @@
     "description": "git commit --fixup, but automatic",
     "homepage": "https://github.com/tummychow/git-absorb",
     "license": "BSD-3-Clause",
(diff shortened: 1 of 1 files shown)


HUMORCE (changes requested):
For 64bit programs, `architecture` is required.

daniel-liuzzi:
@HUMORCE according to the [App Manifest wiki][1], "if the app is not 32bit, architecture **can be** used to wrap the differences" so I understood it was optional, especially given there's only a single binary. In any case, I already added architecture.

[1]: https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties

HUMORCE:
Yeah, but It's mandatory for all official buckets.

> If the application provides only a 32bit download, the architecture field is not required. In all other cases, architecture field is mandatory.
> 
> [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

HUMORCE (approved):

PR #6374, opened 2024-12-09
mise: Add Version 2025.1.6

FilipDeVos:
This pull request adds the manifest to install the "mise-en-place" tool from https://github.com/jdx/mise/

Closes #6329

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #6453, opened 2025-01-13
mise: manifest updated to more portable config

ev-dev:
Included manifest changes suggested by @chawyehsu in PR comment which sets the internal mise configurable env var MISE_DATA_DIR to a directory that scoop persists

Relates to #6374

- removed install/uninstall scripts
- set env var `MISE_DATA_DIR` to a directory inside scoop and persist it
- add the directory mise places its own shims to the Path env var
- set env var `MISE_GLOBAL_CONFIG_FILE` to file inside app install dir and persist it
- added pre_install script to check mise config.toml exists or is created as a file
- added note about configuring specific shells with link to mise docs

### DIFF
--- a/bucket/mise.json
+++ b/bucket/mise.json
@@ -17,12 +17,12 @@
     "bin": "bin/mise.exe",
     "env_set": {
         "MISE_DATA_DIR": "$dir\\mise",
-        "MISE_GLOBAL_CONFIG_ROOT": "$dir\\config"
+        "MISE_GLOBAL_CONFIG_FILE": "$dir\\config.toml"
     },
     "env_add_path": "mise\\shims",
     "persist": [
         "mise",
-        "config"
+        "config.toml"
     ],
     "notes": "See documentation for notes on configuring your shell: https://mise.jdx.dev/installing-mise.html",
     "checkver": {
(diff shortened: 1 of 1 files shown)


ev-dev:
Reverting to draft after realizing the main config file mise uses also defaults to a path outside scoop. This also has a configurable env var so makes sense to set in manifest allowing the mise global config file to persist within scoop

chawyehsu (changes requested):

chawyehsu (approved):

stevehipwell (approved):
LGTM

chawyehsu:
> @chawyehsu is there a reason you didn't merge this after approving it?

I just forgot to

eg-bernardo:
With this change the config is lost on every update.

I think the issue is that it's using `$dir` where `$persist_dir` should be used

eg-bernardo:
I was wrong, because `config.toml` is in `persist` key of manifest.
This change still had the issue that it did not migrate the config from the previous location, so the config was lost on the update to this version, but future updates should not lose the config.

ev-dev:
Yea good point @eg-bernardo really wish I hadn't forgotten to include that before the manifest went live in the `main` bucket 🤦 #6478 

It may be best to submit a quick patch that just prompts new user installs with a warning message about manually migrating any previous configs/data.

ev-dev:
#6484 adds warning message for now

Issue #6477, opened 2025-01-21
[Bug]: Mise doesn't work under the `scoop` dir


stevehipwell:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

mise

### Expected/Current Behaviour

I expect to be able to follow the Mise [getting started guide](https://mise.jdx.dev/getting-started.html) but I get an error unless I manually set the `MISE_DATA_DIR` & `MISE_GLOBAL_CONFIG_FILE` env variables to a path outside of the scoop dir and then modify my path to use the updated data directory.

### Steps to Reproduce

```console
scoop install mise
mise use --global node@22
```

### Possible Solution

Remove the code to configure mise to run under the scoop dir.

### Scoop and Buckets Version

```console
Current Scoop version:
859d1db5 (HEAD -> master, tag: v0.5.2, origin/master, origin/HEAD) chore(release): Bump to version 0.5.2 (#6080)

'main' bucket:
f3792e06a (HEAD -> master, origin/master, origin/HEAD) mysql: Update to version 9.2.0

'extras' bucket:
9408ed609 (HEAD -> master, origin/master, origin/HEAD) trafficmonitor: Update to version 1.85

'nerd-fonts' bucket:
de210753 (HEAD -> master, origin/master, origin/HEAD) LXGWWenKaiTC: Update to version 1.510

'java' bucket:
5d5ab48c (HEAD -> master, origin/master, origin/HEAD) intellij-jbr21: Update to version 21.0.5-b792.48

'artifacthub' bucket:
0b68ea0 (HEAD -> main, origin/main, origin/HEAD) Scoop update for ah version v1.20.0

'volllly' bucket:
f1cc167 (HEAD -> main, origin/main, origin/HEAD) rotz: Update to version 1.0.0
```

### Scoop Config

```console
last_update      : 21/01/2025 12:45:08
alias            : @{upgrade=scoop-upgrade; ls=scoop-ls}
use_lessmsi      : True
scoop_branch     : master
scoop_repo       : https://github.com/ScoopInstaller/Scoop
use_sqlite_cache : True
```

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.4.6
PSEdition                      Core
GitCommitId                    7.4.6
OS                             Microsoft Windows 10.0.22631
[...]

PR #6484, opened 2025-01-24
mise add warning about migration

ev-dev:
- Adds warning message about manual config migration.
- Update to version 2025.1.13


ev-dev:
I'll look into scripting a migration of previous `mise` installs, but it may not be necessary. Is a warning message alone an acceptable UX?

stevehipwell:
@ev-dev it'd be best if scoop didn't modify existing `mise` configuration, the install script could check for non-scoop values in the env. See #6477.

ev-dev:
Indeed, I agree @stevehipwell this PR simply adds the warning message:

> This manifest does not currently migrate any previous mise installations!
> 
> Suggested:
> * Backup existing config.toml before fully uninstalling mise as documented.
> * Install with scoop then manually add the backed up options to the new config.toml that scoop now manages in persist

My thought was that the best way to ovoid an install conflict with previous installs, would be to have a clean install with scoop. I figured this message would at least be better than not addressing or mentioning previous installs no?

stevehipwell:
The whole `mise` Scoop installation breaks existing configuration, the warning message doesn't really deal with that as it suggest that the intended behaviour is to replace the existing env variables and paths. The best solution would be to use an install script to detect if `mise` has been configured already to point outside the Scoop dir and if so not change anything related to that.

HUMORCE:
I changed it to keep the synergy with #6677

PR #6556, opened 2025-02-19
jaeger: add version 2.3.0

ogerardin:

### DIFF
--- a/bucket/jaeger.json
+++ b/bucket/jaeger.json
@@ -1,0 +1,17 @@
+{
+    "version": "2.3.0",
+    "description": "Jaeger is a distributed tracing platform",
+    "homepage": "https://www.jaegertracing.io/",
+    "license": {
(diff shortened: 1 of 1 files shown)


ogerardin:
Note: I haven't been able to add autoupdate, as the publishers use a tricky system for releases where both v1 and v2 share a single github tag, which is the v1 version... 
So for example while 2.3.0 is the latest version, and you could extract that from the download page https://www.jaegertracing.io/download/ by using regex "Latest version.*(\\d\\.\\d\\.\\d)", to build the download URL you would also need to extract the v1 version, in this case 1.66.0, because the release tag is "v1.66.0" for both v1.66.0 and v2.3.0 😮 

The actual URL includes both: https://github.com/jaegertracing/jaeger/releases/download/v1.66.0/jaeger-2.3.0-windows-amd64.tar.gz

If anyone can make this work, feel free.

PR #6573, opened 2025-02-24
mergiraf: Correct license

wetneb:
Thanks for packaging Mergiraf!

I noticed that the license is wrong (see https://codeberg.org/mergiraf/mergiraf/src/branch/main/LICENSE.txt), so here is a PR to fix that.

Follow up to #6312 by @nithinphilips.

### DIFF
--- a/bucket/mergiraf.json
+++ b/bucket/mergiraf.json
@@ -2,7 +2,7 @@
     "version": "0.5.1",
     "description": "A syntax-aware git merge driver",
     "homepage": "https://mergiraf.org",
-    "license": "MIT",
+    "license": "GPLv3",
     "architecture": {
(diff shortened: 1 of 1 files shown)


bucket/mergiraf.json
```diff
@@ -2,7 +2,7 @@
     "version": "0.5.1",
     "description": "A syntax-aware git merge driver",
     "homepage": "https://mergiraf.org",
-    "license": "MIT",
+    "license": "GPLv3",
```
HUMORCE, line 5:
```suggestion
    "license": "GPL-3.0-only",
```

HUMORCE (changes requested):
SPDX license identifier

PR #6661, opened 2025-03-28
Submission: Edgescape Navigator

thedoggybrad:
Submission for Edgescape Navigator


thedoggybrad:
For an app to be acceptable for the main bucket, it should be:

a reasonably well-known and widely used developer tool (e.g. if it's a GitHub project, it should have at least 500 stars and 150 forks)

Edgescape Navigator definitely did not have this

PR #6700, opened 2025-04-09
icann-rdap: Add version 0.0.22

chrismeller:
Add a manifest for the [ICANN implementation of RDAP](https://github.com/icann/icann-rdap), the successor to WHOIS.

Note that I'm not sure about the installer script portion. The binary needs to be renamed before it can be shimmed, but I've never done that with Scoop before. Please let me know if there's a better approach.


chrismeller:
A note for anyone who might find this PR in the future, I've [added it to my scoop bucket](https://github.com/chrismeller/scoop-bucket/blob/main/bucket/icann-rdap.json).

PR #6462, opened 2025-01-16
bitwarden-cli: Persist data

Ra2-IFV:
Reopen for #6219
See https://github.com/bitwarden/cli/issues/117

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #6762, opened 2025-04-24
pnpm: persist global data and add to path

### LINKED ISSUE #6713: [Bug]: pnpm does not add a global bin directory
### Prerequisites

### Package Name

pnpm

Currently the global bin directory is not configured then installing pnpm.

### Steps to Reproduce

```console
scoop install main/pnpm
pnpm add -g @microsoft/inshellisense

console output:
ERROR The configured global bin directory "C:\Users\USER\AppData\Local\pnpm" is not in PATH
For help, run: pnpm help add
```

### Possible Solution

Ideally pnpm should create a custom global bin directory in the scoop persist file so that all packages installed globally can be retained. Furthermore, this directory should be added to the path.

I can see there is a `--global-dir: Specify a custom directory to store global packages` option which maybe can be used.

### Scoop and Buckets Version

```console
Current Scoop version:
859d1db5 (HEAD -> master, tag: v0.5.2, origin/master, origin/HEAD) chore(release): Bump to version 0.5.2 (#6080)

'main' bucket:
cc31681e6 (HEAD -> master, origin/master, origin/HEAD) trid: Update to version 2.24-25.04.12
```

### Scoop Config

```console
last_update          scoop_repo                              scoop_branch aria2-warning-enabled
-----------          ----------                              ------------ ---------------------
13-Apr-25 8:14:02 PM https://github.com/ScoopInstaller/Scoop master                       False
```

### PowerShell Version

```console
PSVersion                      7.5.0
GitCommitId                    7.5.0
OS                             Microsoft Windows 10.0.19044
```

### Additional Softwares

_No response_

jack-mil:
Following off of the similar python and rustup manifests, I have updated pnpm to make use of the $PNPM_HOME env variable in order to install global packages or cli utilities into a scoop-managed and persisted directory (instead of the default on windows: `~/AppData/Local/pnpm/global`

An alternative approach would be to:

a) Leave the pnpm default global directory as is, and simply add `~/AppData/Local/pnpm/global` to PATH when installing the manifest. This would also fix issues like #6713. But the application data would not be contained within the `~/scoop` folder (generally prefered).

b) Instruct users to run `pnpm setup` after installation, which will
    - create a home directory for the pnpm CLI (`~/AppData/Local/pnpm/global`)
    - adds the pnpm home directory to the PATH environment variable
    - copies the pnpm executable to the pnpm home directory
  In this case however, scoop will no longer manage pnpm at all, and it will be no different than installing manually (without scoop).

Note: $PNPM_HOME cannot be set to "$persist_dir\pnpm_data" because it seems that `pnpm` checks at runtime if the exact directory $PNPM_HOME is in $PATH, and doesn't follow symlinks.

Fixes: #6713


jack-mil:
I realize that this is a pretty major breaking change to those using this manifest already. Migration or at the very least a warning on updating is probably required. I would need advice on what is the canonical way to do that; if this change is even deemed necessary in the first place.

Lutra-Fs:
See https://github.com/ScoopInstaller/Main/pull/6462#issuecomment-2812476661

jack-mil:
I have finally gotten to this. I am not a `pnpm` user, so I don't know if this change is even wanted across the board. This manifest will have similar behavior to other scoop apps now, and at-least one issue has a desire to manage the global installation environment.

HUMORCE:
For #6713, We should provide script similar `pnpm setup`, i.e.: 

1. Set `$env:PNPM_HOME` to `$env:LOCALAPPDATA\pnpm`
2. Add `$env:PNPM_HOME` to `$env:PATH`

---

`cacheDir` and `stateDir` can set by `pnpm config`(config file), but the location of config file of pnpm(Windows) is fixed `$env:LOCALAPPDATA\pnpm\config\rc`.

- https://pnpm.io/cli/config
- https://pnpm.io/settings

`pnpm self-update` will install its newer version to `$env:PNPM_HOME`, when current version is standalone executable(Scoop provided), instead, use a NPM package manager will not caused this issue.

---

Personally, I prefer to install `pnpm` or other global packages via `bun add -g <package>`. `npm add -g pnpm` are recommend by pnpm official document. I understand that users need persistence to migrate/backup their data files quickly.

Issue #6787, opened 2025-05-01
[Bug]: PresentMon autoupdate not working


BoringBoredom:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

presentmon

### Expected/Current Behaviour

the latest version is 2.3.0: https://github.com/GameTechDev/PresentMon/releases
scoop is stuck on 2.1.1: https://github.com/ScoopInstaller/Main/blob/master/bucket/presentmon.json

BoringBoredom:
https://github.com/ScoopInstaller/Main/actions/runs/14782920316/job/41505587610#step:3:832

the issue seems to be that support for the 32 bit binary got dropped

PR #6788, opened 2025-05-02
presentmon: remove deprecated 32 bit version preventing auto update

BoringBoredom:
addresses https://github.com/ScoopInstaller/Main/issues/6787


HUMORCE:
#6789

PR #6790, opened 2025-05-02
presentmon: fix erroneous full removal of 32 bit version

BoringBoredom:
fix erroneous full removal of 32 bit version in https://github.com/ScoopInstaller/Main/commit/f5ef12b0c11026b93b40be0502cd20dd15f49749
only remove the 32 bit `autoupdate` entry


HUMORCE:
Keep it at local or version bucket.

atjoao:
So rn scoop downloads the x64 for x32?
Is that right?

PR #6848, opened 2025-05-27
use dvc.org url for download urls

skshetry:
We currently only support download URLs from dvc.org. The S3 URL is a result of us using s3 bucket for hosting packages, and we're in the process of transitioning to Cloudflare R2.

Related:
- #4942
- #4931


skshetry:
Hi, @HUMORCE, is it Scoop’s policy to always use the original source URL? My intention with this PR was to use dvc.org, since it’s likely to be more stable over time (since that's what we use in the website). 

What you’ve merged is something I did not intend.

PR #6954, opened 2025-07-09
mingit-busybox-lean: Add version 2.50.1

ddarfantasy:
This version of mingit uses the upstream busybox-lean package.

### DIFF
--- a/bucket/mingit-busybox-lean.json
+++ b/bucket/mingit-busybox-lean.json
@@ -1,0 +1,48 @@
+{
+    "version": "2.50.1",
+    "description": "Minimal Git for Windows (MinGit) is a reduced sized package designed to support application integration (like integrated development environments, graph visualizers, etc.) where full console support (colorization, pagniation, etc.) is not needed. (BusyBox-backed MinGit, experimental, smaller version of MinGit, using the upstream version of busybox-lean)",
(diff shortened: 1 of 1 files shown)


ddarfantasy:
@HUMORCE May I have a review please? This is rather straightforward. Mostly copied from the spec of mingit.

HUMORCE:
If you don't need to expose `busybox.exe`, execute `scoop shim rm busybox` after installing `main/mingit-busybox`.

ddarfantasy:
> If you don't need to expose `busybox.exe`, execute `scoop shim rm busybox` after installing `main/mingit-busybox`.

Actually I need an available `busybox` executable. But the one comes with the `mingit-busybox` package is not the upstream latest version.

Theoretically, I can remove the `busybox` shim from the `mingit-busybox` and install the `busybox-lean` package. But I don't know if it would cause any conflicts when upgrading one of these.

HUMORCE:
`scoop reset busybox-lean` or `scoop shim alter busybox` after `mingit-busybox` updates.

ddarfantasy:
Thanks for providing a workaround. But I personally think duplicating a manifest is better than manually modifying shims, because `mingit-busybox` is updated rather frequently.

Issue #773, opened 2020-02-02
Fix repology’s 34 reported problems


rasa:
See https://repology.org/repository/scoop/problems

tech189:
I have been going through the outdated packages starting with A and making PRs for all that needed updating. Here is a list of what to ignore:
- 7zip: not outdated, it's just picking out the old beta version
- 7zip-zstd: different formatting of version
- apache-directory-studio: different formatting of version
- apngasm: waiting for https://github.com/apngasm/apngasm/issues/99
- armagetronad: waiting for https://github.com/Calinou/scoop-games/pull/368
- awareness: not outdated, macOS version is higher
- axel: waiting for https://github.com/axel-download-accelerator/axel/pull/313

rashil2000:
Thanks for helping out @tech189, appreciate it very much :)

Repology lists out too many problems, it might be a daunting task 😅

A more tame approach would be: pick the latest [excavator run](https://github.com/ScoopInstaller/Main/actions/workflows/excavator.yml), search through the log for 403 or 404 errors, and have a go at fixing them.

KaranKad:
Current [axel](https://github.com/ScoopInstaller/Main/blob/master/bucket/axel.json) version in scoop is outdated. I've uploaded latest version of axel at https://github.com/KaranKad/axel (took it from MSYS2). So, till official axel binary for windows is added at https://github.com/axel-download-accelerator/axel will it be all right to update to https://github.com/KaranKad/axel? Also its 64 bit only. I'll create a PR for it if you guys approve it.

rashil2000:
We try not to use third-party downloads for packages. Since Scoop does not maintain packages on its own, source of downloads needs to be direct for trust purposes.

ismaell:
> Current [axel](https://github.com/ScoopInstaller/Main/blob/master/bucket/axel.json) version in scoop is outdated. I've uploaded latest version of axel at https://github.com/KaranKad/axel (took it from MSYS2). So, till official axel binary for windows is added at https://github.com/axel-download-accelerator/axel will it be all right to update to https://github.com/KaranKad/axel? Also its 64 bit only. I'll create a PR for it if you guys approve it.

There's no official windows builds of Axel at the moment.

rashil2000:
Any plans to include them?

ismaell:
> Any plans to include them?

No plan; I won't mind to maintain it but I'm not inclined to go out of my way to add support either.

PR #6970, opened 2025-07-16
balena-cli: Fix and update to version 22.1.4

jack-mil:
Fixes 404 errors in CI Excavator.

Apparently they have completely changed the distribution format and contents. A standalone executable is no longer provided, instead a cmd/bash script uses a redistributable node executable to execute the JS/TS app. Upstream recommends the installer for Windows.

Seems like no one has noticed this for 2 months, maybe should be removed from scoop?

Relates to #773

### DIFF
--- a/bucket/balena-cli.json
+++ b/bucket/balena-cli.json
@@ -1,16 +1,16 @@
 {
-    "version": "22.1.2",
+    "version": "22.1.4",
     "description": "Official balena CLI for interacting with balenaCloud and balena API.",
     "homepage": "https://github.com/balena-io/balena-cli",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/balena-io/balena-cli/releases/download/v22.1.2/balena-cli-v22.1.2-windows-x64-standalone.tar.gz",
-            "hash": "c2c82f85dd9046b2277e6def30b30b52c96e9701d72667a0338a82ce3a476998"
+            "url": "https://github.com/balena-io/balena-cli/releases/download/v22.1.4/balena-cli-v22.1.4-windows-x64-standalone.tar.gz",
+            "hash": "8ec9bdbfaf34f179015297fcbd397d89eac220e1cdada117f273dabc18e9e440"
         }
     },
-    "extract_dir":"balena",
-    "env_add_path": "bin",
+    "extract_dir": "balena",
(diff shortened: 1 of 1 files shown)


bucket/balena-cli.json
```diff
@@ -1,21 +1,21 @@
...
-            "url": "https://github.com/balena-io/balena-cli/releases/download/v21.1.14/balena-cli-v21.1.14-windows-x64-standalone.zip",
-            "hash": "fecd5a9374611d0c5e3477b46e480b1b4b3ab0781583acc42ed8368b96608d0b"
+            "url": "https://github.com/balena-io/balena-cli/releases/download/v22.1.2/balena-cli-v22.1.2-windows-x64-standalone.tar.gz",
+            "hash": "c2c82f85dd9046b2277e6def30b30b52c96e9701d72667a0338a82ce3a476998"
         }
     },
-    "extract_dir": "balena-cli",
-    "bin": "balena.exe",
+    "extract_dir":"balena",
+    "env_add_path": "bin",
```
HUMORCE: add `bin\balena.cmd` as shim, dir `bin` contain a `node.exe` which will override node shim

HUMORCE (changes requested):
shim instead env

jack-mil:
Their app doesn't work as a shim. It uses relative paths or something. Only way I could make it work was by adding to path (according to their docs). There is only a bash script `balena.cmd` that executes the node.exe with the app entry point (JavaScript).

I agree that having node.exe in that folder would break other installations of node. Maybe this app can't be supported by scoop anymore.

HUMORCE:
I don't know exactly how it works, but shim seems to not affect node environment.


```
PS C:\Users\WDAGUtilityAccount> Get-Command balena | Select-Object Source

Source
------
C:\Users\WDAGUtilityAccount\scoop\shims\balena.cmd


PS C:\Users\WDAGUtilityAccount> balena login
 _            _
| |__   __ _ | |  ____  _ __    __ _
| '_ \ / _` || | / __ \| '_ \  / _` |
| |_) | (_) || ||  ___/| | | || (_) |
|_.__/ \__,_||_| \____/|_| |_| \__,_|


Logging in to balena-cloud.com
? How would you like to login? (Use arrow keys)
> Web authorization (recommended)
  Credentials
  Authentication token
  I don't have a balena account!
```

PR #7028, opened 2025-08-04
Mill: Update to 1.0.2

nightscape:
- [ ] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

### DIFF
--- a/bucket/mill.json
+++ b/bucket/mill.json
@@ -1,18 +1,18 @@
 {
-    "version": "0.12.5",
+    "version": "1.0.2",
     "description": "Build tool aiming for Java and Scala.",
     "homepage": "https://mill-build.com/",
     "license": "MIT",
     "suggest": {
         "JDK": "java/openjdk"
     },
-    "url": "https://github.com/com-lihaoyi/mill/releases/download/0.12.5/0.12.5-assembly#/mill.bat",
-    "hash": "0c7b25412feecf06d955d013418de9a9080ce755bedbac7601c9109c33d5a057",
-    "bin": "mill.bat",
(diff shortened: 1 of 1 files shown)


bucket/mill.json
```diff
@@ -1,18 +1,18 @@
 {
-    "version": "0.12.5",
+    "version": "1.0.2",
     "description": "Build tool aiming for Java and Scala.",
     "homepage": "https://mill-build.com/",
     "license": "MIT",
     "suggest": {
         "JDK": "java/openjdk"
```
Lutra-Fs, line 7: I would suggest removing OpenJDK deps as we have migrated to the native runner

bucket/mill.json
```diff
@@ -1,18 +1,18 @@
...
     },
-    "url": "https://github.com/com-lihaoyi/mill/releases/download/0.12.5/0.12.5-assembly#/mill.bat",
-    "hash": "0c7b25412feecf06d955d013418de9a9080ce755bedbac7601c9109c33d5a057",
-    "bin": "mill.bat",
+    "url": "https://repo1.maven.org/maven2/com/lihaoyi/mill-dist-native-windows-amd64/1.0.2/mill-dist-native-windows-amd64-1.0.2.exe#/mill.exe",
```
Lutra-Fs, line 9: We should use 
"architecture" field to limit the native runner only run with AMD64(64 bit) machine. Because https://mill-build.org/mill/cli/installation-ide.html#_mill_native_and_jvm_executables clearly states that the native one does not support ARM64 architecture

Issue #7067, opened 2025-08-16
[Bug]: dart@3.8.0 has not been updated since May 2025


nureyev:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

dart

### Expected/Current Behaviour

Dart 3.8.0 onwards dropped ia32 (32-bit) releases. This may be stopping the manifest from updating.

### Steps to Reproduce

```console
The manifest is at 3.8.0 while Dart current release is at 3.9.0 with 4 updates missed.
```

### Possible Solution

Remove the 32bit:{} sections from the manifest.

### Scoop and Buckets Version

```console
Current Scoop version:
b588a06e (HEAD -> master, origin/master, origin/HEAD) chore(release): Bump to version 0.5.3 (resync) (#6436)

'main' bucket:
eb0bec458 (HEAD -> master, origin/master, origin/HEAD) edgedriver: Update to version 139.0.3405.102
```

### Scoop Config

```console
last_update                       scoop_branch scoop_repo
-----------                       ------------ ----------
2025-08-17T00:08:13.3272874+02:00 master       https://github.com/lukesampson/scoop
```

### PowerShell Version

```console
-
```

Lutra-Fs:
Hi @nureyev. Would you be happy to raise a PR for that?

nureyev:
@Lutra-Fs The updated manifest is failing on the pull request checks (Checkver). I'm wondering if the issue is something else and not the 32bit sections?

PR #7070, opened 2025-08-17
dart@3.8.0: Remove 32bit installer sections

nureyev:
The manifest was failing to update because Dart removed ia32 (32-bit) packages from 3.8.0 onwards.

Relates to #7067.


Lutra-Fs:
> 3.9.0 (scoop version is 3.8.0)
> autoupdate available

You just need to submit the newest version :)

nureyev:
@Lutra-Fs All checks passed. Thanks for your help

Lutra-Fs:
Thank you for your help! Would you be able to create a `dart38` manifest for the versions bucket, so we can continue to support 32-bit compatibility?

Lutra-Fs (approved):

PR #7264, opened 2025-10-16
corepack: Add version 0.34.0

brian6932:
Node JS 25.0.0 removed built in corepack, this is an alternative to having to to install it with npm.

### DIFF
--- a/bucket/corepack.json
+++ b/bucket/corepack.json
@@ -1,0 +1,29 @@
+{
+    "version": "0.34.0",
+    "description": "Zero-runtime-dependency package acting as bridge between Node projects and their package managers",
+    "homepage": "https://github.com/nodejs/corepack",
+    "license": "MIT",
+    "suggest": {
+        "nodejs": "main/nodejs"
+    },
+    "url": "https://github.com/nodejs/corepack/releases/download/v0.34.0/corepack.tgz",
+    "hash": "0c85699ec30b495273df7f8391235135d9aa43297265f03cf8c8005a17fa693f",
+    "extract_dir": "package",
+    "env_add_path": "shims/nodewin",
+    "post_install": [
+        "Get-ChildItem \"$dir/shims/nodewin\" | ForEach-Object {",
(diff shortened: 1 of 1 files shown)


z-Fng:
> Node JS 25.0.0 removed built in corepack, this is an alternative to having to to install it with npm.

Packages reliant on an external Node.js runtime should not be distributed via Scoop manifests; use native package managers instead.

brian6932:
> Packages reliant on an external Node.js runtime should not be distributed via Scoop manifests; use native package managers instead.

This one's somewhat of an exception to this, it was part of the actual package previously, so it's expected to be in the main repo Imo.

>use native package managers instead.

Installing via npm means that the tool you're using to manage the language's package manager, requires using a package manager from the language, it makes no sense. This is supposed to let you choose which package manager to install, use, manage, and update...

z-Fng:
> https://github.com/nodejs/corepack/blob/d4dcb1f89741603e776bba9d457425750fa26987/README.md?plain=1#L38-L40
We do acknowledge the irony and overhead of using npm to install Corepack, which
is at least part of why the preferred option is to use the Corepack version that
is distributed along with Node.js itself.

I understand the paradox of using npm to install a package manager manager, but I don't this belong in Scoop for two key reasons:
1. Upstream decisions belong to upstream: The choice to unbundle Corepack and distribute it via npm was made by Node.js. Scoop's role is to package software as provided by upstream. 
2. Scoop is not a replacement for native package managers: Scoop distributes standalone binaries and runtimes, not language-specific scripts that tightly couple to a Node environment. Just as we don't duplicate pip packages, tools that require the Node runtime belong in npm, not Scoop. And accepting a package that depends on Node.js would only invite more requests of the same kind.

brian6932:
Yes but it doesn't make any sense, for example in corepack you can disable npm, then how the hell would I update it? Personally, I don't like or use npm, I exclusively use pnpm, and don't want npm even enabled. I get your decision, and have been using this in my own personal bucket for a long time, but I do think it should be in main.

PR #7319, opened 2025-11-03
yt-dlp: Add `js-runtime` to `suggest`

### LINKED ISSUE #7359: Needs clarification over yt-dlp's updated requirements
https://github.com/yt-dlp/yt-dlp/wiki/EJS

So yeah this isn't quite news already. However it's ambiguous how Scoop users should best handle it.

What's the recommended way? Just `scoop install deno` and that's it, or something else?

It's best to have it mentioned in the manifest, especially "suggest" and "notes" field, etc.

candrapersada:
suggest add `deno` as dependency `yt-dlp`

Closes #7359 

### DIFF
--- a/bucket/yt-dlp.json
+++ b/bucket/yt-dlp.json
@@ -5,7 +5,6 @@
     "notes": [
         "Requires a JavaScript runtime for full functionality (e.g., deno, node, bun).",
         "By default yt-dlp prefers 'deno' — use `--js-runtimes <runtime>` as an argument, or edit `yt-dlp.conf` to override."
-
     ],
     "license": "Unlicense",
     "suggest": {


Vinfall:
Deno is 64-bit only, 32-bit users have to use nodejs, which is not enabled in `yt-dlp` by default.

I think it would be better if you could add the above line to `notes` field.

The easiest way to enable nodejs support is [Alpine's approach](https://gitlab.alpinelinux.org/alpine/aports/-/commit/ee127274a99528fcc0ed9b1658d3a25693c74d20?#c8b6ed604b4b34aa5f3387959a921673f06c94ca) but I'm not sure how to integrate that into scoop, probably via postinstall script. Otherwise you should notify 32-bit nodejs users that they have to use `yt-dlp --js-runtimes node` everytime they want to download YouTube videos.

candrapersada:
If use node add `args = --js-runtimes node` in yt-dlp.shim 32-bit

candrapersada:
```
        "32bit": {
            "bin": [
                [
                    "yt-dlp.exe",
                    "yt-dlp",
                    "--no-js-runtimes --js-runtimes node"
                ]
```

bucket/yt-dlp.json
```diff
@@ -4,6 +4,7 @@
     "homepage": "https://github.com/yt-dlp/yt-dlp",
     "license": "Unlicense",
     "suggest": {
+        "deno": "deno",
```
Lutra-Fs:
```suggestion
        "js-runtime": [
            "deno",
        	"nodejs",
            "nodejs-lts",
            "versions/nodejs22"
            "bun"
        ],
```

`nodejs22` are the latest `nodejs` manifest that includes 32-bit support, located in `versions` bucket.

Lutra-Fs:
> ```
>         "32bit": {
>             "bin": [
>                 [
>                     "yt-dlp.exe",
>                     "yt-dlp",
>                     "--no-js-runtimes --js-runtimes node"
>                 ]
> ```

I think adding a note is sufficient here, since 32-bit users can enable YouTube support by opting in to `quickjs` or `versions/nodejs22`. There’s no need for additional changes to update the shim for 32bit user specifically.

Lutra-Fs:
I'll submit a manifest PR for [`quickjs`](https://bellard.org/quickjs/) shortly, allowing us to recommend `quickjs` for 32-bit users as well.

For reference, yt-dlp has the following requirements:
- **Minimum supported Node.js version:** 20.0.0
- **Strongly recommended Node.js version:** 25+ (for security reasons)

But for 32bit users, they can only install `nodejs22` in `versions` bucket. So `quickjs` might be a viable option here

candrapersada:
but by default it uses deno or none if without enable `--js-runtime` even though it uses 32-bit

bucket/yt-dlp.json
```diff
@@ -2,12 +2,17 @@
     "version": "2025.11.12",
     "description": "A youtube-dl fork with additional features and fixes.",
     "homepage": "https://github.com/yt-dlp/yt-dlp",
+    "notes": [
+        "Requires a JavaScript runtime for full functionality (e.g., deno, node, bun).",
+        "By default yt-dlp prefers 'deno' — use `--js-runtimes <runtime>` to override."
```
candrapersada:
```suggestion
        "By default yt-dlp prefers 'deno' — use `--js-runtimes <runtime>` as an argument, or edit `yt-dlp.conf` to override."

```

Lutra-Fs: I don't think this addition is needed, since using long-style arguments in bash (like `--js-runtimes`) is already self-explanatory for users familiar with the command line. The requirement and usage are clear without further suggestion or documentation tweaks.

Lutra-Fs (approved):

bucket/yt-dlp.json
```diff
@@ -2,8 +2,20 @@
...
+
+    ],
     "license": "Unlicense",
     "suggest": {
+        "js-runtime": [
+            "deno",
+            "nodejs",
+            "nodejs-lts",
+            "versions/nodejs22",
+            "bun"
```
Lutra-Fs:
```suggestion
            "bun",
            "quickjs"
```

bucket/yt-dlp.json
```diff
@@ -2,8 +2,21 @@
     "version": "2025.12.08",
     "description": "A youtube-dl fork with additional features and fixes.",
     "homepage": "https://github.com/yt-dlp/yt-dlp",
+    "notes": [
+        "Requires a JavaScript runtime for full functionality (e.g., deno, node, bun).",
+        "By default yt-dlp prefers 'deno' — use `--js-runtimes <runtime>` as an argument, or edit `yt-dlp.conf` to override."
+
```
Lutra-Fs:
```suggestion
```

PR #7379, opened 2025-11-18
helm: Fix regex for v4

stevehipwell:
This PR fixes the regex for Helm so it can pick up the latest v4 version. It also adds support for the arm64 version.

### DIFF
--- a/bucket/helm.json
+++ b/bucket/helm.json
@@ -1,24 +1,31 @@
 {
-    "version": "3.19.2",
-    "description": "The package manager for Kubernetes",
+    "version": "4.0.0",
+    "description": "The package manager for Kubernetes.",
     "homepage": "https://helm.sh",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://get.helm.sh/helm-v3.19.2-windows-amd64.zip",
-            "hash": "8b3ea37bcd7c5eda68eadd278690498fb886a17ec541b449216ab3b45748600e",
+            "url": "https://get.helm.sh/helm-v4.0.0-windows-amd64.zip",
+            "hash": "0f9a8c891b8d908a37fbb68f12dea92b633eb29e49070bd650f5760a1a99aa8d",
             "extract_dir": "windows-amd64"
+        },
+        "arm64": {
+            "url": "https://get.helm.sh/helm-v4.0.0-windows-arm64.zip",
+            "hash": "f3ff262427547cc1b1dc3356d587ed8ffaa23f2abf24bc06660a350b9b7925f9",
+            "extract_dir": "windows-arm64"
         }
     },
     "bin": "helm.exe",
     "checkver": {
-        "url": "https://github.com/kubernetes/helm/releases",
-        "regex": "/tag/v(3\\.[\\d.]+)\""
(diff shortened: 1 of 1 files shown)


bucket/helm.json
```diff
@@ -1,24 +1,32 @@
...
+            "hash": "f3ff262427547cc1b1dc3356d587ed8ffaa23f2abf24bc06660a350b9b7925f9",
+            "extract_dir": "windows-arm64"
         }
     },
     "bin": "helm.exe",
     "checkver": {
-        "url": "https://github.com/kubernetes/helm/releases",
-        "regex": "/tag/v(3\\.[\\d.]+)\""
+        "url": "https://github.com/helm/helm/releases",
+        "regex": "\"/helm/helm/releases/tag/v([\\d]+\\.[\\d]+\\.[\\d]+)\""
```
Lutra-Fs: generally speaking, I believe "/helm/helm/releases" is not required to match the release here, match the tag should be enough to get the releases. Simply matching "/tag/v([\\d]+\\.[\\d]+\\.[\\d]+)" should be enough.

stevehipwell: This regex matches the GitHub generated content for the release tag link and makes the likelihood of finding a match in the release description significantly lower. The regex you've provided wouldn't work as it would pick up pre-release versions due to the removal of the `\"` at the end.

Lutra-Fs: Actually it should be simplified to `"github": "https://github.com/helm/helm"` instead of that one since we have GitHub matcher that uses GitHub API. I have checked that the helm team is following the practise to release the RC versions with pre-release tag. That means they won't be matched

Lutra-Fs: I didn't mean to remove the trailing slash, that's not align with what I said and it's an unintentional error.

stevehipwell: @Lutra-Fs we can't use the GitHub matcher as the binary isn't coming from the GitHub release, which is why we need to use the releases page and a regex.

Lutra-Fs: That's not true. GitHub matcher in this case is only used to get the latest release's tag that is not marked as pre release. It will then pass the version variable to the autoupdate part so autoupdate can construct the url correctly.

stevehipwell: Thanks; that's what I was originally attempting to do, but it looks like I made a typo and then made an incorrect assumption.

AzureMarker:
@Lutra-Fs FYI this triggered the bug https://github.com/ScoopInstaller/Scoop/issues/5582 for us because `arm64` was added. More context: https://github.com/helm/helm/issues/31532

Lutra-Fs:
> @Lutra-Fs FYI this triggered the bug https://github.com/ScoopInstaller/Scoop/issues/5582 for us because `arm64` was added. More context: https://github.com/helm/helm/issues/31532

Please use the install via git history feature available in a custom branch. Find it in the Scoop issues page. It should be pinned. The history version installation via autoupdate is maintained in best effort and does not guarantee its reliability. Thank you.

AzureMarker:
@Lutra-Fs There is also an open PR to fix this specific issue: https://github.com/ScoopInstaller/Scoop/pull/6325

PR #7380, opened 2025-11-18
chore: Update compatible checkver to github

stevehipwell:
This PR updates manifests to the GitHub checkver pattern where possible. This is based on the discussion on #7379.

CC @Lutra-Fs 

### DIFF
--- a/bucket/k0s.json
+++ b/bucket/k0s.json
@@ -11,13 +11,17 @@
     },
     "bin": "k0s.exe",
     "checkver": {
-        "url": "https://github.com/k0sproject/k0s/releases",
-        "regex": "v([\\d.]+\\+k0s[\\d.]+)"
+        "github": "https://github.com/k0sproject/k0s",
+        "regex": "v(\\d+\\.\\d+\\.\\d+\\+k0s\\.\\d+)"
     },
     "autoupdate": {
         "architecture": {
             "64bit": {
-                "url": "https://github.com/k0sproject/k0s/releases/download/v$version/k0s-v$version-amd64.exe#/k0s.exe"
+                "url": "https://github.com/k0sproject/k0s/releases/download/v$version/k0s-v$version-amd64.exe#/k0s.exe",
+                "hash": {
+                    "url": "$baseurl/sha256sums.txt",
+                    "regex": "$sha256\\s\\*k0s-v$version-amd64.exe"
+                }
             }
         }
     }
--- a/bucket/atomicparsley.json
+++ b/bucket/atomicparsley.json
@@ -15,8 +15,8 @@
     },
     "bin": "AtomicParsley.exe",
     "checkver": {
-        "url": "https://github.com/wez/atomicparsley/releases",
-        "regex": "tree\\/([\\d.[a-f]+)\""
+        "github": "https://github.com/wez/atomicparsley",
+        "regex": "/releases/tag/(\\d+\\.\\d+\\.[a-f0-9]+)"
     },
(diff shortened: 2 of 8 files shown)


Lutra-Fs:
That's quite a lot of changes, so I'll need some extra time to review and test everything thoroughly. ⏳

Lutra-Fs:
Just reverted my comment since it does not work. Github API is only used if we do not supply extra parameters like regex, so we cannot use jsonpath.

stevehipwell:
> Just reverted my comment since it does not work. Github API is only used if we do not supply extra parameters like regex, so we cannot use jsonpath.

@Lutra-Fs I know this has been discussed further in https://github.com/ScoopInstaller/Scoop/issues/6559 but is this the current behaviour?

Lutra-Fs:
No, it will use GitHub API when `gh_token` is set in scoop config

stevehipwell:
> No, it will use GitHub API when gh_token is set in scoop config

@Lutra-Fs does that not mean that the regex needs to work in both contexts?

Lutra-Fs:
I think "/releases/tag/" or "tag/" match are working well on both context. So maybe we'd better keep better using that unless we modified the GitHub checkver behaviour.

Lutra-Fs:
One issue we previously discussed in the core repository is that when checkver falls back to HTML parsing, there's currently no logging to inform users why this happens, such as not having correctly set up their `gh_token` or encountering another issue. If we can make the reason for falling back clearly visible in the logs, I think modifying this behaviour is acceptable.

Additionally, I believe using the GitHub API by default is more intuitive than HTML parsing. This is technical debt we need to address. Since the GitHub API provides anonymous access (albeit with a lower rate limit), it should work fine for normal users even if they don't set a `gh_token`. Also, our autoupdate excavator is already using GitHub API for a long time. So I guess it might be the time to switch to the API completely.

bucket/k0s.json
```diff
@@ -11,13 +11,17 @@
     },
     "bin": "k0s.exe",
     "checkver": {
-        "url": "https://github.com/k0sproject/k0s/releases",
-        "regex": "v([\\d.]+\\+k0s[\\d.]+)"
+        "github": "https://github.com/k0sproject/k0s",
+        "regex": "v(\\d+\\.\\d+\\.\\d+\\+k0s\\.\\d+)"
```
Lutra-Fs, line 15: Should we also use 
```suggestion
        "regex": "/releases/tag/v(\\d+\\.\\d+\\.\\d+\\+k0s\\.\\d+)"
```
here?

stevehipwell: This doesn't work as in the API mode the `+` is URL escaped in the only part of the payload where the tag is listed with the `/releases/tag/` prefix.

stevehipwell: @Lutra-Fs see https://github.com/ScoopInstaller/Scoop/issues/6559#issuecomment-3575532053.

Lutra-Fs:
I think overall LGTM, we can merge it now and find a better solution in https://github.com/ScoopInstaller/Scoop/issues/6559

stevehipwell:
@Lutra-Fs I've rebased to fix the failing checkver due to the k0s version having been updated since I created the PR. Are we good to get this merged?

Lutra-Fs:
Hi steve,

Yeah I am really busy today and I will merge it tmr or you can merge yourself.

Regards

在2025年11月27日周四 下午8:31，Steve Hipwell写道：

PR #7406, opened 2025-11-30
cuda: Keep CUDA Visual Studio Integration files

rhjdvsgsgks:

to solve "No CUDA toolset found" error

ref: https://stackoverflow.com/questions/56636714/cuda-compile-problems-on-windows-cmake-error-no-cuda-toolset-found/77085810#77085810

### DIFF
--- a/bucket/cuda.json
+++ b/bucket/cuda.json
@@ -6,6 +6,10 @@
         "identifier": "Freeware",
         "url": "https://docs.nvidia.com/cuda/eula/index.html"
     },
+    "notes": [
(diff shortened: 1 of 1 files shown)


Lutra-Fs:
Hi @rhjdvsgsgks, thanks for your contribution. I tested your PR with Visual Studio Build Tools and CMake in WSB. I found that `-DCMAKE_GENERATOR_TOOLSET` still needs to be explicitly set when configuring CMake projects.

Can you a note to the manifest mentioning that users may need to set `-DCMAKE_GENERATOR_TOOLSET` when using CMake with CUDA?

PR #7440, opened 2025-12-14
ddev@1.24.10: Expose bundled utilities and add HTTPS setup notes

Yi-pixel:

### Fixed

```text
$ ddev start
Starting laravel-12...
The hostname laravel-12.ddev.site is not currently resolvable, trying to add it to the hosts file
Unable to run ddev-hostname.exe, please check it; err=exec: "ddev-hostname.exe": executable file not found in %PATH%; output=
Failed to start laravel-12: Binary not found: ddev-hostname is not installed, please install it, see https://docs.ddev.com/en/stable/users/usage/commands/#hostname
```

### Summary

This PR updates the `ddev` Scoop manifest to better reflect the contents of the official Windows release and improve the user onboarding experience.

### Changes

* Expose additional bundled executables:

  * `mkcert.exe`
  * `ddev-hostname.exe`
* Add post-installation notes to guide users through HTTPS setup using `mkcert`.

### Rationale

The official Windows release archive already includes `mkcert.exe` and `ddev-hostname.exe`, but they were not previously exposed via the `bin` section of the Scoop manifest. As a result, users could not directly invoke these utilities after installation.

Additionally, HTTPS support in DDEV on Windows requires installing a custom root certificate. Adding explicit notes helps prevent confusion and reduces setup friction for new users.

### User Impact

After this change:

* Users can directly run `mkcert` and `ddev-hostname` from the command line.
* Users receive clear guidance on enabling HTTPS by installing the local root certificate.

### DIFF
--- a/bucket/ddev.json
+++ b/bucket/ddev.json
@@ -19,7 +19,6 @@
     },
     "bin": [
         "ddev.exe",
-        "mkcert.exe",
         "ddev-hostname.exe"
     ],
     "checkver": {


bucket/ddev.json
```diff
@@ -3,6 +3,13 @@
     "description": "An open source tool that makes it simple to get local PHP development environments up and running in minutes",
     "homepage": "https://ddev.readthedocs.io/en/latest/",
     "license": "Apache-2.0",
+    "notes": [
+        "If you need to use HTTPS, you must install a custom root certificate.",
+        "Please run the following Command: ",
+        "",
+        "> mkcert -install",
+        ""
```
Lutra-Fs: Hi can you have a look at https://github.com/ScoopInstaller/Extras/blob/master/bucket/wireshark.json and update the note field to match its style?

Yi-pixel: Hello, I’ve made an update. However, should I also consider removing `mkcert.exe` from the `bin`?

Lutra-Fs: If `mkcert` is only needed once during the lifetime of the installation (e.g. just to run `mkcert -install`), I don't think we need to add a shim for it.  
We can instead just document the full path to the bundled `mkcert` binary in the notes section, so users can run it manually when needed. It will also help us avoid shim conflict with `extras/mkcert`

Yi-pixel: Ok, removed, ready now.

Issue #7394, opened 2025-11-23
[Bug] Various packages stopped updating due to checkver errors


Lutra-Fs:
This issue is a sister issue to https://github.com/ScoopInstaller/Extras/issues/16379.

There are some errors during excavator execution. Various packages stopped updating due to checkver errors.
See: https://github.com/ScoopInstaller/Main/actions/runs/19605894658/job/56144713136

Contributors are warmly welcome to help fix one or more of these broken manifest. Please note:
- Make sure there are no existing fix PRs.
- Create a separate PR or a separate commit for each manifest, with a title following the naming rules.
- Please do not attempt to close this issue in your PR.
- For manifests that are still available but will no longer receive updates, remove their checkver fields instead of deprecating them.
- For autoupdate filed, if we can still install a specific version via `scoop install app@version` (i.e., autoupdate is version-specific), removing it is also not recommended. https://github.com/ScoopInstaller/Extras/pull/16392#issuecomment-3438955656
> IMO, if we keep the autoupdate field, users can still install a specific version via `scoop install app@version`. Otherwise:
> ```
> PS> scoop install fbflipper@0.238.0
> WARN  Given version (0.238.0) does not match manifest (0.239.0)
> WARN  Attempting to generate manifest for 'fbflipper' (0.238.0)
> 'fbflipper' does not have autoupdate capability
> ```
> P.S. If https://github.com/ScoopInstaller/Scoop/pull/6370 gets merged, this will no longer be necessary and can be removed later.

### Package List
Here are all the packages stopped updating due to checkver errors.

The package list was matched by lines containing "ERROR Could not update" in the log.

- [ ] capstone
- [ ] cockroachdb
- [ ] cloak
- [ ] cppcryptfsctl
- [ ] dioxus
- [ ] dotcover-clt
- [ ] docker
- [ ] dvc
- [ ] ffsend
- [ ] espeak-ng
- [ ] git-crypt
- [ ] gource
- [ ] gum
- [ ] jd-cli
- [ ] mailsend-go
- [ ] loft
- [ ] minio
- [ ] mill
- [ ] odbc2parquet
- [ ] nixpacks
- [ ] ossutil
- [ ] ossgadget
- [ ] postgrest
- [ ] psake
- [ ] prs
- [ ] q-dns
- [ ] qrcp
- [ ] sacad
- [ ] rustic
- [ ] qtifw
- [ ] gnutls
- [ ] pomsky
- [ ] silicon
- [ ] serverless
[...]

PR #7447, opened 2025-12-19
yara: Update to version 4.5.5, fix checkver & autoupdate

3xp0rt:
Relates to https://github.com/ScoopInstaller/Main/issues/7394

### DIFF
--- a/bucket/yara.json
+++ b/bucket/yara.json
@@ -37,8 +37,7 @@
         }
     },
     "checkver": {
-        "url": "https://api.github.com/repositories/7037996/releases",
-        "jsonpath": "$..browser_download_url",
+        "github": "https://github.com/VirusTotal/yara",
         "regex": "yara-(?<version>\\d+(?:\\.\\d+)+)-(?<build>\\d+)-win(?:32|64)\\.zip"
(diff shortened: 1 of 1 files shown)


bucket/yara.json
```diff
@@ -37,15 +37,16 @@
         }
     },
     "checkver": {
-        "github": "https://github.com/VirusTotal/yara"
+        "url": "https://api.github.com/repos/VirusTotal/yara/releases/latest",
```
Lutra-Fs: Can you try using `"github": "https://github.com/VirusTotal/yara"`, with a custom regex?

3xp0rt: It doesn’t work because the response from this URL doesn’t include the build number.

z-Fng: > It doesn’t work because the response from this URL doesn’t include the build number.

It works when GH_TOKEN is configured. See: https://github.com/ScoopInstaller/Scoop/issues/6559#issuecomment-3567131614

> The behavior when no GitHub token is configured is inconsistent with when a token is set, and there's not even any related logs. This is very counterintuitive for contributors.

PR #7464, opened 2025-12-28
ffmpeg-shared: set environment for developer

MiyakoMeow:

### DIFF
--- a/bucket/ffmpeg-shared.json
+++ b/bucket/ffmpeg-shared.json
@@ -10,20 +10,11 @@
             "extract_dir": "ffmpeg-8.0.1-full_build-shared"
         }
     },
-    "post_install": [
-        "# Workaround for https://github.com/ScoopInstaller/Main/issues/2611",
-        "Remove-Item \"$scoopdir\\shims\\ffmpeg.ps1\", \"$scoopdir\\shims\\ffplay.ps1\", \"$scoopdir\\shims\\ffprobe.ps1\" -ErrorAction SilentlyContinue"
-    ],
     "persist": "fonts",
     "env_set": {
         "FFMPEG_DIR": "$dir"
     },
     "env_add_path": "bin",
-    "bin": [
-        "bin\\ffmpeg.exe",
-        "bin\\ffplay.exe",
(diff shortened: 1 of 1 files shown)


bucket/ffmpeg-shared.json
Lutra-Fs: `bin` creates shims for specific executables, while `env_add_path` exposes every executable in that directory by adding it to `PATH`.
In this case, adding `bin` to `env_add_path` already makes `ffmpeg.exe`, `ffplay.exe` and `ffprobe.exe` available, so listing them again in `bin` seems redundant.  
Could you share the reason for switching from `bin` to `env_add_path` here?

brightmeows: I'm a developer, using `ffmpeg-next` crate, [See this](https://github.com/zmwangx/rust-ffmpeg/wiki/Notes-on-building#msvc-toolchain).
Both `FFMPEG_PATH` and the `Path` entry is for libraries like above. I also see `llvm` has set `LIBCLANG_PATH`, and would like to act like that.

Lutra-Fs: I understand you want to expose the libraries for development use, similar to how `LIBCLANG_PATH` is used. However, if the `bin` directory is already added via `env_add_path`, then `ffmpeg.exe`, `ffplay.exe`, and `ffprobe.exe` are available on `PATH` from that directory, so creating shims for them in `bin` is redundant, which are L22-26. 

If the goal is to provide additional variables like `FFMPEG_PATH` for libraries, then `env_add_path` with `env_set` might be more suitable and the explicit `bin` entries for those executables can be removed.

bucket/ffmpeg-shared.json
```diff
@@ -15,6 +15,10 @@
         "Remove-Item \"$scoopdir\\shims\\ffmpeg.ps1\", \"$scoopdir\\shims\\ffplay.ps1\", \"$scoopdir\\shims\\ffprobe.ps1\" -ErrorAction SilentlyContinue"
```
Lutra-Fs: Since we are changing to `env_add_path`, custom handling of shims should also be removed, if applicable.

PR #7476, opened 2026-01-03
7zip: Refactor context-menu handling

SorYoshino:
### Summary

Refines the `7zip` manifest to improve accuracy, reliability of context-menu registration, and clarity of metadata, while aligning the structure with current Scoop best practices.

### Related issues or pull requests

- Relates to #7475 
- Relates to https://github.com/ScoopInstaller/Extras/pull/16918

### Changes

- Replace simple `license` string with structured SPDX identifiers and license URL
- Revise `notes` entry to provide clearer instructions for context-menu registration
- Reworke `post_install` to:
  - unify handling of `.reg` templates
  - support both per-user (HKCU) and global (HKLM) contexts via `$global`
  - replace `{{7zip_dir}}` placeholder in templates
  - ensure correct escaping and encoding
- Add Start Menu shortcuts grouping under **7-Zip** folder  
- Add shortcut to `7-zip.chm` help file  
- Add `uninstaller.script` to import `uninstall-context.reg` automatically on uninstall

### Notes

```reg
; Effective only when configured under the HKEY_CURRENT_USER hive.

[HKEY_CURRENT_USER\Software\7-Zip\Options]
"CascadedMenu"=dword:00000001
"MenuIcons"=dword:00000001

```

### Testing

<details>

<summary>The test results are as follows:</summary> <br>

```powershell
┏[ ~]
└─> scoop install Unofficial/7zip
Installing '7zip' (25.01) [64bit] from 'Unofficial' bucket
Loading 7z2501-x64.msi from cache.
Checking hash of 7z2501-x64.msi ... ok.
Extracting 7z2501-x64.msi ... done.
Linking D:\Software\Scoop\Local\apps\7zip\current => D:\Software\Scoop\Local\apps\7zip\25.01
Creating shim for '7z'.
Creating shim for '7zG'.
Making D:\Software\Scoop\Local\shims\7zg.exe a GUI binary.
Creating shim for '7zFM'.
Making D:\Software\Scoop\Local\shims\7zfm.exe a GUI binary.
Creating shortcut for 7-Zip\7-Zip File Manager (7zFM.exe)
Creating shortcut for 7-Zip\7-Zip Help (7-zip.chm)
Persisting Codecs
Persisting Formats
Running post_install script...done.
'7zip' (25.01) was installed successfully!
Notes
-----
To register the context menu entry, please execute the following command:
reg import "D:\Software\Scoop\Local\apps\7zip\current\install-context.reg"
```

</details>


bucket/7zip.json
```diff
@@ -30,31 +33,39 @@
...
         ]
     ],
     "persist": [
         "Codecs",
         "Formats"
     ],
+    "uninstaller": {
+        "script": [
+            "if ($cmd -eq 'uninstall') { reg import \"$dir\\uninstall-context.reg\" *> $null }",
+            "Stop-Process -Name 'explorer' -Force -ErrorAction SilentlyContinue",
```
Lutra-Fs: Should we always uninstall the association and stop the explorer?

SorYoshino: > uninstall the association

The context menu is unregistered only when `$cmd` is set to `uninstall`. The logic may need to be adjusted when `no_junction` is `true`, but since most manifests do not account for this case, I do not believe a change is necessary at this time.

> stop the explorer

Because 7-Zip uses CLSIDs when registering context menu entries, updates or uninstalls may fail due to files being in use if `explorer` is not stopped. As most users who install 7-Zip are likely to enable the context menu integration, I believe it is unnecessary to explicitly check whether the context menu is registered, and stopping `explorer` directly is sufficient.

SorYoshino: Additionally, if the user has registered file associations, uninstalling via Scoop does not currently remove the associated settings. Should `.reg` files be provided to manage and clean up file associations?

z-Fng: > https://github.com/ScoopInstaller/Extras/pull/17011#discussion_r2699642384
It is safe to use Stop-Process/Stop-Service in the uninstaller script (or afterwards) to stop background processes/services, as the checks for running processes and the [ignore_running_processes](https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/libexec/scoop-config.ps1#L101-L104) configuration have already been performed. At that stage, either no related process is running, or the user has explicitly configured it to ignore running processes.

I think I'm referring to the programs in the installation directory.

https://github.com/ScoopInstaller/Extras/pull/10476#issuecomment-1712477541
> Maybe not for everyone, but I cannot accept something like stop-process -name explorer. For scripts that destroy the workspace, I think it should be executed manually by users.

SorYoshino: > I think I'm referring to the programs in the installation directory.

I am aware of that, but requiring users to manually run `Stop-Process -Name 'explorer'` to stop Explorer every time they update or uninstall seems unnecessarily cumbersome.

This becomes especially problematic when multiple applications that require manually stopping Explorer are updated at the same time. In such cases, the impact can be significant, as users may have to stop Explorer multiple times manually to complete all updates successfully.

> For scripts that destroy the workspace, I think it should be executed manually by users.

I understand the considerations behind this, but personally I would prefer the program to handle these operations automatically.

As a compromise, I can add logic to the script to detect whether the user has registered the context menu entries, and implement behavior that only stops Explorer automatically if those entries are registered. I will also update the relevant content in the `notes` field to clarify this behavior.

```
Notes
-----
To register file associations, please execute the following command:
- `reg import "D:\Software\Scoop\Global\apps\7-Zip\current\register-file-associations.reg"`
To register the context menu entry, please execute the following command:
- `reg import "D:\Software\Scoop\Global\apps\7-Zip\current\register-context-menu.reg"`
Please note that if the context menu entry is registered, Windows Explorer will be automatically restarted during future updates.
-----
```

z-Fng: > I am aware of that, but requiring users to manually run `Stop-Process -Name 'explorer'` to stop Explorer every time they update or uninstall seems unnecessarily cumbersome.

File deletion failures seem to only occur during uninstallation. Reminding users to restart Explorer manually should be enough. Since the uninstaller doesn't kill the Explorer process itself, it usually just flags those files for deletion on the next reboot anyway (e.g., MOVEFILE_DELAY_UNTIL_REBOOT).

z-Fng (approved):
Thanks for your contribution!

PR #7527, opened 2026-01-13
proto: Add version 0.54.1

lvyuemeng:
Closes: #7526

## Summary by CodeRabbit

* **Chores**
  * Added a package manifest for Proto v0.54.1 enabling 64‑bit Windows distribution.
  * Provides installer/uninstaller with PROTO_HOME and PATH/shim handling.
  * Adds upstream version check and auto-update rules for 64‑bit builds, including integrity (SHA256) verification.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/bucket/proto.json
+++ b/bucket/proto.json
@@ -1,0 +1,51 @@
+{
+    "version": "0.54.1",
+    "description": "Proto is a pluggable version manager, a unified toolchain.",
+    "homepage": "https://moonrepo.dev/proto",
+    "license": "MIT",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/moonrepo/proto/releases/download/v0.54.1/proto_cli-x86_64-pc-windows-msvc.zip",
+            "hash": "77e5cf3b6ee89882d7f9b8004a664ece33ab36d0634ee285e4886f64d84e261c"
+        }
+    },
+    "pre_install": [
+        "$protoHome = if ($env:PROTO_HOME) { $env:PROTO_HOME } else { \"$HOME\\.proto\" }",
+        "if (!(Test-Path \"$protoHome\\bin\")) { New-Item -ItemType Directory -Force -Path \"$protoHome\\bin\" | Out-Null }"
+    ],
+    "installer": {
+        "script": [
+            "$protoHome = if ($env:PROTO_HOME) { $env:PROTO_HOME } else { \"$HOME\\.proto\" }",
+            "Move-Item \"$dir\\*.exe\" -Destination \"$protoHome\\bin\" -Force",
+            "& \"$protoHome\\bin\\proto.exe\" setup --yes --no-modify-profile 2>&1 | Out-Null"
+        ]
+    },
+    "uninstaller": {
+        "script": [
+            "$protoHome = if ($env:PROTO_HOME) { $env:PROTO_HOME } else { \"$HOME\\.proto\" }",
+            "",
+            "# Remove proto paths from PATH",
+            "$userPath = [Environment]::GetEnvironmentVariable('PATH', 'User')",
+            "$binDir = \"$protoHome\\bin\"",
+            "$shimsDir = \"$protoHome\\shims\"",
+            "$paths = $userPath -split ';' | Where-Object { $_ -and $_ -ne $binDir -and $_ -ne $shimsDir }",
+            "[Environment]::SetEnvironmentVariable('PATH', ($paths -join ';'), 'User')",
+            "",
+            "# Remove PROTO_HOME directory",
+            "if (Test-Path $protoHome) { Remove-Item -Path $protoHome -Recurse -Force }"
+        ]
+    },
+    "checkver": {
+        "github": "https://github.com/moonrepo/proto"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/moonrepo/proto/releases/download/v$version/proto_cli-x86_64-pc-windows-msvc.zip"
+            }
+        },
+        "hash": {
+            "url": "$url.sha256"
+        }
+    }
+}


bucket/proto.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "0.54.1",
+    "description": "Proto is a pluggable version manager, a unified toolchain.",
+    "homepage": "https://moonrepo.dev/docs/proto",
```
z-Fng:
```suggestion
    "homepage": "https://moonrepo.dev/proto",
```

bucket/proto.json
```diff
@@ -0,0 +1,34 @@
...
+        }
+    },
+    "bin": [
+        "proto.exe",
+        "proto-shim.exe"
+    ],
+    "env_set": {
+        "PROTO_HOME": "$persist_dir"
+    },
+    "env_add_path": "$persist_dir\\bin",
```
z-Fng: Why is it required to include `$persist_dir\bin` directory in the PATH?

bucket/proto.json
```diff
@@ -0,0 +1,34 @@
...
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/moonrepo/proto/releases/download/v$version/proto_cli-x86_64-pc-windows-msvc.zip"
+            }
+        },
+        "hash": {
+            "url": "$url.sha256",
+            "regex": "([A-Fa-f0-9]{64})"
+        }
```
z-Fng, line 49:
```suggestion
        "hash": {
            "url": "$url.sha256"
        }
```

bucket/proto.json
```diff
@@ -0,0 +1,34 @@
...
+            "url": "https://github.com/moonrepo/proto/releases/download/v0.54.1/proto_cli-x86_64-pc-windows-msvc.zip",
+            "hash": "77e5cf3b6ee89882d7f9b8004a664ece33ab36d0634ee285e4886f64d84e261c"
+        }
+    },
+    "bin": [
+        "proto.exe",
+        "proto-shim.exe"
+    ],
+    "env_set": {
+        "PROTO_HOME": "$persist_dir"
```
z-Fng: Use [`extract_to`](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties) to unpack files into the `bin` folder.

Shouldn't we set `PROTO_HOME` to `$dir` instead, and then `persist` specific folders as needed?

Ref: https://github.com/ScoopInstaller/Main/blob/master/bucket/nginx.json

bucket/proto.json
nostalume: I don't know whether the regex is needed due to the name suffix.

bucket/proto.json
nostalume: > Why is it required to include `$persist_dir\bin` directory in the PATH?

`proto` could achieve the same role by `proto setup` as install script shown. But directly download won't execute that as install script does.

bucket/proto.json
nostalume: > Use [`extract_to`](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties) to unpack files into the `bin` folder.
> 
> Shouldn't we set `PROTO_HOME` to `$dir` instead, and then `persist` specific folders as needed?
> 
> Ref: https://github.com/ScoopInstaller/Main/blob/master/bucket/nginx.json

Good idea, I guess it's a better choice. Here the quote from [proto](https://moonrepo.dev/docs/proto/commands/debug/env):

```
$ proto debug env

Store ─────────────────────────────────────────────────────────────────────

  Root: /Users/name/.proto
  Bins: /Users/name/.proto/bin
  Shims: /Users/name/.proto/shims
  Plugins: /Users/name/.proto/plugins
  Tools: /Users/name/.proto/tools
  Temp: /Users/name/.proto/temp

```

Lutra-Fs:
> @z-Fng
> 
> I will try to identify the `persist`, currently, only `.prototools` is needed.
> 
> The new workflow is shown below, please tell me any potential improvement. If it's the final one, I will refactor the manifest.
> 
> * Extract and move the binaries into `$dir/bin`.
> * Remove `env_add_path`, add post script akin to `proto.ps1` install script that executes `/path/to/proto setup` to generate automatically.
> * link `.prototools` configuration only.
> 
> Quote from `proto.ps1`:
> 
> > `irm https://moonrepo.dev/install/proto.ps1 | iex`
> 
> ```powershell
>   32 │ $TempDir = "${HOME}\.proto\temp\proto\${Target}"
>   33 │ $DownloadFile = "${TempDir}.zip"
>   34 │
>   35 │ $InstallDir = if ($env:PROTO_HOME) {
>   36 │   "$($env:PROTO_HOME)\bin"
>   37 │ } else {
>   38 │   "${Home}\.proto\bin"
>   39 │ }
>   40 │
>   41 │ $BinPath = "${InstallDir}\proto.exe"
>   42 │ $ShimPath = "${InstallDir}\proto-shim.exe"
> ```
> 
> > proto supports 3 locations in which a .prototools file can exist. These locations are used throughout the command line and proto's own settings.
> > local -> ./.prototools, ..prototools (current directory)
> > global -> ~/.proto/.prototools, %USERPROFILE%.proto.prototools
> > user -> ~/.prototools, %USERPROFILE%.prototools
> 
> A example of proto environment:
> 
> ```
> $ proto debug env
> 
> Store ─────────────────────────────────────────────────────────────────────
> 
>   Root: /Users/name/.proto
>   Bins: /Users/name/.proto/bin
>   Shims: /Users/name/.proto/shims
>   Plugins: /Users/name/.proto/plugins
>   Tools: /Users/name/.proto/tools
>   Temp: /Users/name/.proto/temp
>  
> Environment ───────────────────────────────────────────────────────────────
> 
>   Proto version: 0.44.0
>   Operating system: macos
>   Architecture: arm64
>   Config sources:
>     - /Users/name/Projects/example/.prototools
>     - /Users/name/.proto/.prototools
>   Virtual paths:
>     /userhome = /Users/name
>     /proto = /Users/name/.proto
>   Environment variables:
>     PROTO_APP_LOG = proto=info,schematic=info,starbase=info,warpgate=info,extism::pdk=info
>     PROTO_HOME = /Users/name/.proto
>     PROTO_OFFLINE_TIMEOUT = 750
>     PROTO_VERSION = 0.44.0
> ```

The current `persist` configuration has an issue: Scoop's persist creates directories by default, but proto expects configuration files. You can reference https://github.com/ScoopInstaller/Main/blob/5e985e0a98140354e18c3ec5089ce8be2ba9243b/bucket/px.json#L12-L14

bucket/proto.json
Lutra-Fs: ref: https://moonrepo.dev/docs/proto/workflows

Lutra-Fs: Should we add a note like https://github.com/ScoopInstaller/Main/blob/5e985e0a98140354e18c3ec5089ce8be2ba9243b/bucket/starship.json#L6-L9 so users can activate themselves?

nostalume: > Should we add a note like
> 
> https://github.com/ScoopInstaller/Main/blob/5e985e0a98140354e18c3ec5089ce8be2ba9243b/bucket/starship.json#L6-L9
> 
> so users can activate themselves?

No, because if we didn't setup, user should resolve the path of `proto` to setup, that's the reason why `setup` exists, to load in the environment.

nostalume:
@Lutra-Fs 

> The current persist configuration has an issue: Scoop's persist creates directories by default, but proto expects configuration files. You can reference.

It doesn't seems a problem, where `scoop` **should** automatically detects file. Your example only check the existence of dir only, notice the backslash `px.ini\`.

bucket/proto.json
Lutra-Fs: So it basically means we need to persist all above except temp?

bucket/proto.json
Lutra-Fs: > > Should we add a note like
> > https://github.com/ScoopInstaller/Main/blob/5e985e0a98140354e18c3ec5089ce8be2ba9243b/bucket/starship.json#L6-L9
> > 
> > so users can activate themselves?
> 
> No, because if we didn't setup, user should resolve the path of `proto` to setup, that's the reason why `setup` exists, to load in the environment.

I missed that command, was always put activate command manually to the PROFILE before 😵

bucket/proto.json
Lutra-Fs: or we should persist temp as well?

Lutra-Fs:
I am against calling proto setup directly via CLI, since it will not only modify the PATH, but also the terminal profile. And I did not see a command to reverse the effect by the command, which will left unmanaged env var and scripts in PROFILE, in this case.

Lutra-Fs:
> @Lutra-Fs
> 
> > The current persist configuration has an issue: Scoop's persist creates directories by default, but proto expects configuration files. You can reference.
> 
> It doesn't seems a problem, where `scoop` **should** automatically detects file. Your example only check the existence of dir only, notice the backslash `px.ini\`.

From my memory, full file support in persist is still pending review, so this might not work reliably yet. If you have tested this, and it behaves correctly in your environment, that would be good to confirm.

bucket/proto.json
nostalume: > So it basically means we need to persist all above except temp?

No, they are not configuration.

nostalume:
> I am against calling proto setup directly via CLI, since it will not only modify the PATH, but also the terminal profile. And I did not see a command to reverse the effect by the command, which will left unmanaged env var and scripts in PROFILE, in this case.

It's the official installation procedure of install script, which is the workflow of proto itself canonically.

Quote from [moonrepo](https://moonrepo.dev/docs/proto/install#uninstalling)

> Uninstalling
> To uninstall proto, delete the ~/.proto directory, and remove any PROTO_HOME references from your shell profile.

bucket/proto.json
Lutra-Fs: > > So it basically means we need to persist all above except temp?
> 
> No, they are not configuration.

If a user had tools installed and shims generated under the old layout, then upgraded, any shims that lived in the old location would not be preserved across the upgrade, and this might be an unintended behaviour.

nostalume: > > > So it basically means we need to persist all above except temp?
> > 
> > 
> > No, they are not configuration.
> 
> If a user had tools installed and shims generated under the old layout, then upgraded, any shims that lived in the old location would not be preserved across the upgrade, and this might be an unintended behaviour.

I haven't consider that, it's a resonable consideration. I will modify workflow. But it means when user remove the package, the shim shouldn't exist too. It's not a permanent configuration in structure. But in such view, All subdirectories have to be placed in persist in update. I guess the only way to update is to replace binary only, so it's compatible both for user to update by `proto` or by `scoop`.

Lutra-Fs:
> > I am against calling proto setup directly via CLI, since it will not only modify the PATH, but also the terminal profile. And I did not see a command to reverse the effect by the command, which will left unmanaged env var and scripts in PROFILE, in this case.
> 
> It's the official installation procedure of install script, which is the workflow of proto itself canonically.
> 
> Quote from [moonrepo](https://moonrepo.dev/docs/proto/install#uninstalling)
> 
> > Uninstalling
> > To uninstall proto, delete the ~/.proto directory, and remove any PROTO_HOME references from your shell profile.

No, that uninstall note applies to the generic install flow, not to proto setup specifically, and setup is effectively invoking proto activate behind the scenes.

> The proto setup command will setup proto in your current shell by modifying an applicable profile file and appending proto's bin directory to PATH.

which introduces additional, persistent side effects that are not reverted by the documented uninstall steps.

lvyuemeng:
AI(Claude) version of manifest:

```json
    "pre_install": [
        "# Determine PROTO_HOME",
        "if ($env:PROTO_HOME) {",
        "    $protoHome = $env:PROTO_HOME",
        "} else {",
        "    $protoHome = \"$HOME\\.proto\"",
        "}",
        "",
        "# Create directory structure",
        "New-Item -ItemType Directory -Force -Path \"$protoHome\\bin\" | Out-Null",
        "New-Item -ItemType Directory -Force -Path \"$protoHome\\shims\" | Out-Null",
        "New-Item -ItemType Directory -Force -Path \"$protoHome\\plugins\" | Out-Null",
        "New-Item -ItemType Directory -Force -Path \"$protoHome\\tools\" | Out-Null",
        "New-Item -ItemType Directory -Force -Path \"$protoHome\\temp\" | Out-Null",
        "",
        "# Store PROTO_HOME location for later use",
        "Set-Content -Path \"$dir\\proto_home.txt\" -Value $protoHome"
    ],
    "installer": {
        "script": [
            "$protoHome = Get-Content \"$dir\\proto_home.txt\" -Raw | ForEach-Object { $_.Trim() }",
            "$binDir = \"$protoHome\\bin\"",
            "",
            "# Copy proto binaries to PROTO_HOME/bin",
            "Copy-Item \"$dir\\proto.exe\" -Destination \"$binDir\\proto.exe\" -Force",
            "if (Test-Path \"$dir\\proto-shim.exe\") {",
            "    Copy-Item \"$dir\\proto-shim.exe\" -Destination \"$binDir\\proto-shim.exe\" -Force",
            "}",
            "",
            "# Add PROTO_HOME/bin to PATH if not already present",
            "$userPath = [Environment]::GetEnvironmentVariable('PATH', 'User')",
            "if ($userPath -notlike \"*$binDir*\") {",
            "    [Environment]::SetEnvironmentVariable('PATH', \"$binDir;$userPath\", 'User')",
            "    Write-Host \"Added $binDir to PATH\" -ForegroundColor Green",
            "}",
            "",
            "# Set PROTO_HOME environment variable only if not already set",
            "if (-not $env:PROTO_HOME) {",
            "    [Environment]::SetEnvironmentVariable('PROTO_HOME', $protoHome, 'User')",
            "    Write-Host \"Set PROTO_HOME to $protoHome\" -ForegroundColor Green",
            "} else {",
            "    Write-Host \"PROTO_HOME already set to $env:PROTO_HOME\" -ForegroundColor Yellow",
            "}"
        ]
    },
    "uninstaller": {
        "script": [
            "$protoHome = Get-Content \"$dir\\proto_home.txt\" -Raw | ForEach-Object { $_.Trim() }",
            "$binDir = \"$protoHome\\bin\"",
            "$shimsDir = \"$protoHome\\shims\"",
            "",
            "# Remove from PATH",
            "$userPath = [Environment]::GetEnvironmentVariable('PATH', 'User')",
            "$paths = $userPath -split ';' | Where-Object { $_ -ne $binDir -and $_ -ne $shimsDir }",
            "[Environment]::SetEnvironmentVariable('PATH', ($paths -join ';'), 'User')",
            "Write-Host \"Removed proto paths from PATH\" -ForegroundColor Green",
            "",
            "# Remove PROTO_HOME environment variable only if it points to our managed location",
            "if ($env:PROTO_HOME -eq $protoHome) {",
            "    [Environment]::SetEnvironmentVariable('PROTO_HOME', $null, 'User')",
            "    Write-Host \"Removed PROTO_HOME environment variable\" -ForegroundColor Green",
            "}",
            "",
            "# Remove the entire PROTO_HOME directory tree",
            "if (Test-Path $protoHome) {",
            "    Remove-Item -Path $protoHome -Recurse -Force -ErrorAction SilentlyContinue",
            "    Write-Host \"Removed $protoHome directory\" -ForegroundColor Green",
            "}"
        ]
    },
        "notes": [
        "Proto has been installed to $env:PROTO_HOME\\bin (or %USERPROFILE%\\.proto\\bin if PROTO_HOME was not set)",
        "The following directories have been created:",
        "  - $env:PROTO_HOME\\bin (binaries)",
        "  - $env:PROTO_HOME\\shims (tool shims)",
        "  - $env:PROTO_HOME\\plugins",
        "  - $env:PROTO_HOME\\tools",
        "  - $env:PROTO_HOME\\temp",
        "",
        "Please restart your terminal for PATH changes to take effect.",
        "",
        "To update proto itself, run: scoop update proto",
        "To install tools, run: proto install <tool> <version>"
    ]
```

nostalume:
@z-Fng 

Here the manifest updated, sadly, due to its special property, the script is necessary in following reasons.

Reasons:

- Due to `proto` the role of package manager, exposing `$PROTO_HOME/bin` and `$PROTO_HOME/shims` for setup. I choose not to interfere any functionality by scoop own shims and apps management.
- Due to `proto` detecting its location in a constraint of `$PROTO_HOME/bin`, I choose move `proto` into `$PROTO_HOME/bin` rather in `$dir`.
- Due to persistency and tree structure maintaining of `$PROTO_HOME` subdirectories, any update will break the tree structure in `$dir`, so anything is located in default or custom `$PROTO_HOME`.

bucket/proto.json
```diff
@@ -0,0 +1,51 @@
...
+            "",
+            "# Remove proto paths from PATH",
+            "$userPath = [Environment]::GetEnvironmentVariable('PATH', 'User')",
+            "$binDir = \"$protoHome\\bin\"",
+            "$shimsDir = \"$protoHome\\shims\"",
+            "$paths = $userPath -split ';' | Where-Object { $_ -and $_ -ne $binDir -and $_ -ne $shimsDir }",
+            "[Environment]::SetEnvironmentVariable('PATH', ($paths -join ';'), 'User')",
+            "",
+            "# Remove PROTO_HOME directory",
+            "if (Test-Path $protoHome) { Remove-Item -Path $protoHome -Recurse -Force }"
```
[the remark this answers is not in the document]
nostalume: It's necessary quote from:

> Uninstalling
> To uninstall proto, delete the ~/.proto directory, and remove any PROTO_HOME references from your shell profile.

PR #7459, opened 2025-12-26
git(-with-openssh): Improve registry srcipt & relevant logic, update checkver & autoupdate

SorYoshino:
### Summary

Improves the `git` and `git-with-openssh` manifests by refining metadata, restructuring post-install behavior, modernizing registry script handling, and aligning both packages with Scoop manifest conventions.

### Related issues or pull requests

- Relates to #7061 
- Relates to #3685 
- Relates to #6259 

### Changes

- Refine `description` to a formal and complete wording  
- Update `license` to `GPL-2.0-or-later` and added license URL  
- Refactor `notes` section using formal language and clear execution instructions  
- Reorganize `bin` entries
- Add `env_add_path` for automatic PATH integration  
- Standardize `shortcuts` names (grouped under “Git\*”)   
- Simplify and generalize `post_install` scripts
- Add uninstall script blocks to restore registry changes on removal  
- update checkver & autoupdate  
- Normalize registry file generation encoding and escaping rules  

### Notes

- The official installer directly appends the `cmd` directory to the `Path` environment variable, and the useful tools contained in this directory extend well beyond `git`, `gitk`, and `scalar`. Rather than enumerating these utilities individually in the `bin` field, it is more appropriate to add the entire `cmd` directory to `Path`.
- Tools related to `gpg` are not in the same category as `ssh`; however, they are critical during development and should be included in the `bin` field for both `git` and `git-with-openssh`.
- The contents of `install-context.reg` and `install-associations.reg` have been realigned with the behavior of the official installer, and the associated logic has been refined accordingly.

### Testing

<details>
<summary>The test results are as follows:</summary><br>

```powershell
┏[ D:\Software\Scoop\Local\apps\scoop\current\bin][ master ≡]
└─> .\checkver.ps1 -App git -Dir 'D:\Temporary\Software\Microsoft\Windows Sandbox\Repositories\Scoop\Buckets\Main\bucket' -f
git: 2.52.0 (scoop version is 2.52.0)
Forcing autoupdate!
Autoupdating git
DEBUG[1766751456] $substitutions (hashtable) -> D:\Software\Scoop\Local\apps\scoop\current\lib\autoupdate.ps1:221:5
DEBUG[1766751456] $substitutions.$dotVersion                    2.52.0
DEBUG[1766751456] $substitutions.$match1                        2.52.0
DEBUG[1766751456] $substitutions.$basenameNoExt                 PortableGit-2.52.0-arm64.7z
[...]

Issue #2894, opened 2021-11-05
Suggest change git from portable to setup


wentong-wow:
## Package Request
- git

### Why
I have install git to gitlab ci runner with main bucket. However, gitlab ci runner execute job will be stucked and timeout at:

> created fresh repository.

It was because portable has default config will cause git always popup to ask for user credentials when fetch. After I try to install by setup, everything is OK. So I suggest to change git from portable to setup. Thanks.

P.S. this issue take me a lots of time to debug. QQ

### Suggested App Manifest
``` json
{
    "version": "2.33.1.windows.1",
    "description": "Distributed version control system",
    "homepage": "https://gitforwindows.org",
    "license": "GPL-2.0-only",
    "architecture": {
        "64bit": {
            "url": "https://github.com/git-for-windows/git/releases/download/v2.33.1.windows.1/Git-2.33.1-64-bit.exe",
            "hash": "b1b69fcf56d50199536f7e6fc79b75ab16734d4d9a4b85c8e931596f02dd0688"
        },
        "32bit": {
            "url": "https://github.com/git-for-windows/git/releases/download/v2.33.1.windows.1/Git-2.33.1-32-bit.exe",
            "hash": "bc2ca8f6ffcc6914ed04cfb13b2d0b6b994bd2437283cb3b43188b969bd5ef8b"
        }
    },
    "bin": [
        "cmd\\git.exe",
        "cmd\\gitk.exe",
        "cmd\\git-gui.exe",
        "usr\\bin\\tig.exe",
        "git-bash.exe"
    ],
    "installer": {
        "args": [
            "/DIR=$dir",
            "/SILENT"
        ]
    },
    "shortcuts": [
        [
            "git-bash.exe",
            "Git Bash",
            "--cd-to-home"
        ],
        [
            "cmd\\git-gui.exe",
            "Git GUI"
        ]
    ],
    "env_set": {
        "GIT_INSTALL_ROOT": "$dir"
    },
    "checkver": {
        "github": "https://github.com/git-for-windows/git",
        "regex": "v([\\w.]+)/Git-(?<full>[\\w.]+)-64-bit"
    },
    "autoupdate": {
        "architecture": {
            "64bit": {
                "url": "https://github.com/git-for-windows/git/releases/download/v$version/Git-$matchFull-64-bit.exe"
            },
            "32bit": {
                "url": "https://github.com/git-for-windows/git/releases/download/v$version/Git-$matchFull-32-bit.exe"
            }
        },
[...]

PR #7531, opened 2026-01-13
git(-with-openssh): Restore `git` shim

Hahihula:
The recent update of the git.json and git-with-openssh.json remove the shim for git.exe which was previously there. This update returns it back an not only git-bash.exe should be exposed in the shims

Relates to #7459

### DIFF
--- a/bucket/git-with-openssh.json
+++ b/bucket/git-with-openssh.json
@@ -40,6 +40,7 @@
     },
     "bin": [
         "bin\\sh.exe",
+        "bin\\git.exe",
         "git-bash.exe",
         "usr\\bin\\gpg.exe",
         "usr\\bin\\gpg-agent.exe",
--- a/bucket/git.json
+++ b/bucket/git.json
@@ -40,6 +40,7 @@
     },
     "bin": [
         "bin\\sh.exe",
+        "bin\\git.exe",
         "git-bash.exe",
         "usr\\bin\\gpg.exe",
         "usr\\bin\\gpg-agent.exe",


SorYoshino:
> The official installer directly appends the `cmd` directory to the `Path` environment variable, and the useful tools contained in this directory extend well beyond `git`, `gitk`, and `scalar`. Rather than enumerating these utilities individually in the `bin` field, it is more appropriate to add the entire `cmd` directory to `Path`.

- The reason for removing `cmd\git.exe` has already been provided in #7459, as shown above.

- Adding `$dir\cmd` to the `Path` variable does not affect the user's ability to use Git in `powershell` or `cmd`. Therefore, there is no need to re-add `git.exe` to the `bin` field.

```powershell
┏[ ~]
└─> Get-Command -Name 'git.exe' -CommandType Application

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     git.exe                                            2.52.0.1   D:\Software\Scoop\Global\apps\Git\current\cmd\git.exe

┏[ ~]
└─> git --version
git version 2.52.0.windows.1
```

SorYoshino:
Additionally, I removed `bin\bash.exe` from the `bin` field because the hash of `bin\bash.exe` is identical to that of `bin\sh.exe`. Furthermore, users who have WSL (Windows Subsystem for Linux) installed may experience conflicts or issues when attempting to use `bin\bash.exe`.

Hahihula:
> > The official installer directly appends the `cmd` directory to the `Path` environment variable, and the useful tools contained in this directory extend well beyond `git`, `gitk`, and `scalar`. Rather than enumerating these utilities individually in the `bin` field, it is more appropriate to add the entire `cmd` directory to `Path`.
> 
> * The reason for removing `cmd\git.exe` has already been provided in [git(-with-openssh): Improve registry srcipt & relevant logic, update checkver & autoupdate #7459](https://github.com/ScoopInstaller/Main/pull/7459), as shown above.
> * Adding `$dir\cmd` to the `Path` variable does not affect the user's ability to use Git in `powershell` or `cmd`. Therefore, there is no need to re-add `git.exe` to the `bin` field.
> 
> ```powershell
> ┏[ ~]
> └─> Get-Command -Name 'git.exe' -CommandType Application
> 
> CommandType     Name                                               Version    Source
> -----------     ----                                               -------    ------
> Application     git.exe                                            2.52.0.1   D:\Software\Scoop\Global\apps\Git\current\cmd\git.exe
> 
> ┏[ ~]
> └─> git --version
> git version 2.52.0.windows.1
> ```

This is only true if there would be no existing third party integration expecting it to be there, also this means that after installing adding the shims folder to path is not enough which previously was.
So it does not affect the ability of users to use git in new user session, but they actually can not use it right away as they previously could.

SorYoshino:
> So it does not affect the ability of users to use git in new user session, but they actually can not use it right away as they previously could.

In practice, this does not prevent users from immediately using `git`, as can be seen from the relevant code and the examples below.

https://github.com/ScoopInstaller/Scoop/blob/b588a06e41d920d2123ec70aee682bae14935939/lib/system.ps1#L96-L120

```powershell
┏[ ~]
└─> scoop list ldc
Installed apps matching 'ldc':

┏[ ~]
└─> $env:Path -match 'ldc'
False

┏[ ~]
└─> scoop install ldc
Installing 'ldc' (1.41.0) [64bit] from 'main' bucket
Loading ldc2-1.41.0-windows-x64.7z from cache.
Checking hash of ldc2-1.41.0-windows-x64.7z ... ok.
Extracting ldc2-1.41.0-windows-x64.7z ... done.
Linking D:\Software\Scoop\Local\apps\ldc\current => D:\Software\Scoop\Local\apps\ldc\1.41.0
Adding D:\Software\Scoop\Local\apps\ldc\current\bin to your path.
'ldc' (1.41.0) was installed successfully!

┏[ ~]
└─> $env:Path -match 'ldc'
True

┏[ ~]
└─> Get-Command -Name 'ldc2.exe' -CommandType Application

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     ldc2.exe                                           0.0.0.0    D:\Software\Scoop\Local\apps\ldc\current\bin\ldc2.exe

```

z-Fng:
Is there a conflict if the `bin\git.exe` shim co-exists with `cmd\git.exe` in the PATH? If not, we can consider keeping both.

SorYoshino:
> Is there a conflict if the `bin\git.exe` shim co-exists with `cmd\git.exe` in the PATH?

Under normal usage, there should be no conflicts. Which `git` is actually used depends on the order of `$shimdir` and `cmd` in the PATH environment variable.

---

The following content is newly added:

However, do we really need to recreate the `bin\git.exe` shim?

I previously fell into a habitual line of thinking and did not consider shells other than PowerShell. That said, the situation where `git` cannot be found would generally occur only once. Repeated occurrences would only be possible if `no_junction` is set to true and `scoop cleanup git` is executed after updating Git, while continuing to use the same session without starting a new one.

The likelihood of all these conditions being met simultaneously is very low. It is also reasonable to assume that most users would not keep using the same session for an extended period without reloading environment variables. In my view, a one-time occurrence is acceptable, as this behavior is consistent with how traditional installers work.

Do we really want to introduce an approach that appears inherently contradictory in order to address an issue that is unlikely to occur?

eugenesvk:
I've also been hit by this breakage and don't see any good rationale for that

> The official installer directly appends the cmd ... Rather than enumerating these utilities individually in the bin field, it is more appropriate to add the entire cmd directory to Path.

How is it more appropriate? One of the main benefits of using Scoop is that I don't have to care about peculiarities of many of these individual installers and instead rely on a common architecture, in this case on one of the core https://scoop.sh feature

> What does Scoop do?
> Prevents PATH pollution from installing lots of programs

Why are you changing Scoop fundamentals without any substantial explanation ("appropriate" isn't one)?


> Adding $dir\cmd to the Path variable does not affect the user's ability to use Git in powershell or cmd. Therefore, there is no need to re-add git.exe to the bin field

There is because users aren't limited to powershell/cmd. For example, I'm using the shim in a GUI app and it took some time to understand why it started to fail all of a sudden.

> Additionally, I removed bin\bash.exe from the bin field because the hash of bin\bash.exe is identical to that of bin\sh.exe.

(not about git, but the faulty logic still bothers me)
So? How does it help if it still breaks users and requires them wasting time doing manual workarounds?

SorYoshino:
> How is it more appropriate?

Since there are nine files in the `cmd` directory, I believe that instead of adding them to the `bin` field one by one, it would be better to add the `cmd` directory directly to the PATH environment variable, as the official installer does. When implementing #7459, I did not take into account environment variable handling in shells other than PowerShell, nor the possibility that users may have hardcoded shims in workflows or application configurations, which led to the removal of `cmd\git.exe` from the `bin` field. This was an oversight on my part, and I sincerely apologize.

> Prevents PATH pollution from installing lots of programs

Scoop’s responsibility should be to minimize pollution of the PATH environment variable, rather than to eliminate it entirely. In this case, I think adding to the PATH environment variable is entirely reasonable.

eugenesvk:
You've just repeated your belief, which I understood the first time. But you again didn't explain what problem replacing 9 shims with one extra path solves (and again, reference to official installers is an empty one, it provides no justification, scoop explicitly overrides official installers for various  user benefits).

Same with path - ok, you find it reasonable, but the question was why? It directly contradicts the core design of Scoop. Is git the only app with multiple binaries? What if all other similar apps are added to path, how would that be reasonable instead of pollution?

SorYoshino:
> But you again didn't explain what problem replacing 9 shims with one extra path solves

I think the manifest can be made more concise.

> and again, reference to official installers is an empty one

This is the default value used by the Git installer; you can refer to [`install.iss`](https://github.com/git-for-windows/build-extra/blob/d42e7d86d7434fc6a8f5a9c5bca7348c27c1b915/installer/install.iss#L2216-L2238) and the screenshot below.

Screenshot 2026-02-11 050822 <br>

> It directly contradicts the core design of Scoop. Is git the only app with multiple binaries? What if all other similar apps are added to path, how would that be reasonable instead of pollution?

I do not believe this violates Scoop’s core design principles. As I mentioned earlier, the goal is to minimize pollution of the PATH environment variable, not to eliminate it entirely.

You are right that Git is not the only application with multiple binaries, but users clearly will not install every manifest that contains an `env_add_path` field. Moreover, at what point can the number of PATH entries truly be considered “pollution”? I believe the answer varies from person to person. Excluding duplicates, there are currently a little over 70 manifests with an `env_add_path` field in the Main bucket, and roughly over 500 across all official buckets. In typical usage, I estimate that the number actually installed would not exceed 30, and likely far fewer. Admittedly, this is only an estimate and is not backed by concrete data.

At present, there are about 6,200 manifests across all official buckets, and fewer than one tenth of them include an `env_add_path` field. This alone sufficiently demonstrates that efforts are already being made to minimize pollution of the PATH environment variable. Considering normal, real-world usage, there is little cause for concern about PATH pollution; even if it occurs, it remains well within a controllable range.

z-Fng:
I'd vote for keeping both if they don't conflict. https://github.com/ScoopInstaller/Main/issues/7620#issuecomment-3873834371 https://github.com/ScoopInstaller/Main/pull/7531#issuecomment-3848984784

> Shims are about more than just the PATH. Some users have hardcoded them into their workflows or other environment variables.

Any objections/thoughts from other maintainers? cc @ScoopInstaller/maintainers

z-Fng (approved):

ofek:
I think that this should be reverted as I perceive adding duplicate entries for the same binary to be quite a poor user experience.

```
❯ which -a git
╭───┬─────────┬──────────────────────────────────────────────────┬──────────╮
│ # │ command │                       path                       │   type   │
├───┼─────────┼──────────────────────────────────────────────────┼──────────┤
│ 0 │ git     │ C:\Users\ofek\scoop\apps\git\current\cmd\git.exe │ external │
│ 1 │ git     │ C:\Users\ofek\scoop\shims\git.exe                │ external │
╰───┴─────────┴──────────────────────────────────────────────────┴──────────╯
```

I found out about this just now when I was looking into a Git issue and Codex had to spend cycles determining why there were two, which one was actually used, etc. Incidentally, their bug report form has a section for a dump of the debug command that happens to contain such information (I assume due to it being a common source of issues):

```nu
❯ codex doctor --json | from json | get checks."git.environment".details
╭──────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ .git entry           │ directory                                                                                                                                                                    │
│ PATH git #1          │ C:\Users\ofek\scoop\apps\git\current\cmd\git.exe                                                                                                                             │
│ PATH git #2          │ C:\Users\ofek\scoop\shims\git.exe                                                                                                                                            │
│ PATH git entries     │ 2                                                                                                                                                                            │
│ git branch           │ ofek/pyemscripten                                                                                                                                                            │
│ git build options    │ git version 2.54.0.windows.1; cpu: x86_64; built from commit: 2b8a3ab140826ac423c2845ef81d4c6ac4f7bf3c; sizeof-long: 4; sizeof-size_t: 8; shell-path:                        │
│                      │ D:/git-sdk-64-build-installers/usr/bin/sh; rust: disabled; feature: fsmonitor--daemon; gettext: enabled; libcurl: 8.19.0; OpenSSL: OpenSSL 3.5.6 7 Apr 2026; zlib: 1.3.2;    │
│                      │ SHA-1: SHA1_DC; SHA-256: SHA256_BLK; default-ref-format: files; default-hash: sha1                                                                                           │
│ git exec path        │ C:/Users/ofek/scoop/apps/git/2.54.0/mingw64/libexec/git-core                                                                                                                 │
│ git version          │ git version 2.54.0.windows.1                                                                                                                                                 │
│ repo detected        │ true                                                                                                                                                                         │
│ repo root            │ C:\Users\ofek\Desktop\code\msgspec                                                                                                                                           │
│ selected git         │ C:\Users\ofek\scoop\apps\git\current\cmd\git.exe                                                                                                                             │
╰──────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

ofek:
Many apps I come across prominently show that there are multiple Git binaries available and it's very confusing. Is it possible to revert this change?

image

z-Fng:
As seen in the previous discussion, reverting this change would disrupt the workflows of many users who expect the Git shim to be in the Scoop shims directory.

ofek:
I guess my question would be then if we started from scratch would we choose a shim?

ofek:
It's not in the current working directory, but here's an example of how multiple Git binaries on PATH can lend itself to insecure resolution mechanisms and potential flagging by security scanners. https://mindgard.ai/blog/cursor-0day-when-full-disclosure-becomes-the-only-protection-left

z-Fng:
> Many apps I come across prominently show that there are multiple Git binaries available and it's very confusing. Is it possible to revert this change?

A temporary workaround is to remove the Git shim via command `scoop shim rm git`.

> I guess my question would be then if we started from scratch would we choose a shim?

For me personally, I'd still go with shims. Maybe reverting the change that adds the `cmd` directory to the PATH is a better fit? But I'm not sure if this might break things. If I remember correctly, adding the cmd directory to the PATH is exactly what allows Claude-Code to properly recognize Scoop-installed Git.

Any thoughts? cc: @ScoopInstaller/maintainers

chawyehsu:
> Any thoughts? cc: @ScoopInstaller/maintainers

I've long since switched to maintaining [my own Git variant](https://github.com/chawyehsu/dorado/blob/master/bucket/git.json) (https://github.com/ScoopInstaller/Main/issues/2894#issuecomment-1125150531), so I can't really comment on the current state of this one. That said, I have two points to make. 1) I strongly prefer using shims, so I personally don't like the way that `cmd` has to be added to `PATH` (and I'd keep its `interface` minimal by not exposing the bundled gpg stuff, which do have some issue). 2) if you're unhappy with the current state of a manifest and can't reach a consensus with others, the path of least resistance is to maintain your own manifest and bucket. I've always leaned toward this approach. It's not about fragmenting the community; it's about recognizing that - all so-called workarounds to manifests on the main bucket will cause problems on the other side. Changes driven by strongly held personal preferences should be added to an official repository, which everyone depends on, **only with great caution**, if at all. Of course, there are always exceptional cases where maintaining a separate manifest is simply the more practical solution.

z-Fng:
> I've long since switched to maintaining [my own Git variant](https://github.com/chawyehsu/dorado/blob/master/bucket/git.json) ([#2894 (comment)](https://github.com/ScoopInstaller/Main/issues/2894#issuecomment-1125150531)), so I can't really comment on the current state of this one. That said, I have two points to make. 1) I strongly prefer using shims, so I personally don't like the way that `cmd` has to be added to `PATH` (and I'd keep its `interface` minimal by not exposing the bundled gpg stuff, which do have some issue). 2) if you're unhappy with the current state of a manifest and can't reach a consensus with others, the path of least resistance is to maintain your own manifest and bucket. I've always leaned toward this approach. It's not about fragmenting the community; it's about recognizing that - all so-called workarounds to manifests on the main bucket will cause problems on the other side. Changes driven by strongly held personal preferences should be added to an official repository, which everyone depends on, **only with great caution**, if at all. Of course, there are always exceptional cases where maintaining a separate manifest is simply the more practical solution.

Thank you for sharing your thoughts. I am currently on the same page with you. While I believed some of the PRs (e.g., #7653) I previously approved were beneficial, they appear to have been overly aggressive in retrospect. Community feedback and my personal usage habits prompted me to approve this PR which conflicts with #7459, leaving us in a dilemma now. We will likely need to allocate time to address and resolve this issue.

> For me personally, I'd still go with shims. Maybe reverting the change that adds the `cmd` directory to the PATH is a better fit? But I'm not sure if this might break things. If I remember correctly, adding the cmd directory to the PATH is exactly what allows Claude-Code to properly recognize Scoop-installed Git.

I might test the approach of reverting the change that adds the cmd directory to the PATH in a little while. Hopefully, this resolves the issue.

By the way, persistence for system-wide git config has been implemented in #7632.

PR #2931, opened 2021-11-11
lua: Change to Lua Binaries's build

niheaven:
Original one is moved to https://github.com/ScoopInstaller/Versions/pull/310

PR #7540, opened 2026-01-15
Lua@5.4.8-1: Remove troublesome environment variables

SereneRuby12:

I have been having issues with LUA_CPATH causing trouble, because it was being set to a wrong value, so I investigated why it even was added to the lua manifest here:

It was first added in #2931 and I can't find a reason why, the directories for it don't make any sense too, since LUA_PATH and LUA_CPATH (which replace `package.path` and `package.cpath` default values) must include the current directory (`.\?.lua` and `.\?.dll`) to be able to load modules via `require "my_module"`, which is the most common way to include files.

We can see those env vars didn't make sense because it caused issues in #3829, but the PR for it wasn't a proper fix, it just renamed LUA_PATH to a random environment path that isn't used anywhere (in the stack overflow page mentioned, the person that renamed it, was because they wanted to have an env var to the lua executable, not because it was used anywhere by any program), it kinda solved the issue though, because LUA_PATH isn't being set to wrong values anymore by scoop.

The proper solution is to just remove the LUA_PATH and LUA_CPATH so lua can use its default paths which work normally.

### DIFF
--- a/bucket/lua.json
+++ b/bucket/lua.json
@@ -24,10 +24,6 @@
         "bin\\lua.exe",
         "bin\\luac.exe"
     ],
-    "env_set": {
-        "LUA_EXE_PATH": "$dir\\bin",
-        "LUA_CPATH": "$dir\\bin"
-    },
     "checkver": {
         "url": "https://packages.msys2.org/api/search?query=lua&qtype=binpkg",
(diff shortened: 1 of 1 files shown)


Lutra-Fs:
Thanks for your contribution

PR #7563, opened 2026-01-25
git-xet: Add version 0.2.1

haussmann:
git-xet: Add version v0.2.0

- Can't use traditional Github checkver, because the repository contains releases for multiple different topics and no latest tag

Closes: https://github.com/ScoopInstaller/Main/issues/7516

## Summary by CodeRabbit

* **New Features**
  * Added support for the Xet protocol as a Git LFS custom transfer agent.

### DIFF
--- a/bucket/git-xet.json
+++ b/bucket/git-xet.json
@@ -20,8 +20,9 @@
     },
     "bin": "git-xet.exe",
     "checkver": {
-        "url": "https://github.com/huggingface/xet-core/raw/refs/heads/main/git_xet/windows_installer/Package.wxs",
-        "regex": "Version=\"([\\w.]+)\""
+        "url": "https://api.github.com/repos/huggingface/xet-core/releases?per_page=100",
+        "jsonpath": "$..tag_name",
+        "regex": "git-xet-v([\\d.]+)"
     },
     "autoupdate": {
         "architecture": {


z-Fng:
> * Tests necessary whether `git xet install` is required

I think adding a note to guide users is sufficient, since git xet install can be applied to different scopes.

```console
PS> git xet install --help
Perform the following action to register Git-Xet:

Set up the "xet" transfer agent under the name "lfs.customtransfer.xet" in
global Git config with the below values
    path = git-xet
    args = transfer
    concurrent = <according to value from option "--concurrency" (default true)>

If "--concurrency" is specified, set "lfs.concurrenttransfers" in the global
Git config to this number.

Usage: git-xet.exe install [OPTIONS]

Options:
      --system
          Set the "xet" transfer agent and the number of concurrent LFS transfers in the system git config, e.g. /etc/gitconfig instead of the global git config (~/.gitconfig)

      --local
          Set the "xet" transfer agent and the number of concurrent LFS transfers in the local repository's git config, instead of the global git config (~/.gitconfig)

      --path <PATH>
          The path to the repository. This argument only applies when "--local" is specified

      --concurrency <CONCURRENCY>
          The number of concurrent LFS uploads/downloads. Default 8

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

bucket/git-xet.json
```diff
@@ -0,0 +1,33 @@
...
+        "64bit": {
+            "url": "https://github.com/huggingface/xet-core/releases/download/git-xet-v0.2.0/git-xet-windows-x86_64.zip",
+            "hash": "070d91735c29dfadb3aa560687d5f803f3ff111a7d7df736a12d15aca058b111"
+        }
+    },
+    "bin": "git-xet.exe",
+    "checkver": {
+        "url": "https://api.github.com/repos/huggingface/xet-core/git/matching-refs/tags/git-xet-v",
+        "regex": "git-xet-v([0-9.]+)",
+        "reverse": true
```
z-Fng: I'd be cautious about using the `/git/matching-refs/tags` endpoint here. Because of its ascending sort order, pagination will break our logic. We should probably use the `/releases` endpoint instead.

haussmann:
> I think adding a note to guide users is sufficient [...]

Thank you! That makes sense to me. I added the following note:
Follow instructions in `git xet install --help` after installation.

bucket/git-xet.json
haussmann: This particular github repository has many other releases besides git-xet, so the releases api endpoint has an even higher chance to fall under the pagination issue.

I now pulled the version tag from their windows-installer file. What do you think about this?

haussmann:
/verify

@coderabbitai full review

bucket/git-xet.json
```diff
@@ -0,0 +1,33 @@
+{
+    "version": "0.2.1",
+    "description": "Git LFS custom transfer agent that implements upload and download of files using the Xet protocol.",
+    "homepage": "https://huggingface.co/docs/hub/xet/",
```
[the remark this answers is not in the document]
haussmann: Fixed in https://github.com/ScoopInstaller/Main/pull/7563/commits/75ab3a99e41f8e8749ba77ed6c7eb688c5cd9c56

bucket/git-xet.json
z-Fng: > This particular github repository has many other releases besides git-xet, so the releases api endpoint has an even higher chance to fall under the pagination issue.

Since the `releases` endpoint is sorted by time in descending order, pagination will simply cause a match failure, not the downgrade issues we'd get with the `tags` endpoint.

> I now pulled the version tag from their windows-installer file. What do you think about this?

I'd still prefer to the `releases` endpoint, keeping it consistent with most other manifests.

```json
    "checkver": {
        "url": "https://api.github.com/repos/huggingface/xet-core/releases",
        "jsonpath": "$..tag_name",
        "regex": "git-xet-v([\\d.]+)"
    },
```

haussmann: Thank you for the explanation 👍  I changed it to your suggestion with the added `?per_page=100` parameter. According to [github api docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases), this is the allowed maximum.

SorYoshino:
I believe the current `checkver` field should be revised, as it has several issues:

- Most manifests that use the GitHub API reference the repository ID rather than the repository name.
- By default, `per_page` is set to 30, which is sufficient for most cases. For `checkver` to fail under this setting, 30 consecutive releases would need to exclude git-xet, which is highly unlikely. Setting `per_page` to 100 is therefore unnecessary. If additional precaution is desired, I believe 45 would be sufficient.
- A pre-release version may be detected. Although this would not cause the manifest to update to a pre-release version, it would lead to a failure during update checking.

One possible solution are provided below:

```json
{
    "checkver": {
        "url": "https://api.github.com/repositories/855423236/releases?per_page=45",
        "jsonpath": "$[?(@.prerelease == false)].assets[?(@.name =~ /git-xet/i)].browser_download_url",
        "regex": "(?i)download/git-xet-v([\\d.]+)/git-xet"
    }
}
```

z-Fng:
> * Most manifests that use the GitHub API reference the repository ID rather than the repository name.

I'd suggest using repository names instead of repo IDs in the GitHub API, as the API now properly supports [redirects](https://github.com/ScoopInstaller/Extras/pull/16758#issuecomment-3642693909). To complement this, we're going to introduce an automation to automatically correct and sync repo names in our buckets.

> * By default, `per_page` is set to 30, which is sufficient for most cases. For `checkver` to fail under this setting, 30 consecutive releases would need to exclude git-xet, which is highly unlikely. Setting `per_page` to 100 is therefore unnecessary. If additional precaution is desired, I believe 45 would be sufficient.
> * A pre-release version may be detected. Although this would not cause the manifest to update to a pre-release version, it would lead to a failure during update checking.

It should be fine for now. But if you'd like to make it more robust, feel free to submit a PR.

SorYoshino:
> I'd suggest using repository names instead of repo IDs in the GitHub API, as the API now properly supports [redirects](https://github.com/ScoopInstaller/Extras/pull/16758#issuecomment-3642693909). To complement this, we're going to introduce an automation to automatically correct and sync repo names in our buckets.

Thank you for your explanation.

> It should be fine for now. But if you'd like to make it more robust, feel free to submit a PR.

I will submit a PR later to refine it.

Issue #7593, opened 2026-02-03
[Bug]: uutils-coreutils-lean upgrade to 0.6.0 fails: Could not find 'coreutils-0.6.0-x86_64-pc-windows-msvc'


przemoc:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

uutils-coreutils-lean

### Expected/Current Behaviour

Expected: Successful upgrade of the package.  
Current: Failure.

P.S. `ERROR 'uutils-coreutils-lean' isn't installed correctly.` in steps below is because I repeated update command for copy-pasting here, as in the first run I was updating all packages.

### Steps to Reproduce

```console
PS> scoop update uutils-coreutils-lean
ERROR 'uutils-coreutils-lean' isn't installed correctly.
uutils-coreutils-lean: 0.5.0 -> 0.6.0
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
WARN  To disable this warning, run 'scoop config aria2-warning-enabled false'.
Updating one outdated app:
Updating 'uutils-coreutils-lean' (0.5.0 -> 0.6.0)
Downloading new version
Loading coreutils-0.6.0-x86_64-pc-windows-msvc.zip from cache.
Checking hash of coreutils-0.6.0-x86_64-pc-windows-msvc.zip ... ok.
Uninstalling 'uutils-coreutils-lean' (0.5.0)
Installing 'uutils-coreutils-lean' (0.6.0) [64bit] from 'main' bucket
Loading coreutils-0.6.0-x86_64-pc-windows-msvc.zip from cache.
Extracting coreutils-0.6.0-x86_64-pc-windows-msvc.zip ... DEBUG[1770115391] $stdoutTask.Result = -------------------------------------------------------------------------------
   ROBOCOPY     ::     Robust File Copy for Windows
-------------------------------------------------------------------------------

  Started : February 3, 2026 11:43:10
   Source : D:\scoop\apps\uutils-coreutils-lean\0.6.0\coreutils-0.6.0-x86_64-pc-windows-msvc\
     Dest : D:\scoop\apps\uutils-coreutils-lean\0.6.0\

    Files : *.*

[...]

PR #7596, opened 2026-02-03
uutils-coreutils(-lean)@0.6.0: Fix extract_dir

chr56:
Fix `extract_dir` of `uutils-coreutils@0.6.0`.

It looks likes that there is no version string in artifacts now.

See #7593 for detail.

- [ ] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

### DIFF
--- a/bucket/uutils-coreutils-lean.json
+++ b/bucket/uutils-coreutils-lean.json
@@ -34,15 +34,15 @@
         "architecture": {
             "64bit": {
                 "url": "https://github.com/uutils/coreutils/releases/download/$version/coreutils-$version-x86_64-pc-windows-msvc.zip",
-                "extract_dir": "coreutils-x86_64-pc-windows-msvc"
+                "extract_dir": "coreutils-$version-x86_64-pc-windows-msvc"
             },
             "32bit": {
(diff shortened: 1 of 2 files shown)


SorYoshino:
Since https://github.com/uutils/coreutils/pull/10676 has already been merged, future releases will revert to the `coreutils-$version-<arch>-pc-windows-msvc` naming convention.

I do not think it is wise to modify the existing `extract_dir` directly. It may be better to wait for a while, as the upstream project might update the assets for version 0.6.0 or publish a new release instead.

z-Fng:
> Since [uutils/coreutils#10676](https://github.com/uutils/coreutils/pull/10676) has already been merged, future releases will revert to the `coreutils-$version-<arch>-pc-windows-msvc` naming convention.
> 
> I do not think it is wise to modify the existing `extract_dir` directly. It may be better to wait for a while, as the upstream project might update the assets for version 0.6.0 or publish a new release instead.

I agree. However, it looks like version 0.6.0 won't be re-released. We can just modify `extract_dir` in the manifest directly, and revert the `extract_dir` handling in `autoupdate`.

z-Fng (approved):
Thanks for your contribution!

PR #7601, opened 2026-02-04
uutils-coreutils: Implement dynamic shim generation

SorYoshino:
### Summary

Restructures `uutils-coreutils` to use dynamic shim generation instead of a hard-coded `bin` list, significantly improving maintainability and ensuring all supported utilities are automatically shimmed.

### Related issues or pull requests

- Relates to #3482 
- Relates to #7303 
- Relates to #4415 
- Relates to #7454 

### Changes

- **Remove** the extensive, manually maintained list of shims from the `bin` field.
- **Implement** a `post_install` script that executes `coreutils.exe --list` to dynamically discover and create shims for all available utilities.
- **Implement** an `uninstaller` script to ensure all dynamically created shims are properly removed during uninstallation.
- **Switch** to `System.Diagnostics.ProcessStartInfo` within the scripts to reliably capture command output in a headless environment.

### Notes

- **Synchronization & Cleanliness**: This approach solves the "ghost shim" problem. If a future version of `uutils` adds or removes a utility, the dynamic script will only create shims for what actually exists in the binary.
- **Architecture Aware**: Different architectures (x64 vs. ARM64) might support different subsets of tools. By querying the binary at install time, we ensure the shims are accurate for the specific environment.
- **Maintainability**: Reduces the manifest size by ~90%, making it significantly easier to audit and maintain.
- Due to issues in certain logic within Scoop Core, errors may occur when updating or removing `[.shim` file.
  - See: https://github.com/ScoopInstaller/Scoop/issues/6592

### Testing

<details>

<summary>The test results are as follows:</summary> <br>

```powershell
┏[ ~]
└─> scoop install Unofficial/uutils-coreutils
Installing 'uutils-coreutils' (0.6.0) [64bit] from 'Unofficial' bucket
Loading coreutils-0.6.0-x86_64-pc-windows-msvc.zip from cache.
Checking hash of coreutils-0.6.0-x86_64-pc-windows-msvc.zip... OK.
Extracting coreutils-0.6.0-x86_64-pc-windows-msvc.zip... Done.
Linking D:\Software\Scoop\Local\apps\uutils-coreutils\current => D:\Software\Scoop\Local\apps\uutils-coreutils\0.6.0
Running post_install script... Creating shim for 'coreutils'.
Creating shim for '['.
Creating shim for 'arch'.
Creating shim for 'b2sum'.
Creating shim for 'base32'.
Creating shim for 'base64'.
Creating shim for 'basename'.
Creating shim for 'basenc'.
Creating shim for 'cat'.
Creating shim for 'cksum'.
Creating shim for 'comm'.
Creating shim for 'cp'.
Creating shim for 'csplit'.
Creating shim for 'cut'.
Creating shim for 'date'.
Creating shim for 'dd'.
Creating shim for 'df'.
Creating shim for 'dir'.
Creating shim for 'dircolors'.
Creating shim for 'dirname'.
Creating shim for 'du'.
Creating shim for 'echo'.
Creating shim for 'env'.
Creating shim for 'expand'.
Creating shim for 'expr'.
Creating shim for 'factor'.
Creating shim for 'false'.
Creating shim for 'fmt'.
Creating shim for 'fold'.
Creating shim for 'head'.
Creating shim for 'hostname'.
Creating shim for 'join'.
Creating shim for 'link'.
Creating shim for 'ln'.
Creating shim for 'ls'.
Creating shim for 'md5sum'.
Creating shim for 'mkdir'.
Creating shim for 'mktemp'.
Creating shim for 'more'.
Creating shim for 'mv'.
Creating shim for 'nl'.
Creating shim for 'nproc'.
Creating shim for 'numfmt'.
Creating shim for 'od'.
Creating shim for 'paste'.
Creating shim for 'pr'.
Creating shim for 'printenv'.
Creating shim for 'printf'.
Creating shim for 'ptx'.
Creating shim for 'pwd'.
Creating shim for 'readlink'.
Creating shim for 'realpath'.
Creating shim for 'rm'.
Creating shim for 'rmdir'.
Creating shim for 'seq'.
Creating shim for 'sha1sum'.
Creating shim for 'sha224sum'.
Creating shim for 'sha256sum'.
Creating shim for 'sha384sum'.
Creating shim for 'sha512sum'.
Creating shim for 'shred'.
Creating shim for 'shuf'.
Creating shim for 'sleep'.
Creating shim for 'sort'.
Creating shim for 'split'.
Creating shim for 'sum'.
Creating shim for 'sync'.
Creating shim for 'tac'.
Creating shim for 'tail'.
Creating shim for 'tee'.
Creating shim for 'test'.
Creating shim for 'touch'.
Creating shim for 'tr'.
Creating shim for 'true'.
Creating shim for 'truncate'.
Creating shim for 'tsort'.
Creating shim for 'uname'.
Creating shim for 'unexpand'.
Creating shim for 'uniq'.
Creating shim for 'unlink'.
Creating shim for 'vdir'.
Creating shim for 'wc'.
Creating shim for 'whoami'.
Creating shim for 'yes'.
Done.
'uutils-coreutils' (0.6.0) was installed successfully!
'uutils-coreutils' suggests installing 'extras/vcredist2022'.

┏[ ~]
└─> Get-Command -CommandType All | Where-Object { $_.Name -match '^\[' }

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     [.exe                                              0.0.0.0    D:\Software\Scoop\Local\shims\[.exe

┏[ ~]
└─> arch --version
arch (uutils coreutils) 0.6.0

```

</details>


bucket/uutils-coreutils.json
```diff
@@ -26,424 +26,36 @@
...
+    "post_install": [
+        "$process_info = [System.Diagnostics.ProcessStartInfo]@{",
+        "    FileName = \"$dir\\coreutils.exe\"; Arguments = '--list'",
+        "    CreateNoWindow = $true; UseShellExecute = $false; RedirectStandardOutput = $true",
+        "}",
+        "$process = [System.Diagnostics.Process]::Start($process_info)",
+        "$stdout_content = $process.StandardOutput.ReadToEnd()",
+        "$process.WaitForExit()",
+        "Write-Output \"`nCreating shim for 'coreutils'.\"",
+        "shim \"$dir\\coreutils.exe\" $global 'coreutils'",
```
z-Fng, line 38: Can we simply add `coreutils` to the `bin` field to generate the shims directly? I think dynamic generation is unnecessary for the `coreutils` shim.

bucket/uutils-coreutils.json
```diff
@@ -26,424 +26,36 @@
...
+        "$stdout_content -split \"`r?`n\" | Where-Object { $_ -match '^\\S' } | ForEach-Object {",
+        "    Write-Output \"Creating shim for '$($_.Trim())'.\"",
```
z-Fng, line 40: > Creating shim for '['.

I noticed there is a shim named `[`. Is this expected?

Image

z-Fng: > * Due to issues in certain logic within Scoop Core, errors may occur when updating or removing `[.shim` file.

Can we just skip creating this shim? A command named `[.exe` looks really weird in any shell.

bucket/uutils-coreutils.json
SorYoshino: Yes, but doing so would result in two stages of shim creation when installing or updating uutils-coreutils with Scoop: one before executing the `post_install` script and another within the `post_install` script. I think this would create a visual fragmentation of steps that serve the same purpose.

Additionally, the dynamic generation refers to the subcommands supported by `coreutils.exe`, rather than `coreutils.exe` itself. I believe that steps serving the same function should be grouped together, which is why I also placed them in the `post_install` script.

bucket/uutils-coreutils.json
SorYoshino: > I noticed there is a shim named `[`. Is this expected?

> Can we just skip creating this shim? A command named `[.exe` looks really weird in any shell.

Yes, because I only used `^\S` to exclude whitespace characters and did not exclude `[`.

To be honest, I am not certain whether a shim should be created for `[.exe`. As you mentioned, it appears weird in the Windows shell. In PowerShell, `[...]` is treated as a wildcard pattern, which makes `[.exe` unusable. However, in cmd, this executable can be used normally. I am unsure whether there are users who rely on this program, so I chose to keep it.

```batch
╭─ 03:36:06 |  25 Feb, Wednesday |   in  
╰─❯ [ 1 -lt 2 ]

╭─ 03:36:13 |  25 Feb, Wednesday |   in  
╰─❯ echo %errorlevel%
0

╭─ 03:36:17 |  25 Feb, Wednesday |   in  
╰─❯ [ 1 -lt 0 ]

╭─ 03:57:42 |  25 Feb, Wednesday |   in  
╰─❯ echo %errorlevel%
1

```

Regardless of whether a shim should be created for `[.exe`, there are undoubtedly issues in the related code within Scoop Core that, in my opinion, should be addressed.

bucket/uutils-coreutils.json
z-Fng: Fine by me.

bucket/uutils-coreutils.json
z-Fng: True, `rm_shim` has other bugs too that need fixing. It will probably take a while though, since there is currently only one active maintainer with write access reviewing Scoop-Core code (and it's not me).

SorYoshino: > True, `rm_shim` has other bugs too that need fixing. It will probably take a while though, since there is currently only one active maintainer with write access reviewing Scoop-Core code (and it's not me).

The problematic parts may not be limited to `rm_shim`. During my initial investigation, I found that several shim-related functions did not raise errors, but their use of the `Test-Path` function was inconsistent. In some cases the `-Path` parameter was explicitly specified, while in others it was not, which led to the issue. For the same functionality, the usage of functions or cmdlets should be consistent to facilitate troubleshooting and ensure fixes are complete.

In addition to `rm_shim`, both `scoop shim info` and `scoop which` exhibit issues. The former produces partial output but raises errors during execution, and some results may be missing due to incorrect information retrieval. The latter reports “WARN  '[' not found, not a scoop shim, or a broken shim.”, which clearly does not reflect the actual situation.

The observations above are based only on a preliminary investigation. The root cause of the issue reported in ScoopInstaller/Scoop#6592, as well as related problems, may be broader. Fortunately, issues of this type initially occur only during update and uninstall processes and do not affect the usability of the related shims. Unless users carefully inspect the process output, these problems are unlikely to be noticed.

z-Fng: Noted. I'll look into fixing this and see if we can expedite the next release.

PR #7612, opened 2026-02-06
sshpass: Add version 1.0.7

sitiom:
- Closes https://github.com/ScoopInstaller/Main/issues/7708

Windows version of http://sourceforge.net/projects/sshpass/

CC: @xhcoding

### DIFF
--- a/bucket/sshpass.json
+++ b/bucket/sshpass.json
@@ -12,6 +12,10 @@
     "bin": "sshpass.exe",
     "checkver": "github",
     "autoupdate": {
-        "url": "https://github.com/xhcoding/sshpass-win32/releases/download/v$version/sshpass.exe"
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/xhcoding/sshpass-win32/releases/download/v$version/sshpass.exe"
+            }
+        }
     }
 }


z-Fng:
Should we keep the original name `sshpass-win32`? Simply calling it `sshpass` might lead people to believe it's the same as the [original `sshpass` project](https://sourceforge.net/projects/sshpass/). Also, do you think it belongs in the 'extras' bucket instead?

https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket

sitiom:
> Should we keep the original name `sshpass-win32`? Simply calling it `sshpass` might lead people to believe it's the same as the [original `sshpass` project](https://sourceforge.net/projects/sshpass/).

I agree, but you can also argue the same for other manifests in this repo (example: [`cowsay`](https://github.com/ScoopInstaller/Main/blob/master/bucket/cowsay.json)). The description states that it's a "Windows version anyway" with the homepage pointing to the repo.

> Also, do you think it belongs in the 'extras' bucket instead?
> 
> [Wiki: Criteria for including apps in the main bucket (ScoopInstaller/Scoop)](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket)

Seeing that it's a command-line app, it definitely belongs in the Main bucket. When it comes to the GitHub stars criteria, I thought we were lenient on that 😅

z-Fng:
> When it comes to the GitHub stars criteria, I thought we were lenient on that 😅

I think we need to tighten up the star criteria. With our limited number of active maintainers, being too flexible will only lead to a PR backlog. What do you think?

Of course, there’s also a case to be made for adding this manifest to the main bucket, as its [download count](https://tooomm.github.io/github-release-stats/?username=xhcoding&repository=sshpass-win32) is actually quite decent.

Edit:
> I agree, but you can also argue the same for other manifests in this repo (example: [`cowsay`](https://github.com/ScoopInstaller/Main/blob/master/bucket/cowsay.json)). The description states that it's a "Windows version anyway" with the homepage pointing to the repo.

Keeping the GitHub repository name consistent with the package name makes it easier for users to search. Additionally, if similar-named projects appear -- such as `sshpass-win64`(https://github.com/ScoopInstaller/Extras/pull/17310) --there will be no confusion.

sitiom:
I agree we should have at least a minimum star count and be reputable at least (which we already do?). I've tested `sshpass-win64` and it doesn't even work 😂

PR #7650, opened 2026-02-17
libjxl@0.11.2: Fix binary paths by adding `extract_dir`

### LINKED ISSUE #7649: [Bug]: libjxl@0.11.2: binaries has been moved to bin subfolder
### Prerequisites

### Package Name

libjxl

Expected: the update should install the new version and create shims for the executables.
Current: The shims are not created because the executables are not found (they are now in /bin subfolder) and the installation fails.

### Steps to Reproduce

```console
`scoop install libjxl` or `scoop update libjxl`
```

### Possible Solution

Change the manifest (PR included).

### Scoop and Buckets Version

```console
Current Scoop version:
b588a06e (HEAD -> master, tag: v0.5.3, origin/master, origin/HEAD) chore(release): Bump to version 0.5.3 (resync) (#6436)

'main' bucket:
f84c47eaa4 (HEAD -> master, origin/master, origin/HEAD) ytt: Update to version 0.53.1

'extras' bucket:
a3b0c921e6 (HEAD -> master, origin/master, origin/HEAD) zed: Update to version 0.223.4
```

### Scoop Config

```console
last_update                     : 17/02/2026 16:19:47
scoop_repo                      : https://github.com/ScoopInstaller/Scoop
scoop_branch                    : master
aria2-retry-wait                : 1
aria2-max-connection-per-server : 10
aria2-min-split-size            : 1M
aria2-options                   : --max-concurrent-downloads=10
aria2-warning-enabled           : False
use_sqlite_cache                : True
debug                           : False
```

### PowerShell Version

```console
PSVersion                      7.5.4
GitCommitId                    7.5.4
OS                             Microsoft Windows 10.0.26200
```

### Additional Softwares

_No response_

frabera:
Updated paths to binaries (bin subfolder).

Closes #7649 


eugenesvk:
This way you remove some potentially useful files like `jxl_jni.dll` that were previously located in the main folder. I just happened to also have added a fix, but retaining all the files, see https://github.com/ScoopInstaller/Main/pull/7652

frabera:
> This way you remove some potentially useful files like `jxl_jni.dll` that were previously located in the main folder. I just happened to also have added a fix, but retaining all the files, see #7652

I initially used the same approach as you but I checked and scoop actually downloads the windows-static version, that means that the libraries are not used by the binaries, and they are provided in a different folder (lib) to distribute them.

eugenesvk:
they are not used by the binaries, but they can be used by other apps. For example, you could use a newer version from scoop to replace a dll library some GUI app uses since those app can be rather slow to update their dependencies.
Or maybe there is even some other scoop app that already relies on the library (though don't know if it exists)

frabera:
It seems to be out of scope to me, since they also selected the static variant to have more portable executables, anyway the size difference with or without libraries is not so large.

BlackWinnerYoshi:
Please add `icc_simplify.exe` to `bin`

frabera:
> Please add `icc_simplify.exe` to `bin`

Done, thanks.

z-Fng:
Closing in favor of #7652. Thanks for your contribution!

eugenesvk:
> that means that the libraries are not used by the binaries, and they are provided in a different folder (lib) to distribute them

FYI i've created a separate manifest only for the binaries, without the lib part, which aren't really needed for most of the users

https://github.com/ScoopInstaller/Main/pull/8365

PR #7653, opened 2026-02-18
uv: Persist user data

### LINKED ISSUE #6955: [request] change UV install to be in scoop dir
the python utility UV has a "tools" feature so
i am requesting that a pr is made to add a pre_install script to set UV_INSTALL_DIR to $persist_dir and add it to path after scoop
as well as a post_uninstall to remove UV_INSTALL_DIR from env and to remove it from path

i am sorry about not creating a pr for this as i dont fully know what im doing

### LINKED ISSUE #8053: [Notice] Scoop Data Persistence for uv v0.11.16: Manual Migration Required
As of uv v0.11.16, Scoop persists uv's data, so users may need to migrate the data manually.

For more information, please refer to: #7653.

ghyatzo:

Hey,

This might be a bit opinionated, but I wanted uv to behave like pyenv (when installed through scoop) behaved.
With this change scoops makes uv store its python and tool installation paths within the scoop uv folder.

I currently have been using this manifest for the past year or so without issues regarding uv.

It also automatically sets the python and tool shims folders in the path so that it's automatically available.

If too breaking/opinionated feel free to close, but wanted to give the option if it was ever considered.

Cheers.

Closes #6955


z-Fng:
Should we also migrate the artifacts originally located in `$Env:APPDATA/uv` to `$persist_dir`?

ghyatzo:
According to https://docs.astral.sh/uv/reference/storage/#tool-executables

I think this manifests already does that by setting all relevant env variables and then save those same folders under `persist` key.

or maybe I am missing something?

z-Fng:
I'm referring to users who have previously installed UV via scoop. Their data resides in `$Env:APPDATA/uv`. After the update, they will 'lose' this data. Should we migrate the data for them, or will UV migrate it automatically?

https://github.com/ScoopInstaller/Main/blob/82c18419590eb377daf7dfb29a4cb52e0713f112/bucket/bun.json#L26-L33

ghyatzo:
Ah that is a very good point.
I believe that uv will redownload the missing python versions on its own the first time that will be needed again. But I don't think it will "reinstall" all the tools's shims automatically. Will need to double check that.

ghyatzo:
I did some quick tests and uv seems to just pick up whatever is in the target folder. Copying over the legacy location should be a pretty transparent operation for the user.
I'll adapt the bun excerpt shared above to uv as well in the following days

ghyatzo:
ok, first attempt.
It's more complicated than bun since by default uv is spread around and dumps all shims inside ~/.local/bin.
Ok that on windows that is probably not used much but it might still contain stuff not related to uv.

here's the overview:
- only automigrate if the persistdir is not present already
- move all python versions to persist dir
- move all executables in ~/.local/bin that match python.exe or pythonX.Y.exe into the shim folder
- move all tools to the persist dir and build a list of names
- move `mytool.exe` only if there is a folder named `mytool` in the tools install dir.

this should only move previously handled by uv. here's the code in plain:
```powershell
# Migrate legacy UV data if it exists and the new persist directory doesn't exist
$legacy_uv = Join-Path -Path $env:APPDATA -ChildPath 'uv'

if ((Test-Path $legacy_uv) -and -not (Test-Path $persist_dir)) {
    Write-Host "Migrating legacy uv data from '$legacy_uv' to '$persist_dir'." -ForegroundColor Yellow

    # --- Legacy locations ---
    $legacy_uv_python_data  = Join-Path -Path $legacy_uv -ChildPath 'python'
    $legacy_uv_python_shims = Join-Path (Join-Path $env:USERPROFILE '.local') 'bin'
    $legacy_uv_tools        = Join-Path -Path $legacy_uv -ChildPath 'tools'
    $legacy_uv_tools_shims  = Join-Path (Join-Path $env:USERPROFILE '.local') 'bin'

    # --- Persist target layout ---
    $persist_python_root     = Join-Path $persist_dir 'python'
    $persist_python_versions = Join-Path $persist_python_root 'versions'
    $persist_python_shims    = Join-Path $persist_python_root 'shims'

    $persist_tools_root      = Join-Path $persist_dir 'tools'
    $persist_tools_versions  = Join-Path $persist_tools_root 'versions'
    $persist_tools_shims     = Join-Path $persist_tools_root 'shims'

    # Create target directories
    $null = New-Item -ItemType Directory -Force -Path @(
        $persist_dir,
        $persist_python_root, $persist_python_versions, $persist_python_shims,
        $persist_tools_root,  $persist_tools_versions,  $persist_tools_shims
    )

    # ---- Move Python "versions" content ----
    if (Test-Path $legacy_uv_python_data) {
        Write-Host "-> Moving Python versions from '$legacy_uv_python_data' to '$persist_python_versions'..." -ForegroundColor Cyan
        Get-ChildItem -Path $legacy_uv_python_data -Force |
            ForEach-Object {
                Move-Item -LiteralPath $_.FullName -Destination $persist_python_versions -Force -ErrorAction Stop
            }
    } else {
        Write-Host "-> No legacy Python versions found at '$legacy_uv_python_data'." -ForegroundColor DarkGray
    }

    # ---- Move Python shims: only python.exe or python<major>.<minor>.exe ----
    if (Test-Path $legacy_uv_python_shims) {
        $pythonShimRegex = '^(python(\d+\.\d+)?)\.exe$'  # matches python.exe and pythonX.Y.exe
        $pythonShims = Get-ChildItem -Path $legacy_uv_python_shims -File -Force |
                       Where-Object { $_.Name -match $pythonShimRegex }

        if ($pythonShims) {
            Write-Host "-> Moving Python shims to '$persist_python_shims': $($pythonShims.Name -join ', ')" -ForegroundColor Cyan
            foreach ($shim in $pythonShims) {
                Move-Item -LiteralPath $shim.FullName -Destination (Join-Path $persist_python_shims $shim.Name) -Force -ErrorAction Stop
            }
        } else {
            Write-Host "-> No valid Python shims found in '$legacy_uv_python_shims'." -ForegroundColor DarkGray
        }
    } else {
        Write-Host "-> Legacy Python shim directory not found at '$legacy_uv_python_shims'." -ForegroundColor DarkGray
    }

    # ---- Move Tools "versions" content ----
    if (Test-Path $legacy_uv_tools) {
        Write-Host "-> Moving Tools versions from '$legacy_uv_tools' to '$persist_tools_versions'..." -ForegroundColor Cyan
        Get-ChildItem -Path $legacy_uv_tools -Force |
            ForEach-Object {
                Move-Item -LiteralPath $_.FullName -Destination $persist_tools_versions -Force -ErrorAction Stop
            }
    } else {
        Write-Host "-> No legacy Tools found at '$legacy_uv_tools'." -ForegroundColor DarkGray
    }

    # ---- Move Tools shims: only those matching folder names under /tools/versions ----
    # Build allowlist from folder names directly under $persist_tools_versions
    $toolNames = @()
    if (Test-Path $persist_tools_versions) {
        $toolNames = Get-ChildItem -Path $persist_tools_versions -Directory -Force |
                     Select-Object -ExpandProperty Name
    }

    if ($toolNames.Count -gt 0 -and (Test-Path $legacy_uv_tools_shims)) {
        # For each tool name, move "<tool>.exe" if present
        $moved = @()
        foreach ($tool in $toolNames) {
            $candidate = Join-Path $legacy_uv_tools_shims ($tool + '.exe')
            if (Test-Path $candidate) {
                Move-Item -LiteralPath $candidate -Destination (Join-Path $persist_tools_shims ($tool + '.exe')) -Force -ErrorAction Stop
                $moved += ($tool + '.exe')
            }
        }

        if ($moved.Count -gt 0) {
            Write-Host "-> Moved tool shims to '$persist_tools_shims': $($moved -join ', ')" -ForegroundColor Cyan
        } else {
            Write-Host "-> No matching tool shims found in '$legacy_uv_tools_shims' for tools: $($toolNames -join ', ')." -ForegroundColor DarkGray
        }
    } else {
        if (-not (Test-Path $legacy_uv_tools_shims)) {
            Write-Host "-> Legacy tools shim directory not found at '$legacy_uv_tools_shims'." -ForegroundColor DarkGray
        } else {
            Write-Host "-> No tool folders found under '$persist_tools_versions'; skipping shim migration." -ForegroundColor DarkGray
        }
    }
} elseif ((Test-Path $legacy_uv) -and (Test-Path $persist_dir)) {
    Write-Host "Found both '$legacy_uv' and '$persist_dir'. Skipping auto-migration; please migrate manually if needed." -ForegroundColor Yellow
}
```

ghyatzo:
Bump

ghyatzo:
Gentle bump

ofek:
It's worth noting that I [asked](https://discord.com/channels/1039017663004942429/1207998321562619954/1503812345800032357) in their Discord server and a core maintainer expressed general disapproval.

> Hello team! Do any uv maintainers or users have opinions on the location of persisted data when installed via a package manager? I noticed [this](https://github.com/ScoopInstaller/Main/pull/7653) PR that updates the Scoop manifest and was wondering what other package managers like Homebrew do.
>
> ---
>
> we prefer the default layout as it's well-documented and we recognize it when users have bug reports. anything custom may lead to confusion

ofek:
I personally would prefer the behavior implemented by this PR however I think the defaults should not be changed. Is there a way to make this behavior opt-in?

z-Fng:
> ok, first attempt. It's more complicated than bun since by default uv is spread around and dumps all shims inside ~/.local/bin. Ok that on windows that is probably not used much but it might still contain stuff not related to uv.
> 
> here's the overview:
> 
> * only automigrate if the persistdir is not present already
> * move all python versions to persist dir
> * move all executables in ~/.local/bin that match python.exe or pythonX.Y.exe into the shim folder
> * move all tools to the persist dir and build a list of names
> * move `mytool.exe` only if there is a folder named `mytool` in the tools install dir.

Thank you for your work. I hadn't realized the migration would be this complicated. Given the [complexity](https://github.com/ScoopInstaller/Scoop/wiki/Criteria-for-including-apps-in-the-main-bucket), I think it's better to just explain this to users directly in the notes.

> Is there a way to make this behavior opt-in?

Not yet.

z-Fng (approved):
Thanks for your contribution!

cderv:
I've just been hit by this change. I can confirm this is quite surprising because you do a usual `scoop update` and then after that `uv` is broken. `uv tool list` finds nothing and even `uv python dir` has changed. 

Indeed, anyone can work around it by reinstalling things but that is a bit unexpected. 

What is the exact rationale of having python versions and tools installed in a scoop path instead of default ? I may have missed something. 

@ofek reply (https://github.com/ScoopInstaller/Main/pull/7653#issuecomment-4435495689) was interesting about uv's maintainers take.

z-Fng:
> If too breaking/opinionated feel free to close, but wanted to give the option if it was ever considered.

> What is the exact rationale of having python versions and tools installed in a scoop path instead of default ?

This change is indeed a very significant breaking change. However, it aligns with Scoop's design philosophy. Scoop expects packages to be portable and, ideally, store cross-update data in a [designated directory](https://github.com/ScoopInstaller/Scoop/wiki/Persistent-data). This ensures that Scoop's per-user installations, global installations, and purge uninstallations work properly. In fact, many other packages have made similar improvements, but we have implemented smooth data migration in their manifests, so users may not even notice. uv's data migration is a bit more complex, which is why we didn't do it — and that admittedly makes it somewhat confusing for users.

PR #7697, opened 2026-02-28
bun: Persist user data

A-Words:
Closes #7696 

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md) 

### Summary
Fix `bun` manifest so Bun data is persisted under Scoop `persist` directory, and add safe migration for legacy `%USERPROFILE%\.bun`.

### Changes
- Add `env_set`:
  - `BUN_INSTALL=$persist_dir`
- Add `env_add_path`:
  - `"bin"`
- Add `persist`:
  - `"bin"`, `"install"`
- Extend `pre_install`:
  - migrate `%USERPROFILE%\.bun` to `$persist_dir` when persist dir does not exist
  - if both paths exist, skip migration and print warning to avoid overwrite
- Keep existing AVX2 selection / download / hash / autoupdate logic unchanged

### Validation
- `bucket/bun.json` parses successfully (`ConvertFrom-Json`)
- `scoop cat .\bucket\bun.json` shows expected new fields
- `bin/test.ps1` could not be executed locally due to missing module: `BuildHelpers`

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added support for ARM64 architecture on Windows
  * Introduced automatic CPU feature detection for optimal binary selection
  * Implemented legacy data migration from previous installation paths
  * Added persistent storage for installation data across sessions

* **Chores**
  * Updated environment configuration and installation process

PR #7769, opened 2026-03-22
bun: Use persisted global bin path

A-Words:
Relates to #7697

## Summary

Address the feedback in https://github.com/ScoopInstaller/Main/pull/7697#issuecomment-4079046852 by aligning Bun's global bin path with the real path Bun reports and checks.

Bun resolves Scoop's junctioned `current\bin` path to the persisted directory, so pointing `BUN_INSTALL` at `current` does not avoid the `not in $PATH` warning. This change instead configures Bun's global install paths explicitly under Scoop's persist directory and adds that real bin path to PATH.

## Changes

- set `BUN_INSTALL` to `$persist_dir`
- add `BUN_INSTALL_BIN` and `BUN_INSTALL_GLOBAL_DIR`
- replace `env_add_path` with explicit PATH management in `installer` / `uninstaller`
- remove stale `current\bin` / versioned `bin` PATH entries during install and uninstall
- keep existing `persist` entries and legacy `~/.bun` migration logic unchanged

## Testing

- `Test-Json -Path .\bucket\bun.json`
- `scoop uninstall bun`
- `scoop install .\bucket\bun.json`
- verified user PATH changed from `E:\Scoop\apps\bun\current\bin` to `E:\Scoop\persist\bun\bin`
- verified `bun pm bin -g` returns `E:\Scoop\persist\bun\bin` without the `not in $PATH` warning
- verified persisted global commands still resolve and run:
  - `figlet hello`

## Notes

This keeps Scoop persistence intact while matching Bun's actual path resolution behavior on Windows, where Bun resolves the junctioned `current\bin` path to the real persisted directory.

### DIFF
--- a/bucket/bun.json
+++ b/bucket/bun.json
@@ -49,9 +49,7 @@
         "Remove-Item \"$dir\\bun-windows-*\" -Recurse"
     ],
     "installer": {
-        "script": [
-            "Add-Path -Path \"$persist_dir\\bin\" -Global:$global -Force"
-        ]
+        "script": "Add-Path -Path \"$persist_dir\\bin\" -Global:$global -Force"
     },
     "env_set": {
         "BUN_INSTALL": "$persist_dir",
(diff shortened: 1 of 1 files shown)


z-Fng:
> * remove stale `current\bin` / versioned `bin` PATH entries during install and uninstall

If these weren't created by Scoop, cleaning them up shouldn't fall under Scoop's responsibility. If you are concerned about the `env_add_path` previously set in the manifest, it will be automatically removed during the update or uninstallation.

PR #7772, opened 2026-03-23
pipx: Use py launcher as fallback when python not found

gaborbernat:
Installing pipx via scoop fails on Windows systems where `python.exe` is not on PATH but the Python Launcher (`py.exe`) is available. This is the default state after a standard python.org installation that doesn't modify PATH. Users hit `'python' is not recognized` immediately after `scoop install pipx`, with no workaround short of manually editing `pipx.bat`. 🐛

The `pre_install` script now checks for `python` first, then falls back to `py`, using `Get-Command` to detect availability. If neither executable is found, the install throws a clear error directing the user to install Python or add it to PATH. This keeps `python` as the preferred interpreter for conda/mamba users while supporting the Python Launcher that ships with every python.org install.

Fixes pypa/pipx#1260

### DIFF
--- a/bucket/1password-cli.json
+++ b/bucket/1password-cli.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.33.0",
+    "version": "2.33.1",
     "description": "The 1Password command-line tool makes your 1Password account accessible entirely from the command line.",
     "homepage": "https://developer.1password.com/docs/cli",
     "license": {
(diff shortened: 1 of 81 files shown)


sitiom (commented):
Note that the `py` launcher will be deprecated in favor of the [Python Installation Manager](https://peps.python.org/pep-0773/):

> The existing `.exe` installer and `py` launcher will be deprecated and no longer released from two years after this PEP is accepted. The existing embeddable distro will no longer be listed as a download, but will be available through the install manager. The Windows Store app will be immediately replaced with the install manager.

gaborbernat:
> Note that the `py` launcher will be deprecated in favor of the [Python Installation Manager](https://peps.python.org/pep-0773/):

Sounds like a good follow-up but we're good as is today.

PR #7894, opened 2026-04-25
probe-rs: Add shim for `cargo-embed` & `cargo-flash`

xavierfortier1:
cargo-flash and cargo-embed can no longer be installed with `cargo install ...` and are now provided with the probe-rs package directly. Both needs to be added in the bin array.


z-Fng (approved):
Thanks for your contribution!

PR #8040, opened 2026-05-24
g: Persist user data

scillidan:
- Modified the installation method of environment variables, referenced [rustup.json](https://github.com/ScoopInstaller/Main/blob/master/bucket/rustup.json)
- Modified the contents of environment variables, referenced https://github.com/voidint/g#faq
- Add notes, persist


bucket/g.json
```diff
@@ -3,6 +3,10 @@
     "description": "A command-line tool that provides convenient management and switching of multiple versions of the Go environment.",
     "homepage": "https://github.com/voidint/g",
     "license": "MIT",
+    "notes": [
+        "Please restart your ternimal for environments to take effect.",
+        "Please add GOPATH\\bin into PATH."
```
z-Fng: Only the CMD terminal requires a restart to update environment variables -- PowerShell doesn't. I don't think we need to mention restarting.

z-Fng: Instead, what we need to remind users is that this change may require manual reinstallation or data migration.

https://github.com/ScoopInstaller/Main/blob/2b938d669d6629d75f596c72fc9615d1f5dfbc1e/bucket/uv.json#L6-L10

z-Fng (changes requested):

bucket/g.json
z-Fng: - https://github.com/ScoopInstaller/Main/pull/7653

z-Fng (approved):
Thank you for your contribution!

scillidan:
Thanks for your review!

Issue #6457, opened 2025-01-15
mosh-client is outdated


petersjt014:
The version of [mosh](https://mosh.org/) in the main bucket is out of date (1.3.2, latest is 1.4.0).

There are also two fields in [the manifest](https://github.com/ScoopInstaller/Main/blob/master/bucket/mosh-client.json) that are of mild concern: the [homepage field](jumptrading/mosh-windows-wrappers) leads to a 404'd repo, and the [url fields](https://github.com/felixse/FluentTerminal) to a project that has not seen updates in some time. Not sure that 2nd concern is that bad (mosh itself is also on a rather slow release schedule, after all), but it might be worth exploring other options to find a 'cleaner' solution, if possible.

Ra2-IFV:
Looks like the author of this manifest followed this comment: https://github.com/mobile-shell/mosh/issues/293#issuecomment-885484861 

> For the mosh-client, native Windows builds are available - https://github.com/jumptrading/mosh-windows-wrappers
> These are used by Fluent Terminal. Binaries can be found here - https://github.com/felixse/FluentTerminal/tree/master/Dependencies/MoshExecutables

And on mosh website:
> There is no "native" mosh executable for Windows available at this time. The [Chrome version of Mosh](https://chrome.google.com/webstore/detail/mosh/ooiklbnjmhbcgemelgfhaeaocllobloj) is the easiest way to use mosh on Windows.

HUMORCE:
Upstream has dropped the project(mosh-windows-wrappers).

Ra2-IFV:
Drop that manifest?

HUMORCE:
> Drop that manifest?

When FluentTerminal repo no longer provides its binaries.

PR #8055, opened 2026-05-25
mosh-client: Deprecate manifest

NECOtype:
Relates to #6457

Upstream dropped the project for years (repo is gone).
FluentTerminal is no longer maintained: https://github.com/felixse/FluentTerminal/commit/ba83ec485eea8d8ee65825711731300bee2e76ed


z-Fng:
https://github.com/ScoopInstaller/Main/issues/6457#issuecomment-2822910060

> > Drop that manifest?
> 
> When FluentTerminal repo **no longer provides its binaries**.

We generally do not deprecate a manifest while the download link is still available.

NECOtype:
> [#6457 (comment)](https://github.com/ScoopInstaller/Main/issues/6457#issuecomment-2822910060)
> 
> > > Drop that manifest?
> > 
> > 
> > When FluentTerminal repo **no longer provides its binaries**.
> 
> We generally do not deprecate a manifest while the download link is still available.

Alright. Since [glide](https://github.com/ScoopInstaller/Main/blob/master/deprecated/glide.json) was deprecated but binaries were still downloadable, I thought this should be deprecated as well.

PR #8109, opened 2026-06-05
supabase: Align manifest with official scoop-bucket

avallete:
- Updates description/homepage metadata to match the official bucket.
- Align description with supabase owned bucket description: https://github.com/supabase/scoop-bucket/blob/main/supabase.json#L4
  Supabase doesn't use the "firebase open-source alternative" as a description for their cli tool anymore.
- Switches download artifacts from .tar.gz to .zip with versioned filenames.
- Fix autoupdate hash extraction.

### DIFF
--- a/bucket/supabase.json
+++ b/bucket/supabase.json
@@ -1,31 +1,33 @@
 {
     "version": "2.105.0",
     "description": "Supabase CLI",
-    "homepage": "https://supabase.com",
+    "homepage": "https://supabase.com/",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_2.105.0_windows_amd64.zip",
-            "hash": "c453ba5573fb1dffdd13e609af436e69be12435040adf30b0a2195f51a1dd046",
-            "bin": ["supabase.exe"]
+            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_windows_amd64.tar.gz",
+            "hash": "f4c11bb8b3c34fcf5ab6dd41cb1e9403416c7c319d668f24b3769f004bd24824"
         },
         "arm64": {
-            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_2.105.0_windows_arm64.zip",
-            "hash": "d2c277a65d8b68d5bf2bc0233ec0bd3e5beab6d3a4798ed7fa119113bc02412e",
-            "bin": ["supabase.exe"]
+            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_windows_arm64.tar.gz",
+            "hash": "676118fb1d5b48cc9b88f06fac87bbc261bc910ce91f7e235c12854644857929"
         }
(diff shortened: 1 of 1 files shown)


bucket/supabase.json
```diff
@@ -1,33 +1,31 @@
...
+    "description": "Supabase CLI",
+    "homepage": "https://supabase.com",
     "license": "MIT",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_windows_amd64.tar.gz",
-            "hash": "f4c11bb8b3c34fcf5ab6dd41cb1e9403416c7c319d668f24b3769f004bd24824"
+            "url": "https://github.com/supabase/cli/releases/download/v2.105.0/supabase_2.105.0_windows_amd64.zip",
+            "hash": "c453ba5573fb1dffdd13e609af436e69be12435040adf30b0a2195f51a1dd046",
+            "bin": ["supabase.exe"]
```
z-Fng: > * Aligns `bucket/supabase.json` with the official [supabase/scoop-bucket](https://github.com/supabase/scoop-bucket/blob/main/supabase.json) manifest

This manifest probably won't pass our CI. You can try submitting a PR to make changes in their scoop bucket instead, or suggest that they don't need to maintain a duplicate manifest.

z-Fng:
~~Closing as per comment https://github.com/ScoopInstaller/Main/pull/8109#discussion_r3363536091.~~

bucket/supabase.json
avallete: Thank's for the quick response, we do release a "beta" package which is why we keep our own bucket available for users that want's to opt-in to that.

I'll see to update our own scoop bucket to match the format of this one to reduce the drift.

However, supabase doesn't present itself as "firebase alternative" anymore, would it be okay for me to draft another PR just changing the description of the package matching the one we now own ("Supabase CLI") ?

avallete: Re-opened the pull request, with only the description field update scope.

Also updated the description.

z-Fng: Ah, I see. I've reopened this PR.

z-Fng: > we do release a "beta" package which is why we keep our own bucket available for users that want's to opt-in to that.

- https://scoop.sh/#/apps?q=supabase&o=true

I noticed there's a `supabase-beta` in the versions bucket. I'll update that manifest and sync the changes over shortly.

z-Fng: It seems like `checksum.txt` doesn't include the hash for the `supabase_windows_amd64.tar.gz` file. I've kept the following change:
- Updates homepage metadata to match the official bucket
- Switches download artifacts from .tar.gz to .zip with versioned filenames.

z-Fng (approved):

bucket/supabase.json
z-Fng: > I noticed there's a `supabase-beta` in the versions bucket. I'll update that manifest and sync the changes over shortly.

https://github.com/ScoopInstaller/Versions/commit/590ecbe055b47f47bb2740a85ad28444876b171a

PR #8107, opened 2026-06-05
microsoft-coreutils(-lean): Add version 2026.5.29

z-Fng:
- Closes #8098
- Relates to #3482
- Relates to https://github.com/microsoft/coreutils/issues/71

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #8110, opened 2026-06-05
microsoft-coreutils@2026.5.29: Adding Scripts, Remake shims

SiriosDev:
Following the *raw* script included in the original repository,  I created a script that creates hard links in `bin` and `cmd`, removes the placeholder file, and modifies and runs `pwsh-install.ps1` to reference `HKCU` instead of `HKLM`, so that `UAC` is no longer required even for `CurrentUser` mode. (And Uninstall too of course).

In this way, the shipped “patch” should apply to all `$PROFILE`, except for legacy versions (>5.0), which are not supported by the original installer

Relates to #8107 


bucket/microsoft-coreutils.json
```diff
@@ -21,391 +20,346 @@
             "uutils"
         ],
         [
-            "coreutils.exe",
-            "arch",
+            "cmd/arch.cmd",
             "arch"
```
z-Fng: This doesn't seem to work.
```
> arch
MZ�$�
'MZ�$�' is not recognized as an internal or external command,
operable program or batch file.

> D:\Scoop\shims\arch.cmd
MZ�$�
'MZ�$�' is not recognized as an internal or external command,
operable program or batch file.
```
I'd suggest waiting for the next portable release. See: https://github.com/microsoft/coreutils/issues/18

SiriosDev: Try rerunning PowerShell or opening another instance directly, otherwise the _new_ `$PROFILE` is not loaded

z-Fng: If it can only be used in PowerShell, then there's no need to create a shim.

> I'd suggest waiting for the next portable release. See: [microsoft/coreutils#18](https://github.com/microsoft/coreutils/issues/18)

SiriosDev: > If it can only be used in PowerShell, then there's no need to create a shim.

As far as I understand, only pwsh requires these specific tweaks; cmd shouldn't, assuming I've got it right:
https://github.com/microsoft/coreutils#windows-caveats

z-Fng: Isn't it confusing to create a shim that doesn't actually work? I think there might be a misunderstanding regarding how Scoop shims work. Please refer to:
- https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties
  > You can also create an alias shim which uses a different name to the real executable and (optionally) passes arguments to the executable. Instead of just using a string for the executable, use e.g: [ "program.exe", "alias", "--arguments" ]. See [busybox](https://github.com/ScoopInstaller/Main/blob/master/bucket/busybox.json) for an example.
- https://github.com/ScoopInstaller/Main/pull/3482
  > [ "coreutils.exe", "arch", "arch" ]

  For example, creating an `arch` shim creates an alias for `coreutils.exe arch`.

SiriosDev: oh right, now should work
I thought about shimming the cmd files to follow the script's instructions, but it turns out that's not necessary because the script references the executables directly in the cmd subfolder

deuteros-gex:
Just for your infomartion, the latest build is now available with zip archives.
I think it is time to revise this manifest?
https://github.com/microsoft/coreutils/releases/download/v2026.6.16/coreutils-2026.6.16-x64.zip
https://github.com/microsoft/coreutils/releases/download/v2026.6.16/coreutils-2026.6.16-arm64.zip

PR #8271, opened 2026-07-07
eac3to: Persist user data

emotion3459:
# Summary
Persists the eac3to.ini config file, so user config is kept across versions (note: we don't need to create the file, a default config is provided with all downloads).
Also persists "error.wav" and "success.wav". These audio files are extremely loud (potentially harmful, even) and have no volume control, so I think it's essential for people to be able to replace these with dummy files and not have them be restored upon updating.
I would even argue the pre_install script should just delete these audio files, but I don't know what the "policy" is for "modifying" applications like that.


emotion3459:
The checks failing are unrelated to this PR.

z-Fng (approved):

PR #8296, opened 2026-07-12
innounp: Update to version 2.67.10, enable checkver

z-Fng:
> A bug was introduced in innounp 2.67.10 that causes all extractions to fail. This PR rolls it back to version 2.67.9 ([0e86906](https://github.com/ScoopInstaller/Main/commit/0e86906e443f1953558f2032e5b50f4d786fdf7f)) and disables checkver.

v2.67.10 has been re-released with the bug fixed.

Relates to #8294

### DIFF
--- a/bucket/innounp.json
+++ b/bucket/innounp.json
@@ -1,11 +1,15 @@
 {
-    "version": "2.67.9",
+    "version": "2.67.10",
     "description": "Inno Setup Unpacker (Unicode)",
     "homepage": "https://www.rathlev-home.de/tools/prog-e.html#unpack",
     "license": "GPL-3.0-only",
-    "url": "https://github.com/jrathlev/InnoUnpacker-Windows-GUI/releases/download/ui_2_2_9/innounp-2.zip",
-    "hash": "1439f8d9e24b19e7d0b31b9c427ba4533387522a370c39280f17d3371eb7febf",
+    "url": "https://raw.githubusercontent.com/jrathlev/InnoUnpacker-Windows-GUI/refs/heads/master/innounp-2/bin/innounp-2.zip",
+    "hash": "5b96dbc57e7c87961366233d823fb471316b1eb6119e4914ced807d52c2b6f7d",
     "bin": "innounp.exe",
+    "checkver": {
+        "url": "https://raw.githubusercontent.com/jrathlev/InnoUnpacker-Windows-GUI/refs/heads/master/innounp-2/docs/innounp.htm",
+        "regex": "(?s)What's new/History.*?<b>(\\d+(?:\\.\\d+){2})</b>"
+    },
     "autoupdate": {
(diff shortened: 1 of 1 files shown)


jans-code:
Isn't this the bad Version where extractions fail?

z-Fng:
> Isn't this the bad Version where extractions fail?

v2.67.10 has been re-released. If you check the hashes, you'll see they don't match.

- https://github.com/ScoopInstaller/Main/commit/0e86906e443f1953558f2032e5b50f4d786fdf7f: `67a1ff8a8c1fccbe418cfeef85e2d1eaaf67b0c2541b37f19356b1749a146e7d`
- https://github.com/ScoopInstaller/Main/commit/e175bdc0113e0bed96e944612000b5e493485c39: `5b96dbc57e7c87961366233d823fb471316b1eb6119e4914ced807d52c2b6f7d`

If you installed the buggy v2.67.10, run the following command to force update.
```
scoop update
scoop cache rm innounp
scoop update innounp --force
```

bucket/innounp.json
```diff
@@ -1,11 +1,15 @@
 {
-    "version": "2.67.9",
+    "version": "2.67.10",
     "description": "Inno Setup Unpacker (Unicode)",
     "homepage": "https://www.rathlev-home.de/tools/prog-e.html#unpack",
     "license": "GPL-3.0-only",
-    "url": "https://github.com/jrathlev/InnoUnpacker-Windows-GUI/releases/download/ui_2_2_9/innounp-2.zip",
-    "hash": "1439f8d9e24b19e7d0b31b9c427ba4533387522a370c39280f17d3371eb7febf",
+    "url": "https://raw.githubusercontent.com/jrathlev/InnoUnpacker-Windows-GUI/refs/heads/master/innounp-2/bin/innounp-2.zip",
```
chadlwilson, line 6: It’s not correct to use mutable references to git branches like this on GitHub - need to use URLs pinned to commits or specific blobs that cannot be changed if you’re trying to pin to a version that will reliably match a hash.

The package is completely broken right now.

z-Fng: > It’s not correct to use mutable references to git branches like this on GitHub - need to use URLs pinned to commits or specific blobs that cannot be changed if you’re trying to pin to a version that will reliably match a hash.

The download URL change in this PR simply reverts [#8294](https://github.com/ScoopInstaller/Main/pull/8294/changes) and introduces no other changes. For various reasons, the [previous](https://github.com/ScoopInstaller/Main/commits/master/bucket/innounp.json) download links were not version-specific. There are many reasons for this - for example, the upstream release binary for the same version could change, and previously there wasn't even a GitHub release...

You're completely right that having a stable download URL point to stable artifacts is crucial. I've actually been meaning to reach out to the upstream maintainer to see if they can publish releases via GitHub releases, but I haven't had much time yet. I'll follow up on those changes later.

> The package is completely broken right now.

Meanwhile, this issue has been fixed by #8327.

PR #8314, opened 2026-07-15
go-swagger: Update to version 0.35.0, drop arm64 support

l46983284-cpu:
## Summary
- v0.35.0 no longer publishes `swagger_windows_amd64.exe` / Windows arm64 binaries (and the bare amd64 asset is double-ext `swagger_windows_amd64.exe.exe`).
- Switch 64bit to `swagger_$version_Windows_x86_64.zip` (contains `swagger.exe`), drop obsolete arm64, and update autoupdate accordingly.

This is an asset layout / rename fix, not a pure version bump.

## Checklist
- [x] Zip contents verified (`swagger.exe`)
- [x] Hash verified from downloaded zip

### DIFF
--- a/bucket/go-swagger.json
+++ b/bucket/go-swagger.json
@@ -1,16 +1,12 @@
 {
-    "version": "0.33.1",
+    "version": "0.35.0",
     "description": "Golang implementation of Swagger 2.0 (OpenAPI 2.0)",
     "homepage": "https://goswagger.io",
     "license": "Apache-2.0",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/go-swagger/go-swagger/releases/download/v0.33.1/swagger_windows_amd64.exe#/swagger.exe",
-            "hash": "0fe476d2f689f229496eef1c22aa8751a511532ad17f5671aadfe6aa4c2c5bf0"
-        },
-        "arm64": {
-            "url": "https://github.com/go-swagger/go-swagger/releases/download/v0.33.1/swagger_windows_arm64.exe#/swagger.exe",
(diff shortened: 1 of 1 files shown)


z-Fng:
> * v0.35.0 no longer publishes `swagger_windows_amd64.exe` / Windows arm64 binaries (and the bare amd64 asset is double-ext `swagger_windows_amd64.exe.exe`).
> * Switch 64bit to `swagger_$version_Windows_x86_64.zip` (contains `swagger.exe`), drop obsolete arm64, and update autoupdate accordingly.

When dealing with architectures like arm64/ia32 missing from upstream, it is always better to find a concrete reference or official confirmation rather than making changes based purely on temporary symptoms.

If upstream intentionally dropped support for an architecture, they would almost certainly mention it in their changelog or release notes. If there is no mention, it is highly likely an unintended bug or an oversight on their end. In such cases, the issue should be addressed and fixed at the upstream level rather than patched here.

z-Fng:
I have opened an issue with upstream (https://github.com/go-swagger/go-swagger/issues/3391), and they have confirmed that the missing arm64 support was an oversight on their end, not an intentional removal.

I'll close this PR for now. Let's wait for the next upstream release, as this will likely be resolved automatically.

alexchen-sys:
Thanks a lot for the explanation and for checking with upstream — really helpful. Agreed: if arm64 isn't intentionally dropped in the release notes, better to fix it there than patch Scoop. Happy to wait for the next go-swagger release.

z-Fng:
This has been resolved upstream. As of v0.35.1, Scoop's [auto-update](https://github.com/ScoopInstaller/Main/commits/master/bucket/go-swagger.json) mechanism can correctly fetch the latest `go-swagger` releases.

PR #8320, opened 2026-07-17
devcontainer-cli: Add version 0.87.0

### LINKED ISSUE #8319: [Request]: devcontainer-cli
### Prerequisites

### Criteria

- [x] Non-GUI tool
- [x] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [x] English interface (or at least English documentation)
- [x] Latest stable version
- [x] Full version (i.e. not a trial version)
- [x] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

devcontainer-cli

### Description

A reference implementation for the specification that can create and configure a dev container from a devcontainer.json.

### Homepage

https://github.com/devcontainers/cli

### Download Link(s)

https://github.com/devcontainers/cli

Image

FHYQ-Dong:
- References the official CLI from https://github.com/devcontainers/cli
- Uses npm registry as download source
- Depends on nodejs runtime (not bundled)
- Entry point: package/devcontainer.js via node shim

Closes https://github.com/ScoopInstaller/Main/issues/8319

### DIFF
--- a/bucket/devcontainer.json
+++ b/bucket/devcontainer.json
@@ -1,0 +1,27 @@
+{
+    "version": "0.87.0",
+    "description": "A reference implementation for the development container specification",
+    "homepage": "https://github.com/devcontainers/cli",
+    "license": "MIT",
+    "depends": "nodejs",
+    "url": "https://registry.npmjs.org/@devcontainers/cli/-/cli-0.87.0.tgz",
+    "hash": "73f5ee2f149bed32f4f69914d439e85c47ea6de2733e6637dc136d964f249c57",
+    "extract_dir": "package",
+    "bin": [
+        [
+            "node.exe",
+            "devcontainer",
+            "$dir\\devcontainer.js"
+        ]
+    ],
+    "checkver": {
+        "url": "https://registry.npmjs.org/@devcontainers/cli/latest",
+        "jsonpath": "$.version"
+    },
+    "autoupdate": {
+        "url": "https://registry.npmjs.org/@devcontainers/cli/-/cli-$version.tgz",
+        "hash": {
+            "mode": "download"
+        }
+    }
+}


bucket/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+    "license": "MIT",
+    "depends": "nodejs",
+    "url": "https://registry.npmjs.org/@devcontainers/cli/-/cli-0.87.0.tgz",
+    "hash": "73f5ee2f149bed32f4f69914d439e85c47ea6de2733e6637dc136d964f249c57",
+    "extract_dir": "package",
+    "bin": [
+        [
+            "node.exe",
+            "devcontainer",
+            "$dir\\devcontainer.js"
```
z-Fng, line 14: https://github.com/ScoopInstaller/Main/issues/6952#issuecomment-3054133762

> It requires a Node.js runtime, please install it with package manager of NPM ecosystem.

I would advise against wrapping packages installed through npm, pacman, or pip into a Scoop manifest.

FHYQ-Dong:
I understand the concern about wrapping npm packages, but I think this case is different from a typical `npm install -g` wrapper for a few reasons:

**1. The official distribution IS self-contained, not npm-first.**

The project's own install script (https://github.com/devcontainers/cli/blob/main/scripts/install.sh) downloads and bundles a specific Node.js runtime alongside the CLI. It never asks the user to `npm install` anything. The npm registry is used as a *download mirror* for the tarball, not as a package manager — the CLI tarball is fetched via HTTPS, extracted, and run with the bundled Node.js, exactly like any other self-contained tool. There's no `npm install`, no `node_modules`, no dependency resolution at install time.

**2. The manifest is a self-contained tool, not an npm wrapper.**

The manifest downloads two things: a Node.js runtime zip and the CLI tarball. It extracts them side by side and creates a thin `.cmd` launcher. The user never touches npm, never sees a `package.json`, and doesn't need Node.js pre-installed. This is functionally identical to how many Electron-based or runtime-bundled tools work in the Main bucket — the runtime is just a payload.

**3. There are precedents in Main bucket for tools that bundle their own runtime.**

Several Main bucket manifests ship tools that include their own runtime or are downloaded from language-specific registries as pre-built archives — the key distinction is whether the manifest *delegates to a foreign package manager at install time* vs. *uses a registry URL as a download source*. This manifest does the latter.

**What I'm proposing:**

The `pre_install` script is minimal (2 lines): rename the extracted Node.js directory, and write a 1-line `.cmd` launcher. No npm, no `npm install`, no post-install dependency fetching. The tool is fully self-contained after `scoop install`.

Would this be acceptable for Main, given that the tool's own distribution strategy is self-contained rather than npm-native?

z-Fng:
> The manifest downloads two things: a Node.js runtime zip and the CLI tarball. It extracts them side by side and creates a thin `.cmd` launcher. The user never touches npm, never sees a `package.json`, and doesn't need Node.js pre-installed. This is functionally identical to how many Electron-based or runtime-bundled tools work in the Main bucket — the runtime is just a payload.

Most npm packages can do this as well, like [Gemini-CLI](https://github.com/anderlli0053/DEV-tools/blob/master/bucket/gemini-cli.json).

> Would this be acceptable for Main, given that the tool's own distribution strategy is self-contained rather than npm-native?

Scoop does not offer any clear advantages for managing npm or Python packages. It is usually better to use their dedicated package managers instead. My personal advice is to avoid using Scoop as a universal replacement for all tools.

Furthermore, our community is not very active at the moment. Adding a package to the official bucket that is installed via npm would likely lead to more similar requests, which would increase our maintenance overhead.

If you still insist on installing it via Scoop, please use a community-maintained bucket or set up [your own personal bucket](https://github.com/ScoopInstaller/Scoop/wiki/Buckets#creating-your-own-bucket) for it.

FHYQ-Dong:
Ok, thanks for your reply. I'll reconsider my demands.

PR #8322, opened 2026-07-19
uv: Stop forcing UV_* env vars on update

### LINKED ISSUE #8116: [Bug]: uv update overwrites env vars
### Prerequisites

### Package Name

uv

Due to #7653 uv update now overwrites system environment variables.
I have a setup where all my coding data is on a separate `D:` drive and also set the uv cache and other data to be there.
Now it just overwrites those custom folders with `persist` folder on every update.
Since uv can't symlink across the drives it slows down everything as it copies the packages every time.

```
warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
```

### Steps to Reproduce

```console
update uv
```

### Possible Solution

remove hardcoded environment variables and let the user decide where to put cache and other uv data

### Scoop and Buckets Version

```console
b588a06e (HEAD -> master, origin/master, origin/HEAD) chore(release): Bump to version 0.5.3 (resync) (#6436)

'main' bucket:
ab0115438a (HEAD -> master, origin/master, origin/HEAD) vacuum: Update to version 0.29.2

'extras' bucket:
e8314d4603 (HEAD -> master, origin/master, origin/HEAD) treesheets: Update to version 3207

'versions' bucket:
e7bbb581fc (HEAD -> master, origin/master, origin/HEAD) yazi-nightly@8e80798: Fix hash (Closes #2910)

'sysinternals' bucket:
1000086 (HEAD -> main, origin/main, origin/HEAD) zoomit: Update to version 12.0

'nonportable' bucket:
af76d959 (HEAD -> master, origin/master, origin/HEAD) burp-suite-community-np@2026.5.1: Fix hash (Closes #595)

'psmux' bucket:
53bdeec (HEAD -> main, origin/main, origin/HEAD) Update psmux to 3.3.5
```

### Scoop Config

```console
06-Jun-26 3:49:16 PM https://github.com/ScoopInstaller/Scoop master
```

### PowerShell Version

```console
PSVersion                      7.6.2
GitCommitId                    7.6.2
OS                             Microsoft Windows 10.0.26200
PSRemotingProtocolVersion      2.4
```

### Additional Softwares

_No response_

l46983284-cpu:
### Prerequisites

## Summary

`uv` currently uses `env_set` for `UV_CACHE_DIR` / `UV_PYTHON_*` / `UV_TOOL_*`, so every Scoop update **overwrites** user-chosen locations (issue #8116). That breaks multi-drive layouts and existing tool/cache paths.

This PR:
- removes the forced `env_set` block (users keep their existing `UV_*` values)
- adds `persist` for `cache` / `python` / `tools` so Scoop-managed data still survives updates if used
- keeps PATH installer/uninstaller hooks for the shims dirs
- documents optional self-managed `UV_*` in `notes`

Closes #8116

## Test plan
- [x] `uv.json` parses as JSON
- [ ] Windows: install/update `uv` no longer rewrites pre-set `UV_CACHE_DIR`
- [ ] PATH shims dirs still added when present

### DIFF
--- a/bucket/uv.json
+++ b/bucket/uv.json
@@ -29,15 +29,15 @@
             "Add-Path -Path \"$persist_dir\\tools\\shims\" -Global:$global -Force"
         ]
     },
-    "persist": [
-        "cache",
-        "python",
-        "tools"
-    ],
     "bin": [
         "uv.exe",
         "uvx.exe",
         "uvw.exe"
+    ],
+    "persist": [
+        "cache",
+        "python",
+        "tools"
     ],
     "uninstaller": {
         "script": [
--- a/bucket/uv.json
+++ b/bucket/uv.json
@@ -4,9 +4,10 @@
     "homepage": "https://docs.astral.sh/uv/",
     "license": "Apache-2.0|MIT",
     "notes": [
-        "Scoop persists data since version 0.11.16. You may need to move data by yourself.",
-        "Before: '$env:APPDATA\\uv', '$env:LOCALAPPDATA\\uv'",
-        "After : '$persist_dir'"
+        "This manifest no longer forces UV_* environment variables on install/update (see #8116).",
+        "Set UV_CACHE_DIR / UV_PYTHON_* / UV_TOOL_* yourself if you want non-default locations.",
+        "Optional Scoop-managed paths under $persist_dir: cache, python, tools (persist keeps them across updates).",
+        "Installer still adds $persist_dir\\python\\shims and $persist_dir\\tools\\shims to PATH when present."
     ],
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


z-Fng:
As discussed in https://github.com/ScoopInstaller/Main/pull/7653#issuecomment-4533026354, setting these environment variables is the expected behavior. Rather than reverting the previous changes, a better approach would be implementing the logic to avoid overriding existing UV_* environment variables (https://github.com/ScoopInstaller/Main/issues/8053#issuecomment-4646839597) within Scoop-Core.

alexchen-sys:
Thanks, that makes sense — I was treating #8116 as a manifest-level fix for the #7653 env_set block, but I understand now that those UV_* defaults are intentional Scoop-managed behavior and the right fix is set-if-unset logic in Scoop-Core.

I'm happy to work on that approach (only set UV_* values when the user hasn't already defined them), as discussed in #8053. I can open a Core issue or PR for it if that's useful — otherwise I'll leave Main alone and won't reopen a wipe-style change here.

PR #8360, opened 2026-08-06
uutils-coreutils@0.10.0: Add shim for `timeout`

God-damnit-all:
uutils coreutils 0.10.0 now has a Windows implementation for `timeout`, so it has been included.

---

As an aside to anyone who reviews this: I'm thinking of doing a PR that changes the `uutils.exe` shim here to `uutils-coreutils.exe` (it made sense back when coreutils was pretty much all uutils put out but now they have several things, it's like having a `microsoft.exe` shim).

Similarly, I want to make the uutils shim in the microsoft implementation of coreutils named `microsoft-coreutils.exe` instead of `uutils.exe` (I think this was the result of a copy-and-paste).

Let me know if you think this is a good idea.

### DIFF
--- a/bucket/uutils-coreutils.json
+++ b/bucket/uutils-coreutils.json
@@ -86,6 +86,7 @@
                 "tail.exe",
                 "tee.exe",
                 "test.exe",
+                "timeout.exe",
                 "touch.exe",
                 "tr.exe",
                 "true.exe",
@@ -466,6 +467,11 @@
                     "coreutils.exe",
                     "test",
                     "test"
+                ],
+                [
+                    "coreutils.exe",
+                    "timeout",
+                    "timeout"
                 ],
                 [
                     "coreutils.exe",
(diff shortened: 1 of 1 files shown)


God-damnit-all:
/verify

(I often forget to do this, sorry about that.)

z-Fng (approved):

z-Fng:
> As an aside to anyone who reviews this: I'm thinking of doing a PR that changes the `uutils.exe` shim here to `uutils-coreutils.exe` (it made sense back when coreutils was pretty much all uutils put out but now they have several things, it's like having a `microsoft.exe` shim).

I'd suggest keeping this as-is. The shim was added intentionally (https://github.com/ScoopInstaller/Main/commit/a04240cac2942626595649e6e6941bc16c0df003). The `uutils` shim was kept because it was previously named `uutils`. And it's unclear what this change might break.

> Similarly, I want to make the uutils shim in the microsoft implementation of coreutils named `microsoft-coreutils.exe` instead of `uutils.exe` (I think this was the result of a copy-and-paste).

Good catch on microsoft-coreutils - that was indeed a copy-paste error, which I've addressed in https://github.com/ScoopInstaller/Main/commit/6892daa643d1b3aaf796241e72046083b425fbfa.

PR #8363, opened 2026-08-06
uutils-coreutils-lean: Update to version 0.10.0, switch back to MSVC builds

God-damnit-all:
`uutils/coreutils` no longer publishes a `gnu`-targeted asset. The `msvc`-targeted 64bit asset contains `coreutils.exe` *in addition to* the individual binaries. 32bit and arm64 still only have `coreutils.exe`.

This is essentially now a stripped-down version of the uutils-coreutils.json bucket, except I'm using a `Get-ChildItem -PSPath $dir -af` invocation to get the extracted files, then deleting any exe files that aren't `coreutils.exe`.

### DIFF
--- a/bucket/uutils-coreutils-lean.json
+++ b/bucket/uutils-coreutils-lean.json
@@ -1,14 +1,37 @@
 {
-    "version": "0.9.0",
-    "description": "Rust implementation of GNU coreutils (binaries compiled with MSVC)",
+    "version": "0.10.0",
+    "description": "A cross-platform Rust reimplementation of GNU coreutils (binaries compiled with MSVC), with some options or behaviors potentially differing from GNU coreutils.",
     "homepage": "https://uutils.github.io/coreutils/",
-    "license": "MIT",
+    "license": {
+        "identifier": "MIT",
+        "url": "https://github.com/uutils/coreutils/blob/HEAD/LICENSE"
+    },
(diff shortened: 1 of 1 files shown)


z-Fng (approved):

z-Fng:
> `uutils/coreutils` no longer publishes a `gnu`-targeted asset. The `msvc`-targeted 64bit asset contains `coreutils.exe` _in addition to_ the individual binaries. 32bit and arm64 still only have `coreutils.exe`.

As a side note, [v0.7.0](https://github.com/uutils/coreutils/releases/tag/0.7.0) previously switched to GNU builds (https://github.com/ScoopInstaller/Main/commit/852bd3e432f054cf3b10443e95077eabbbe320b8) because the MSVC builds didn't include `coreutils.exe`.

God-damnit-all:
@z-Fng Your changes are reliant on wildcard pathing rather than literal pathing, and there's not really even any reason to have both a wildcard for the directory *and* an `-Include` parameter. Also, `-Filter` is considered to be much faster than `-Include`

The way I had it set up was meant for absolute safety, but if you want to stick with this, I strongly recommend switching to the following:

`Get-ChildItem -PSPath $dir -Filter *.exe | Where-Object Name -ne 'coreutils.exe' | Remove-Item`

PR #7652, opened 2026-02-17
libjxl@0.11.2: Fix binary paths

eugenesvk:
binaries have been moved to bin/, so old shims failed

This is similar to https://github.com/ScoopInstaller/Main/pull/7650, but doesn't lose any `.dll` files from the `lib` folder

Closes #7649

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #8365, opened 2026-08-06
jxl: Add version 0.12.0

### LINKED ISSUE #8364: [Request]: libjxl (Binaries only)
### Prerequisites

### Criteria

- [x] Non-GUI tool
- [x] Reasonably well-known and widely used (e.g. if it's a GitHub project, it should have at least 500 stars and/or 150 forks)
- [x] English interface (or at least English documentation)
- [x] Latest stable version
- [x] Full version (i.e. not a trial version)
- [x] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

jxl

### Description

JPEG XL image format (.jxl) encode/decode tools

### Homepage

https://jpeg.org/jpegxl

### Download Link(s)

https://github.com/libjxl/libjxl/releases

3.8k stars, but it already is part of Scoop, this just removes the extra LIB parts that most users don't need, only leaving the cli .exe binaries

eugenesvk:
Closes #8364

This is a copy of libjxl, but stripping all the `lib` parts, only leaving the `exe` binaries

Two potential issues:

  - By the way, is there a way to mark this as conflicting with the other since they share the same binaries and will overwrite each other? didn't find anything in the manifest docs on conflicts (though maybe in this case it's not an issue since the binaries don't depend on the libraries? but maybe if not all binaries are included in this package, then it becomes an issue)

  - Also maybe some of the binaries are only useful for development, so should be excluded from this one (e.g., most useful ones are likely just cjxl/djxl/jxlinfo and maybe ssimulacra), but I haven't looked through all of them to determine what can be dropped, so for now left all in (also to extract only some of the binaries would require post-install script, right, there is no magic manifest field `extract_file_list`?).

### DIFF
--- a/bucket/jxl.json
+++ b/bucket/jxl.json
@@ -1,0 +1,54 @@
+{
+    "version": "0.12.0",
+    "description": "JPEG XL image format (.jxl) encode/decode tools",
+    "homepage": "https://jpeg.org/jpegxl",
+    "license": "BSD-3-Clause",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/libjxl/libjxl/releases/download/v0.12.0/jxl-x64-windows-static.7z",
+            "hash": "ff147dc7ac4ce55392974ccc70f2a8a8ec0eff3ae28529b072258b66c8f01ab2",
+            "extract_dir": "x64-windows-static/bin"
+        },
+        "32bit": {
+            "url": "https://github.com/libjxl/libjxl/releases/download/v0.12.0/jxl-x86-windows-static.7z",
+            "hash": "c6f419659910a68782810a400aadaf5c1bfcbc67ea324223af09d7a7477c16cc",
+            "extract_dir": "x86-windows-static/bin"
+        }
+    },
+    "bin": [
+        "benchmark_xl.exe",
+        "butteraugli_main.exe",
+        "cjxl.exe",
+        "decode_and_encode.exe",
+        "display_to_hlg.exe",
+        "djxl_fuzzer_corpus.exe",
+        "djxl.exe",
+        "exr_to_pq.exe",
+        "generate_lut_template.exe",
+        "icc_simplify.exe",
+        "jxl_from_tree.exe",
+        "jxlinfo.exe",
+        "jxltran.exe",
+        "local_tone_map.exe",
+        "pq_to_hlg.exe",
+        "render_hlg.exe",
+        "ssimulacra_main.exe",
+        "ssimulacra2.exe",
+        "texture_to_cube.exe",
+        "tone_map.exe",
+        "xyb_range.exe"
+    ],
+    "checkver": {
+        "github": "https://github.com/libjxl/libjxl"
+    },
+    "autoupdate": {
+        "architecture": {
+            "64bit": {
+                "url": "https://github.com/libjxl/libjxl/releases/download/v$version/jxl-x64-windows-static.7z"
+            },
+            "32bit": {
+                "url": "https://github.com/libjxl/libjxl/releases/download/v$version/jxl-x86-windows-static.7z"
+            }
+        }
+    }
+}


z-Fng:
Creating multiple variants of a package is discouraged due to long-term maintenance burden. Furthermore, there is not much difference from the original, aside from missing some files. If you wish to create a dedicated package for this reduced file set, we recommend hosting it in a custom bucket and maintaining it independently.

z-Fng:
> * By the way, is there a way to mark this as conflicting with the other since they share the same binaries and will overwrite each other? didn't find anything in the manifest docs on conflicts (though maybe in this case it's not an issue since the binaries don't depend on the libraries? but maybe if not all binaries are included in this package, then it becomes an issue)

Not yet. However, running `scoop reset <app>` lets you easily switch the active package for the shim.

> * Also maybe some of the binaries are only useful for development, so should be excluded from this one (e.g., most useful ones are likely just cjxl/djxl/jxlinfo and maybe ssimulacra), but I haven't looked through all of them to determine what can be dropped, so for now left all in (also to extract only some of the binaries would require post-install script, right, there is no magic manifest field `extract_file_list`?).

Not yet. Implementing it might break existing packages, so it's best to create a custom package instead.

eugenesvk:
But what's the reason to have the original bloated package? Isn't scoop supposed to be about cli tools, not programming hedaers? Then the original oversight should be fixed, just because the upstream is bundling everything is doesn't mean scoop should bloat installations for all its users

z-Fng:
> But what's the reason to have the original bloated package? Isn't scoop supposed to be about cli tools, not programming hedaers? Then the original oversight should be fixed, just because the upstream is bundling everything is doesn't mean scoop should bloat installations for all its users

I agree with your point. However, the main reason for closing this PR is that we shouldn't add more package variants to the Main bucket. If we want to make it less bloated based on the point above, we should directly update the existing libjxl package instead (i.e., revert changes related to #7652 ). Merging #7652 was done considering that if any users need these header files and static libraries, we will also accept PRs to keep them (your proposal). Unfortunately, given our current maintainability limits, we simply cannot meet everyone's needs at the same time. It comes back to what was said before: if you truly cannot accept the so-called bloat, please consider maintaining your own manifest instead.

eugenesvk:
> was done considering that if any users need these header files and static libraries, we will also accept PRs to keep them (your proposal). 

yeah, unfortunatly I still don't know if users depend on that, and have no way of finding out, so don't want to risk breaking them :( That was the reason for the extra package

Re. your concern about multiple versions - do you think the "versions" bucket is the right place for this? (ideally think it should  be the other way around, versions is specifically mentioned as "For ...  dev", so guess will move it there

frabera:
I don't think users depend on that, the libraries are not added to path and the users should manually add them, this is quite out of scope for scoop common usage

z-Fng:
> I don't think users depend on that, the libraries are not added to path and the users should manually add them, this is quite out of scope for scoop common usage

Manually configuring the include and static library search paths should be pretty common, I guess, although it's not recommended. There are some similar packages, such as [apache](https://github.com/ScoopInstaller/Main/blob/master/bucket/apache.json). These extra header files/libraries won't actually take up much space, so most people won't even notice. I honestly don't feel like there's much of a difference.

image

You could revert the changes related to #7652, though someone might submit a PR to add it back later anyway...

eugenesvk:
Well, it's ~20 of the size here, and if other dev binaries are excluded, maybe the whole ~70% is extra. There are a few minor "aesthetical" issues like an extra bin subfolder and getting noise in file search when e.g., you'd like to see what other JXL tools are there.

I've meanwhile added a non-dev version to the Versions bucket, think once it's added the manifests could be swapped between buckets (so that there is always both full/dev versions available) - and then the paths will remain the same, `libjxl` with all the headers if someone has linked to them they should ~only use a different bucket

> though someone might submit a PR to add it back later anyway

They would have the dev version in the Versions bucket by then

PR #8370, opened 2026-08-07
codex@0.147.0: Switch download to package archive

RtYkk:
## Summary

Switch the `codex` manifest to the official pre-built Codex package archives for Windows x86_64 and ARM64.

## Problem

Scoop-installed Codex 0.147.0 currently reports this error on startup and Code Mode fails closed:

```text
Code Mode is unavailable because failed to spawn code-mode host
C:\\Users\\<user>\\scoop\\apps\\codex\\current\\codex-code-mode-host.exe:
host executable was not found. Code mode will fail closed; enable
`features.code_mode_host` and install `codex-code-mode-host`.
```

## Why the official package archive

The official `codex-package-<target>.tar.gz` archive is the canonical Codex package and contains all required Windows runtime files in one layout:

- `bin/codex.exe`
- `bin/codex-code-mode-host.exe`
- `codex-resources/codex-command-runner.exe`
- `codex-resources/codex-windows-sandbox-setup.exe`
- `codex-path/rg.exe`
- `codex-package.json`

Codex detects this package layout from `codex-package.json`, resolves the bundled `rg.exe` from `codex-path`, and resolves the code-mode host beside the entrypoint in `bin`. This follows the package format produced and consumed by the official Codex installer.

## Changes

- Replace the individually listed Windows binaries with the official x86_64 and aarch64 package archives.
- Use the release-provided SHA-256 digests for both archives.
- Point the `codex` shim at `bin\\codex.exe` inside the extracted package.
- Use package archive URLs in both autoupdate architecture entries.

No separate `ripgrep` dependency is declared because the official package already contains its runtime `rg.exe`. No `bin` shim is added for the host. No `autoupdate.hash` section is needed because Scoop's built-in GitHub hash mode reads the release asset digest.

Evidence:

- https://github.com/openai/codex/releases/tag/rust-v0.147.0
- https://github.com/openai/codex/blob/rust-v0.147.0/scripts/codex_package/README.md
- https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/install-context/src/lib.rs

## Verification

- Downloaded both package archives and verified their SHA-256 hashes against release metadata.
- Verified both archives contain the canonical package layout and required executables.
- Verified the manifest parses as JSON and `git diff --check` passes.
- PowerShell is unavailable in this environment, so the repository PowerShell verifier could not be run locally; requesting Main's manifest verifier with `/verify`.

Relates to #8369

[...]

Issue #8369, opened 2026-08-07
codex: package missing code-mode host and ripgrep runtime dependency


RtYkk:
## Problem

The `codex` manifest at version 0.147.0 omits runtime files that the official Codex Windows package provides.

Scoop-installed Codex currently reports on startup:

```text
Code Mode is unavailable because failed to spawn code-mode host
C:\\Users\\<user>\\scoop\\apps\\codex\\current\\codex-code-mode-host.exe:
host executable was not found. Code mode will fail closed; enable
`features.code_mode_host` and install `codex-code-mode-host`.
```

The manifest currently downloads `codex.exe`, `codex-command-runner.exe`, and `codex-windows-sandbox-setup.exe`, but not `codex-code-mode-host.exe`.

## Evidence

- Codex release `rust-v0.147.0` publishes the Windows host assets:
  https://github.com/openai/codex/releases/tag/rust-v0.147.0
- The official package builder documents `codex-code-mode-host.exe` and `codex-path/rg.exe` in the canonical Windows package layout:
  https://github.com/openai/codex/blob/rust-v0.147.0/scripts/codex_package/README.md
- The official Windows installer validates `rg.exe` as a required package file:
  https://github.com/openai/codex/releases/download/rust-v0.147.0/install.ps1
- Release notes for 0.147.0 include #36217, which moved Code Mode exclusively to the standalone host process.

## Proposed changes

Please review these as two separate PRs:

1. Add `codex-code-mode-host.exe` to both Windows architectures, including hashes and autoupdate URLs.
2. Add top-level `"depends": "ripgrep"` as a separate follow-up PR, because the official Codex package includes ripgrep as a runtime component and Codex resolves it through PATH in standalone installs.

The second PR can be changed to `suggest` if maintainers consider ripgrep optional for the Scoop manifest, but it should remain isolated from the host binary fix.

z-Fng:
Switching directly to the pre-built `codex-package-x86_64-pc-windows-msvc.tar.gz` should fix the above issue. It includes all required binaries, so we don't need to list them separately or add `ripgrep` as a dependency.

RtYkk:
Updated #8370 to use the official `codex-package-<target>.tar.gz` archives as suggested. The package contains the code-mode host and bundled `rg.exe`, so the separate ripgrep dependency PR (#8371) has been closed.

RtYkk:
[...]

PR #8371, opened 2026-08-07
codex@0.147.0: Add ripgrep dependency

RtYkk:
## Summary

Declare `ripgrep` as a runtime dependency of the `codex` manifest.

## Why `depends`

The official Codex Windows distribution bundles ripgrep as part of its runtime package:

- The official package builder documents `codex-path\\rg.exe` in the canonical package layout:
  https://github.com/openai/codex/blob/rust-v0.147.0/scripts/codex_package/README.md
- The official Windows installer validates `rg.exe` as a required package file and includes it in the expected package contents:
  https://github.com/openai/codex/releases/download/rust-v0.147.0/install.ps1
- The Codex runtime doctor checks the selected `rg` command and recommends installing ripgrep when it cannot be verified:
  https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/cli/src/doctor/runtime.rs

The Scoop manifest exposes Codex's Windows binaries individually and does not include the bundled `rg.exe`. Scoop's `depends` field automatically installs runtime dependencies, so `"depends": "ripgrep"` preserves the official installation contract and ensures `rg` is available on PATH.

This is intentionally a separate follow-up PR to the `codex-code-mode-host.exe` asset fix in #8370. It changes no URLs, hashes, autoupdate rules, `bin` entries, or `checkver` configuration.

If maintainers consider an externally installed compatible `rg` sufficient for the Scoop package contract, this can be changed to a `suggest` entry; that would be an explicit policy change because `suggest` only displays an optional recommendation and does not install the runtime automatically.

## Verification

- Verified the manifest parses as JSON and the diff contains only the single `depends` field.
- `git diff --check` passes.
- PowerShell is unavailable in this environment, so the repository PowerShell verifier could not be run locally; requesting Main's manifest verifier with `/verify`.

Relates to #8369

### DIFF
--- a/bucket/codex.json
+++ b/bucket/codex.json
@@ -3,6 +3,7 @@
     "description": "OpenAI's Codex AI coding assistant",
     "homepage": "https://github.com/openai/codex",
     "license": "Apache-2.0",
+    "depends": "ripgrep",
     "architecture": {
         "64bit": {
             "url": [


RtYkk:
Superseded by #8370, which now uses the official Codex package archive containing the bundled ripgrep runtime. Closing this separate dependency PR so the two approaches are not combined.

PR #8428, opened 2026-08-22
opus-tools: Update to version 0.2-52-g5f894e9, switch source to rarewares.org

reaseno:
## Summary
- The previous download source (archive.mozilla.org) no longer receives opus-tools updates
- Switched to rarewares.org, which publishes current builds
- Updated to version 0.2-52-g5f894e9 (libopus 1.6.1-11-g788cc89) with fresh hashes
- Adjusted checkver/autoupdate to match rarewares' filename scheme via named capture groups

## Test plan
- [x] Verified both 64bit and 32bit URLs download valid zip archives
- [x] Recomputed and confirmed sha256 hashes for both architectures

### DIFF
--- a/bucket/opus-tools.json
+++ b/bucket/opus-tools.json
@@ -1,16 +1,16 @@
 {
-    "version": "0.2-opus-1.3",
+    "version": "0.2-52-g5f894e9",
     "description": "Command-line utilities to encode, inspect, and decode .opus files.",
     "homepage": "https://www.opus-codec.org/",
     "license": "BSD-2-Clause|GPL-2.0-only",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


z-Fng:
> * The previous download source (archive.mozilla.org) no longer receives opus-tools updates
> * Switched to rarewares.org, which publishes current builds

From what I can tell, `rarewares.org` provides third-party community builds rather than official releases. Switching an existing manifest to a third-party distributor just to bump the version breaks the trust chain.

Unless there is a strong reason or clear verification of the build pipeline, we should probably stick to the official upstream binaries. If you'd like to support this build, it might be better to maintain a separate `opus-tools-rarewares` manifest in your personal bucket instead.

See also: https://scoop.sh/#/apps?q=rarewares&o=false&p=1

PR #8433, opened 2026-08-24
go: Improve installer and uninstaller scripts

m1croalgae:
- The output from the scripts was printed directly at the end of the "Running (un)installer script..." prompt.
  Added `Write-Host` to start a new line.
- Removed the `info` lines because `Add-Path` and `Remove-Path` already print their messages.
- Simplified the `$envgopath` assignment.

Update: Handled multi-entry `GOPATH` by selecting the first non-empty entry.
Update: Wrapped `Remove-Path` in `if ($cmd -eq 'uninstall')` to skip PATH removal during `scoop update`.

---

- [x] Use conventional PR title: `<manifest-name[@version]|chore>: <general summary of the pull request>`
- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md)

PR #8440, opened 2026-08-25
go: Handle multi-entry GOPATH

m1croalgae:
Follow-up to https://github.com/ScoopInstaller/Main/pull/8433

Select the first non-empty `GOPATH` entry when updating `PATH` in install and uninstall scripts to prevent malformed entries when `GOPATH` contains multiple paths.

---

### DIFF
--- a/bucket/go.json
+++ b/bucket/go.json
@@ -20,17 +20,17 @@
     "extract_dir": "go",
     "installer": {
         "script": [
-            "$envgopath = \"$env:USERPROFILE\\go\"",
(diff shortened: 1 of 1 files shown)


z-Fng (approved):

vbnmklop:
When there is only one entry in the `GOPATH` and no semicolon following it, `$first_entry` will be assigned the value `True`.

m1croalgae:
Nice catch, thank you! My bad, completely forgot to test the simplest case. Really struggling with PowerShell.

